PR #43, opened 2019-08-07
Show progress when collecting support bundle

divolgin:


cmd/troubleshoot/cli/run_nocrd.go
```diff
@@ -31,6 +34,9 @@ import (
 )
 
 func runTroubleshootNoCRD(v *viper.Viper, arg string) error {
+	fmt.Printf("%s", cursor.Hide())
```
marccampbell: Let's put this in the preflight command too?

cmd/troubleshoot/cli/run_nocrd.go
```diff
@@ -68,17 +74,46 @@ func runTroubleshootNoCRD(v *viper.Viper, arg string) error {
 		return fmt.Errorf("unable to parse %s collectors", arg)
 	}
 
-	archivePath, err := runCollectors(v, collector)
+	s := spin.New()
+	finishedCh := make(chan bool, 1)
+	progressChan := make(chan string, 10)
```
marccampbell: What's the reason for 10? Simply to prevent blocking? This should be really quick though, what about making this lower? What's the risk there?

marccampbell (commented):
This looks good. One question about the channel, and a request to put this in preflight so we have the same cursor hide/show behavior.

cmd/troubleshoot/cli/run_nocrd.go
divolgin: no risk

cmd/troubleshoot/cli/run_nocrd.go
divolgin: added

marccampbell (approved):

PR #45, opened 2019-08-07
Better message at the end of support bundle generation

divolgin:


cmd/troubleshoot/cli/run_nocrd.go
```diff
@@ -104,7 +104,16 @@ func runTroubleshootNoCRD(v *viper.Viper, arg string) error {
 	}
 
 	fmt.Printf("\r%s", cursor.ClearEntireLine())
-	fmt.Printf("%s\n", archivePath)
+
+	msg := archivePath
+	if appName := collector.Labels["applicationName"]; appName != "" {
+		f := `A support bundle for %q has been created in this directory
+named %s. Please upload it on the Troubleshoot page of
+the %q to begin analysis.`
```
marccampbell: will you change both %q to just %s, and then the last one should be ... page of the %s Admin Console to begin analysis"

marccampbell (approved):

PR #74, opened 2019-10-29
Analyze spec

marccampbell:


cmd/analyze/cli/run.go
```diff
@@ -0,0 +1,75 @@
...
+		} else if analyzeResult.IsFail {
+			fmt.Printf("Fail: %s\n %s\n", analyzeResult.Title, analyzeResult.Message)
+		}
+	}
+
+	return nil
+}
+
+func isURL(str string) bool {
+	_, err := url.ParseRequestURI(str)
```
divolgin: pretty sure this parses absolute file paths

divolgin: you should just check if file exists

marccampbell: this is elsewhere in the code...  seems like a bigger issue?

divolgin (approved):

cmd/analyze/cli/run.go
divolgin: yeah.... I see it

PR #91, opened 2019-12-04
Added Namespace for Collection and Allows for tests to output to stdout

bearium:
--namespace flag can now be passed into the support bundle and will now only collect resource information from specified namespace. if flag not specified default cluster wide resource collection is used.

--interactive=false now results in tests printing to stdout in go test format, allows for easier and quicker test inspection

go mod tidy was ran to clean up dependencies


marccampbell:
@bearium thanks for the contribution! This lgtm, and we will include it in our next release (Dec 9)

cmd/preflight/cli/stdout_results.go
```diff
@@ -0,0 +1,41 @@
...
+
+var failed = false
+
+func showStdoutResults(preflightName string, analyzeResults []*analyzerunner.AnalyzeResult) error {
+	fmt.Printf("\n=== TEST   %s\n", preflightName)
+	for _, analyzeResult := range analyzeResults {
+		fmt.Printf("=== RUN:   %s\n", analyzeResult.Title)
+	}
+	for _, analyzeResult := range analyzeResults {
+		outputResult(analyzeResult)
```
divolgin: `outputResult` should return the value that is currently stored in the global `failed` variable.   The the global variable should be removed.

divolgin (changes requested):
Thanks for the pull request.  There's just one small nit-picky thing.  Otherwise it looks good.

cmd/preflight/cli/stdout_results.go
```diff
@@ -0,0 +1,41 @@
+package cli
+
+import (
+	"fmt"
+
+	analyzerunner "github.com/replicatedhq/troubleshoot/pkg/analyze"
+)
+
+var failed = false
```
laverya: Why the global variable and not something like having `outputResult` return the failure status?

laverya (approved):

laverya:
Thanks for the PR!

Issue #90, opened 2019-12-02
Proposal: Runtime-Configurable Preflights in KOTS


dexhorthy:
Full writeup is in: [Proposal: Runtime-Configurable Preflights in KOTS](https://docs.google.com/document/d/19LbG8gQy3JgpWUx4ZoVGdRCeVGPiJ6wWVpVD4sarmug/edit?ts=5de56ffd#)

The doc outlines a proposal for including a ConfigOption-aware context in Troubleshoot Collectors/Analyzers so that they can be conditionally evaluated based on customer config. This would let us re-use everything in preflight to implement custom test procedures in preflight/kotsadm.

An example use case: 

```yaml
---
apiVersion: troubleshoot.replicated.com/v1beta1
kind: Preflight
metadata:
  name: preflight-sample
spec:
  analyzers:
  - secret:
        checkName: Postgres URI Secret
        secretName: '{{repl ConfigOption "pg_secret_name"}}'
        # we probably need to expose the kotsadm namespace or the downstream namespace
        namespace: default 
        key: uri
        when: '{{repl ConfigOptionEquals "postgres_type" "postgres_from_secret"}}'
        outcomes:
          - fail:
              message: The secret "{{repl ConfigOption "pg_secret_name"}}" was not found or the `uri` key was not detected.
          - pass:
              message: The Postgres URI was found in a secret in the cluster.
```

Open questions:
- Does this require a GVK change? We're currently at v1beta1

marccampbell:
This makes sense to me. My thoughts:

1. v1beta1 can handle this
2. Do you think that using the key name `when` will cause confusion with the current key named `when` a little deeper in the doc (child of each outcome)? I don't know of a recommendation to change, but this causes me a little worry.

This should support Config, License, and Static contexts from KOTS, including sprig and other KOTS built-ins? This should also support both the `{{repl` and the `repl{{` variations?

dexhorthy:
> This should support Config, License, and Static contexts from KOTS, including sprig and other KOTS built-ins? This should also support both the {{repl and the repl{{ variations?

I agree with all these.

For (2), the only alternative I can think of depends on `outcomes` being evaluated in order, but if we depend on that then we could do something like

```yaml
outcomes:
  - skip:
      when: ...
```

marccampbell:
[...]

PR #92, opened 2019-12-18
Optional collectors and analyzers

marccampbell:
This is required for #90


marccampbell:
I also changed `secret.Name` to be `secret.SecretName` for clarity.

And the CRD code is removed now!

divolgin (approved):

PR #99, opened 2019-12-18
Adding a text analyzer

ashwathishiva:
This text analyzer will consume the collector name and file name along with a given regex pattern to check if a matching pattern exists in a given file. Based on the regex match it will return success or failure.


marccampbell:
@ashwathishiva  this looks great. it looks like another PR (#102) was opened today to address the same problem.  

The implementation in your PR allows the analyzer to specify the filename that was collected from the collectorName, which is great.

The implementation in #102 supports match groups, in addition to plain regex. 

How about if we merge the two into a single PR (this one), and extend it with matching?

My first thought is that your struct:

```
type TextAnalyze struct {
	AnalyzeMeta   `json:",inline" yaml:",inline"`
	CollectorName string     `json:"collectorName,omitempty" yaml:"collectorName,omitempty"`
	FileName      string     `json:"fileName,omitempty" yaml:"fileName,omitempty"`
	RegexPattern  string     `json:"regex,omitempty" yaml:"regex,omitempty"`
	Outcomes      []*Outcome `json:"outcomes" yaml:"outcomes"`
}
```

Could also have 1 new member to regex pattern. 

I'd be happy to pull the code from #102 and make it a PR into this PR, if that sounds ok with you

ashwathishiva:
@marccampbell that sounds wonderful! Thanks a lot for reaching out to me about this. Sure, please feel free to carry out your plan as you have proposed above.

divolgin (approved):

marccampbell:
@ashwathishiva thanks for the PR!  This looks great, and we don't want to keep it out of the product. I'm going to merge.

marccampbell (approved):

PR #111, opened 2020-01-08
include binary version info in 'version.yaml' troubleshoot file

laverya:
A static file (shown below) really isn't the most useful
```
kind: SupportBundle
```
The proposed change would produce output like this:
```
kind: SupportBundle
spec:
  versionNumber: v0.9.15
```


cmd/troubleshoot/cli/version.go
```diff
@@ -30,8 +30,9 @@ const VersionFilename = "version.yaml"
 
 func writeVersionFile(path string) error {
 	version := troubleshootv1beta1.SupportBundleVersion{
-		ApiVersion: "troubleshoot.replicated.com/v1beta1",
-		Kind:       "SupportBundle",
+		ApiVersion:    "troubleshoot.replicated.com/v1beta1",
+		Kind:          "SupportBundle",
+		VersionNumber: version.Version(),
```
marccampbell: this feels off here. mind putting it under "spec.versionNumber" instead so that it conforms to a k8s yaml convention?

marccampbell (changes requested):

cmd/troubleshoot/cli/version.go
laverya: done

marccampbell (approved):

PR #126, opened 2020-01-25
Add multipod support bundle scenario

marccampbell:
This is an example that should be added to the GH Action e2e tests on commit


laverya:
the 'tests' directory no longer exists

PR #127, opened 2020-01-29
Node analyzers

marccampbell:
This adds a nodeResources analyzer to the preflight and support bundle.

No new input or collector is needed, this uses the nodes.json that's already collected.

```
kind: Preflight
metadata:
  name: sample
spec:
  analyzers:
    - nodeResources:
        checkName: Must have at least 3 nodes in the cluster
        outcomes:
          - fail:
              when: "< 3"
              message: This application requires at least 3 nodes
          - warn:
              when: "< 5"
              message: This application recommends at last 5 nodes.
          - pass:
              message: This cluster has enough nodes.
    - nodeResources:
        checkName: Must have 3 nodes with at least 6 cores
        filters:
          cpuCapacity: "6"
        outcomes:
          - fail:
              when: "< 3"
              message: This application requires at least 3 nodes with 6 cores each
          - pass:
              message: This cluster has enough nodes with enough codes
    - nodeResources:
        checkName: Must have 1 node with 16 GB (available) memory and 5 cores (on a single node)
        filters:
          allocatableMemory: 16Gi
          cpuCapacity: "5"
        outcomes:
          - fail:
              when: "< 1"
              message: This application requires at least 1 node with 16GB available memory
          - pass:
              message: This cluster has a node with enough memory.
```


pkg/analyze/node_resources.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	operator := parts[0]
+	desiredValue, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return false, errors.Wrap(err, "failed to parse nodeResource value")
+	}
+
+	switch operator {
+	case "=", "==", "===":
```
divolgin, line 111: how about `!=` too?

divolgin (approved):

PR #148, opened 2020-03-07
Fix equality in regex analyzer

marccampbell:


pkg/analyze/text_analyze.go
```diff
@@ -201,16 +201,16 @@ func compareRegex(conditional string, foundMatches map[string]string) (bool, err
...
 		case "<":
-			return lookForValueInt < foundValueInt, nil
+			return lookForValueInt > foundValueInt, nil
 
 		case ">":
 			return lookForValueInt > foundValueInt, nil
 
 		case "<=":
-			return lookForValueInt <= foundValueInt, nil
+			return lookForValueInt >= foundValueInt, nil
```
dexhorthy: would it be more clear to flip lookForValueInt and foundValueInt ?

dexhorthy (approved):

pkg/analyze/text_analyze.go
marccampbell: agreed

PR #150, opened 2020-03-10
Implemented Collectors and Analyzers for Rook-Ceph

moonlight16:
This is a first-pass at implementing collectors and analyzers for Rook-ceph.  At the moment its only supporting a basic analyzer to check ceph status.  Additional checks will be added as needed.  

I still have more testing to do here, but I wanted to get these changes out for review.  These changes were tested on the cli with the following collector and analyzer:

```
kind: Collector
metadata:
  name: collector-sample
spec:
  collectors: 
    - clusterInfo: {}
    - clusterResources: {}
    - rook: {}

```

```
kind: Analyzer
metadata:
  name: analyzer-sample
spec:
  analyzers:
    - rookStatus:
        outcomes:
          - fail:
              when: "Status = HEALTH_ERR"
              message: Rook health is NOT good
          - warn:
              when: "Status = HEALTH_WARN"
              message: Rook status has warnings
          - pass:
              when: "Status = HEALTH_OK"
              message: Rook status is good
```

There will be a separate PR in kotsadm with the same analyzer above as an injector analyzer.


pkg/collect/rook.go
```diff
@@ -0,0 +1,163 @@
...
+func runRookCmd(ctx *Context, client *kubernetes.Clientset, pod corev1.Pod, cmd, args []string) ([]byte, []byte, error) {
+	container := pod.Spec.Containers[0].Name
+	req := client.CoreV1().RESTClient().Post().
+		Resource("pods").
+		Name(pod.Name).
+		Namespace(pod.Namespace).
+		SubResource("exec")
+
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
```
marccampbell: Is corev1 already included in the scheme?

pkg/collect/rook.go
```diff
@@ -0,0 +1,163 @@
...
+
+	b, err := json.MarshalIndent(rookOutput, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+
+	return b, err
+}
+
+func runRookCmd(ctx *Context, client *kubernetes.Clientset, pod corev1.Pod, cmd, args []string) ([]byte, []byte, error) {
```
marccampbell: this function feels like a copy from the run collector. what's unique about it that makes it not desirable to reuse that function?

pkg/collect/rook.go
```diff
@@ -0,0 +1,163 @@
...
+	if len(podsErrors) > 0 {
+		return nil, errors.New("Unable to find pod")
+	}
+
+	rookOutput := make(map[string][]byte)
+	for _, c := range RookCmds {
+		errCh := make(chan error, 1)
+		resultCh := make(chan CmdResult, 1)
+		go func() {
+			// TODO: find pod with 'Running' status, for now use first pod in the array
```
marccampbell: Should we do this? When someone is troubleshooting a cluster, it feels like there's a chance that a pod is in a crashloopbackoff, and this could be relevant

pkg/collect/rook.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func Rook(ctx *Context) ([]byte, error) {
+	client, err := kubernetes.NewForConfig(ctx.ClientConfig)
+	if err != nil {
+		return nil, errors.New("Failed to create kubernetes clientset")
+	}
+
+	namespace := "rook-ceph"
+	selector := []string{"app=rook-ceph-operator"}
```
marccampbell: is this a kURL label or a standard label that all rook pods will have?

pkg/analyze/rook_analyze.go
```diff
@@ -0,0 +1,124 @@
...
+
+func analyzeRook(analyzer *troubleshootv1beta1.RookStatus, getRookCollection func(string) ([]byte, error)) (*AnalyzeResult, error) {
+	title := analyzer.CheckName
+	if title == "" {
+		title = "Rook Status"
+	}
+
+	result := &AnalyzeResult{
+		Title:   title,
+		IconKey: "", // TODO: we'll eventually remove the IconKey
```
marccampbell: what is it?  if it's empty always , can we just remove it from this struct?  isn't "" the default?

pkg/analyze/rook_analyze.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func compareRookStatusToActual(conditional string, getRookCollection func(string) ([]byte, error)) (bool, error) {
+	if conditional == "" {
+		return true, nil
+	}
+
+	parts := strings.Fields(strings.TrimSpace(conditional))
+	if len(parts) != 3 {
+		return false, errors.New("unable to parse nodeResources conditional")
```
marccampbell: this error is a copy/paste error

cmd/troubleshoot/cli/run.go
```diff
@@ -210,6 +210,7 @@ func runCollectors(v *viper.Viper, collector troubleshootv1beta1.Collector, prog
 	collectSpecs = append(collectSpecs, collector.Spec.Collectors...)
 	collectSpecs = ensureCollectorInList(collectSpecs, troubleshootv1beta1.Collect{ClusterInfo: &troubleshootv1beta1.ClusterInfo{}})
 	collectSpecs = ensureCollectorInList(collectSpecs, troubleshootv1beta1.Collect{ClusterResources: &troubleshootv1beta1.ClusterResources{}})
+	collectSpecs = ensureCollectorInList(collectSpecs, troubleshootv1beta1.Collect{Rook: &troubleshootv1beta1.Rook{}})
```
areed, line 213: What happens if Rook isn't installed?

moonlight16: If no pods exist under the 'root-ceph' namespace, then the command won't run and no status files will get created.

moonlight16:
Here's the kotsadm PR:
https://github.com/replicatedhq/kotsadm/pull/1191

marccampbell (changes requested):
This looks like it has some basics of rook collection, but I think we need to provide some options to the developer. Right now, the analysis is abstracted away into this code, not in the analyzer.  The options to write an analyzer feel pretty fixed at :

```
          - fail:
              when: "Status = HEALTH_ERR"
              message: Rook health is NOT good
          - warn:
              when: "Status = HEALTH_WARN"
              message: Rook status has warnings
          - pass:
              when: "Status = HEALTH_OK"
              message: Rook status is good
```

What can we do to expose the options to the developer? What does "WARN" mean? This is awesome to show to a user, but it's not yet going to help debug.

I think we need to make the core metrics and statistics from the rook/ceph operator available so the that developer can decide what they think is warn, error and healthy.

If I used this, I'm not sure I would understand what to do with the output, other than know that rook/ceph was in a "warn" state.

moonlight16:
> This looks like it has some basics of rook collection, but I think we need to provide some options to the developer. Right now, the analysis is abstracted away into this code, not in the analyzer. 

It wasn't my intent to deliver a full analyzer here.  We wanted to build only a basic analyzer to start, at a minimum to warn users that errors exist in cephs.  Then we have the ceph status collected in the support bundle to troubleshoot what might have happened.  It was intended to be an incremental effort to get exactly what would be useful in the analyzer.

pkg/collect/rook.go
moonlight16: kURL label

pkg/analyze/rook_analyze.go
moonlight16: Grayson mentioned we'll eventually remove it.  I left a comment to be explicitly clear, so that nobody thinks this might be a bug or something.  I can remove it if you want.

pkg/collect/rook.go
moonlight16: Yes, the Copy (copyFile) and Exec (getExecOutputs) both have a similar function.  The same argument could have been made there as well.  There's a few minor differences though.  For one, each passes its own collector object so I couldn't just reuse those functions without reorganizing a lot of existing code.  Here, runRookCmd didn't require the analyzer collector object.  

On this pass I didn't want to spend a lot of time refactoring a lot of existing code.  At some point we might want to reorganize the workflow of these collectors and analyzers.  There was some rigidity in other places of code that I didn't want to spend a lot of time refactoring, such as passing functions down through a call stack as one example..  It probably would have been better to build these into Go objects.

pkg/analyze/rook_analyze.go
moonlight16: Fixed

pkg/analyze/rook_analyze.go
moonlight16: I removed it.

pkg/collect/rook.go
moonlight16: See the latest commit.  I moved this to become a more generic function called execPodCmd under util.go.   I'd rather change Exec and Copy later so I don't keep delaying these changes.  We can refactor those functions on a followup commit.

pkg/collect/rook.go
moonlight16: See latest commits.  I just added the ability for the user to specify rook's namespace, selector, and/or container names in the Rook Collector spec (handles the case where user ran KOTS install and manually had installed cephs).  If not specified in the Collector spec, then we'll default to these.  I also added a comment so its clear that kURL labeled this pod.

pkg/collect/rook.go
moonlight16: TODO means "I don't want to forget about this case and it will be eventually done (this commit or some future commit)".  

Now that I've finished a lot of the other testing, I came back around to this and handled this case.  See my latest commit.

pkg/collect/rook.go
```diff
@@ -0,0 +1,135 @@
...
+
+	namespace := rookCollector.Namespace
+	if namespace == "" {
+		namespace = "rook-ceph"
+	}
+
+	selector := rookCollector.Selector
+	if len(selector) == 0 {
+		// kURL labels ceph operator pod as 'rook-ceph-operator'
+		selector = []string{"app=rook-ceph-operator"}
```
emosbaugh, line 82: `app=rook-ceph-tools` in ceph 1.4.x

emosbaugh:
Closed in favor of #295

PR #162, opened 2020-03-26
textanalyze: fixes path handling in windows as well

ashwathishiva:
Fixes a situation where path was not built appropriately for windows

### DIFF
--- a/pkg/analyze/text_analyze.go
+++ b/pkg/analyze/text_analyze.go
@@ -1,7 +1,7 @@
 package analyzer
 
 import (
-	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


marccampbell (approved):

laverya (approved):

marccampbell:
Thanks!

pkg/analyze/text_analyze.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func analyzeTextAnalyze(analyzer *troubleshootv1beta1.TextAnalyze, getCollectedFileContents func(string) ([]byte, error)) (*AnalyzeResult, error) {
-	fullPath := path.Join(analyzer.CollectorName, analyzer.FileName)
+	fullPath := filepath.Join(analyzer.CollectorName, analyzer.FileName)
```
Daanikus, line 14: @CodeLingoBot capture avoid path.Join
path.Join only works for paths separated by forward slashes. This can be problematic on Windows. Use filepath.Join to create paths in a way that is compatible with the target operating systems defined file paths.

PR #166, opened 2020-04-06
support the 'includeValue' flag for secrets

laverya:
also improve path to eliminate collisions (two secrets with the same name in different namespaces, or two keys under a secret)

### DIFF
--- a/pkg/collect/secret.go
+++ b/pkg/collect/secret.go
@@ -3,6 +3,7 @@ package collect
 import (
 	"encoding/json"
 	"fmt"
+	"path/filepath"
 
 	troubleshootv1beta1 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/collect/secret.go
```diff
@@ -60,7 +60,16 @@ func Secret(ctx *Context, secretCollector *troubleshootv1beta1.Secret) ([]byte,
 	return b, nil
 }
 
-func secret(client *kubernetes.Clientset, secretCollector *troubleshootv1beta1.Secret) (*FoundSecret, []byte, error) {
+func secret(client *kubernetes.Clientset, secretCollector *troubleshootv1beta1.Secret) (string, []byte, error) {
+	ns := secretCollector.Namespace
+	if ns == "" {
+		ns = "current-namespace"
```
divolgin: i think this should be empty.  this value will raise questions.

divolgin (approved):

pkg/collect/secret.go
laverya: ...I should be using filepath.join, too

divolgin (approved):

PR #175, opened 2020-04-16
last-mile and per-collector redactor specs

### LINKED ISSUE #171: Feature: add support for last-mile redactors
When running support-bundle, the user should be able to pass a path to a YAML file that contains specific redactors that define data to redact. Because this file is local, it can contain verify specific  patterns and values to redact.

An example spec could be:
```
kind: Redactor
metadata:
  name: sample
spec:
  patterns:
    - `(?i)(\\\"name\\\":\\\"[^\"]*JWT_?TOKEN\\\",\\\"value\\\":\\\")(?P<mask>[^\"]*)(\\\")`
  values:
    - my-secret-value
    - a-password
    - a-api-token
```

The end user would run this with:

```
kubectl support-bundle ./path-to-collectors.yaml --redactors ./path-to-redactors.yaml`
```

laverya:
fixes https://github.com/replicatedhq/troubleshoot/issues/171


cmd/troubleshoot/cli/run.go
```diff
@@ -56,6 +56,19 @@ func runTroubleshoot(v *viper.Viper, arg string) error {
...
+	if v.GetString("redactors") != "" {
+		redactorContent, err := loadSpec(v, v.GetString("redactors"))
+		if err != nil {
+			return errors.Wrap(err, "failed to load redactor spec")
+		}
+		obj, _, err := decode([]byte(redactorContent), nil, nil)
+		if err != nil {
+			return errors.Wrapf(err, "failed to parse redactors %s", v.GetString("redactors"))
+		}
+		additionalRedactors, _ = obj.(*troubleshootv1beta1.Redactor)
```
divolgin: this won't complain if wrong kind is passed in.

pkg/collect/collector_test.go
```diff
@@ -0,0 +1,265 @@
...
+					Name: "data",
+					Data: `abc 123
+another line here
+pwd=somethinggoeshere;`,
+				},
+			},
+			want: map[string]string{
+				"data/datacollectorname": `abc 123
+another line here
+pwd=somethinggoeshere;`,
```
divolgin, line 240: How about also a test for something like environment variables, like

```
env:
    - name: PASSWORD
      value: superrandomsecretstuff
```

divolgin (approved):

pkg/collect/collector_test.go
laverya: sure, but the test is more 'this regex is being run' than 'this specific output is removed'

PR #204, opened 2020-05-27
Don't panic when parsing resources, return error

divolgin:


laverya (approved):

pkg/analyze/node_resources.go
```diff
@@ -94,9 +94,19 @@ func analyzeNodeResources(analyzer *troubleshootv1beta1.NodeResources, getCollec
...
+	defer func() {
+		if r := recover(); r != nil {
+			err = errors.Errorf("failed to evaluate %q: %v", conditional, r)
+		}
+	}()
+
 	if conditional == "" {
-		return true, nil
+		res = true
+		return
```
Daanikus, line 109: @CodeLingoBot capture avoid naked returns
Naked returns harm the readability of code, especially in long functions. Named return parameters do not require returns to be naked.

PR #209, opened 2020-05-29
Make outcome matching less order dependent

divolgin:


pkg/analyze/common_status.go
```diff
@@ -16,16 +16,9 @@ func commonStatus(outcomes []*troubleshootv1beta1.Outcome, title, iconKey string
 	}
 
 	// ordering from the spec is important, the first one that matches returns
+	// outcomes with a when are evaluated first, and the ones with a missing when are treated as an "else"
 	for _, outcome := range outcomes {
-		if outcome.Fail != nil {
-			if outcome.Fail.When == "" {
```
Daanikus, line 21: @CodeLingoBot capture combine conditions in nested if statements
A conditional if statement with a single if statement as a child can be refactored to use a single if statement containing both conditions

PR #223, opened 2020-07-13
update sample preflight and support bundles, update directory name

GraysonNull:


examples/preflight/sample-preflight.yaml
```diff
@@ -24,32 +18,101 @@ spec:
               when: ">= 1.15.0"
               message: Your cluster meets the recommended and required versions of Kubernetes.
     - customResourceDefinition:
-        customResourceDefinitionName: constrainttemplates.templates.gatekeeper.sh
-        checkName: Gatekeeper policy runtime
+        customResourceDefinitionName: ingressroutes.contour.heptio.com
```
markpundsack, line 22: Would this only succeed with contour specifically? What if they have nginx or any number of other ingress controllers? For an example, I'm going checking only for contour, but maybe then the message should specifically say it requires Contour. Ultimately, I'd love a sample check that looks for all the controllers listed on https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/, but maybe the smallest change here is message to be Contour-specific. @marccampbell does that make sense?

examples/preflight/sample-preflight.yaml
```diff
@@ -24,32 +18,101 @@ spec:
...
         outcomes:
           - fail:
-              message: Gatekeeper is required for the application, but not found in the cluster.
-              uri: https://enterprise.support.io/installing/gatekeeper
+              message: IngressRoutes CRD not found!
```
markpundsack:
```suggestion
              message: Contour ingress not found!
```

examples/preflight/sample-preflight.yaml
```diff
@@ -24,32 +18,101 @@ spec:
...
           - pass:
-              message: Found a supported version of Gatekeeper installed and running in the cluster.
-    - imagePullSecret:
-        checkName: Registry credentials for Quay.io
-        registryName: quay.io
+              message: IngressRoutes CRD found!
```
markpundsack:
```suggestion
              message: Contour Ingress found!
```

examples/support-bundle/sample-supportbundle.yaml
```diff
@@ -0,0 +1,100 @@
...
+kind: SupportBundle
+metadata:
+  name: example-support-bundle
+spec:
+  analyzers:
+    - clusterVersion:
+        outcomes:
+          - fail:
+              when: "< 1.16.0"
+              message: The application requires at Kubernetes 1.16.0 or later
```
markpundsack:
```suggestion
              message: The application requires at least Kubernetes 1.16.0 or later
```

markpundsack (changes requested):

examples/preflight/sample-preflight.yaml
markpundsack, line 22:
```suggestion
        checkName: Ingress
        customResourceDefinitionName: ingressroutes.contour.heptio.com
```

markpundsack: The rendered check name is pretty long otherwise.

markpundsack (approved):

Issue #58, opened 2019-08-21
Copy collector does not apply redactions


divolgin:

marccampbell:
bumping to 1.6.0

matias-manavella-nimble:
@divolgin @marccampbell Hi Dimitriy, Marc, since the last update, the redactor opens the tar file collected by the copy collector (and every tar file inside) and apply the redactors to every file, should this be closed?

banjoh:
I believe this can be closed. Will do so unless there are objections @divolgin

PR #237, opened 2020-08-05
Copy collector enhanced

### LINKED ISSUE #212: 'copy' collector should work with directories
Currently the [copy collector](https://troubleshoot.sh/reference/collectors/copy/) calls [cat within the container](https://github.com/replicatedhq/troubleshoot/blob/d016e3269c9e2cf055f1becb546928587c6ae470/pkg/collect/copy.go#L61), which doesn't work with directories - only individual files. This is rather suboptimal.

manavellamnimble:
Hi there! To copy all the files in a given directory, I added this three steps:

1. Execute `find <folder or file path specified> -type f` in the container console, which finds all the files and returns each path
2. I then split the result and store it in an arrange of strings, variable file
3. Lastly I made a loop in the Copy function to cat the file in each path.

I've already tested it with different pods, and with multiple copy statements in the yaml file referencing different containers. Files are stored as before, under namespace/pod.name/containerpath/<subfolders>/ filename in the support-bundle. 

Let me know if there is something you would like to edit/add! 

Fix #212

### DIFF
--- a/pkg/collect/copy.go
+++ b/pkg/collect/copy.go
@@ -1,11 +1,9 @@
 package collect
 
 import (
-	"archive/tar"
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"path"
 
 	"path/filepath"
@@ -66,6 +64,7 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 		container = copyCollector.ContainerName
 	}
 	command := []string{"sh", "-c", fmt.Sprintf("tar -C %v  -cf tmp.tar %v; cat tmp.tar; rm tmp.tar", path.Dir(copyCollector.ContainerPath), path.Base(copyCollector.ContainerPath))}
+
 	req := client.CoreV1().RESTClient().Post().Resource("pods").Name(pod.Name).Namespace(pod.Namespace).SubResource("exec")
 	scheme := runtime.NewScheme()
 	if err := corev1.AddToScheme(scheme); err != nil {
@@ -111,35 +110,10 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 		}
 		return nil, errors
 	}
-	file := make(map[string][]byte)
-	file[path.Base(copyCollector.ContainerPath)+".tar"] = output.Bytes()
-	return file, nil
-}
 
-func untarFile(tarFile *bytes.Buffer, path string) (map[string][]byte, map[string]*tar.Header, map[string]string) {
-	tarReader := tar.NewReader(tarFile)
-	fileHeaders := make(map[string]*tar.Header)
-	files := make(map[string][]byte)
-	for {
-		header, err := tarReader.Next()
-		if err != nil {
-			if err != io.EOF {
-				errors := map[string]string{
-					filepath.Join(path, "error:tarReader:"): err.Error(),
-				}
-				return nil, nil, errors
-			}
-			break
-		}
-
-		if !header.FileInfo().IsDir() {
-			file := new(bytes.Buffer)
-			io.Copy(file, tarReader)
-			files[header.Name] = file.Bytes()
-			fileHeaders[header.Name] = header
-		}
-	}
-	return files, fileHeaders, nil
+	return map[string][]byte{
+		path.Base(copyCollector.ContainerPath) + ".tar": output.Bytes(),
+	}, nil
 }
 
 func getCopyErrosFileName(copyCollector *troubleshootv1beta1.Copy) string {
--- a/pkg/collect/copy.go
+++ b/pkg/collect/copy.go
@@ -1,11 +1,14 @@
 package collect
 
 import (
+	"archive/tar"
 	"bytes"
 	"context"
 	"fmt"
+	"io"
+	"path"
+
 	"path/filepath"
-	"strings"
 
 	troubleshootv1beta1 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -49,7 +52,7 @@ func Copy(c *Collector, copyCollector *troubleshootv1beta1.Copy) (map[string][]b
 			}
 
 			for k, v := range files {
-				copyOutput[filepath.Join(bundlePath, k)] = v
+				copyOutput[filepath.Join(bundlePath, path.Dir(copyCollector.ContainerPath), k)] = v
 			}
 		}
 	}
@@ -58,14 +61,11 @@ func Copy(c *Collector, copyCollector *troubleshootv1beta1.Copy) (map[string][]b
 }
 
 func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyCollector *troubleshootv1beta1.Copy) (map[string][]byte, map[string]string) {
-	files := make(map[string][]byte)
 	container := pod.Spec.Containers[0].Name
 	if copyCollector.ContainerName != "" {
 		container = copyCollector.ContainerName
 	}
-
-	command := []string{"find", copyCollector.ContainerPath, "-type", "f"}
-
+	command := []string{"sh", "-c", fmt.Sprintf("tar -C %v  -cf tmp.tar %v; cat tmp.tar; rm tmp.tar", path.Dir(copyCollector.ContainerPath), path.Base(copyCollector.ContainerPath))}
 	req := client.CoreV1().RESTClient().Post().Resource("pods").Name(pod.Name).Namespace(pod.Namespace).SubResource("exec")
 	scheme := runtime.NewScheme()
 	if err := corev1.AddToScheme(scheme); err != nil {
@@ -99,9 +99,6 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 		Stderr: &stderr,
 		Tty:    false,
 	})
-
-	filepaths := strings.Split(output.String(), "\n")
-
 	if err != nil {
 		errors := map[string]string{
 			filepath.Join(copyCollector.ContainerPath, "error"): err.Error(),
@@ -114,60 +111,35 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 		}
 		return nil, errors
 	}
+	file := make(map[string][]byte)
+	file[path.Base(copyCollector.ContainerPath)+".tar"] = output.Bytes()
+	return file, nil
+}
 
-	for _, file := range filepaths {
-		if file != "" {
-			command := []string{"cat", file}
-
-			req := client.CoreV1().RESTClient().Post().Resource("pods").Name(pod.Name).Namespace(pod.Namespace).SubResource("exec")
-			scheme := runtime.NewScheme()
-			if err := corev1.AddToScheme(scheme); err != nil {
-				return nil, map[string]string{
-					filepath.Join(copyCollector.ContainerPath, "error"): err.Error(),
-				}
-			}
-			parameterCodec := runtime.NewParameterCodec(scheme)
-			req.VersionedParams(&corev1.PodExecOptions{
-				Command:   command,
-				Container: container,
-				Stdin:     true,
-				Stdout:    false,
-				Stderr:    true,
-				TTY:       false,
-			}, parameterCodec)
-
-			exec, err := remotecommand.NewSPDYExecutor(c.ClientConfig, "POST", req.URL())
-			if err != nil {
-				return nil, map[string]string{
-					filepath.Join(copyCollector.ContainerPath, "error"): err.Error(),
-				}
-			}
-
-			output := new(bytes.Buffer)
-			var stderr bytes.Buffer
-			err = exec.Stream(remotecommand.StreamOptions{
-				Stdin:  nil,
-				Stdout: output,
-				Stderr: &stderr,
-				Tty:    false,
-			})
-			files[file] = output.Bytes()
-
-			if err != nil {
+func untarFile(tarFile *bytes.Buffer, path string) (map[string][]byte, map[string]*tar.Header, map[string]string) {
+	tarReader := tar.NewReader(tarFile)
+	fileHeaders := make(map[string]*tar.Header)
+	files := make(map[string][]byte)
+	for {
+		header, err := tarReader.Next()
+		if err != nil {
+			if err != io.EOF {
 				errors := map[string]string{
-					filepath.Join(copyCollector.ContainerPath, "error"): err.Error(),
+					filepath.Join(path, "error:tarReader:"): err.Error(),
 				}
-				if s := output.String(); len(s) > 0 {
-					errors[filepath.Join(copyCollector.ContainerPath, "stdout")] = s
-				}
-				if s := stderr.String(); len(s) > 0 {
-					errors[filepath.Join(copyCollector.ContainerPath, "stderr")] = s
-				}
-				return nil, errors
+				return nil, nil, errors
 			}
+			break
+		}
+
+		if !header.FileInfo().IsDir() {
+			file := new(bytes.Buffer)
+			io.Copy(file, tarReader)
+			files[header.Name] = file.Bytes()
+			fileHeaders[header.Name] = header
 		}
 	}
-	return files, nil
+	return files, fileHeaders, nil
 }
 
 func getCopyErrosFileName(copyCollector *troubleshootv1beta1.Copy) string {
--- a/pkg/collect/copy.go
+++ b/pkg/collect/copy.go
@@ -61,7 +61,7 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 	if copyCollector.ContainerName != "" {
 		container = copyCollector.ContainerName
 	}
-	command := []string{"sh", "-c", fmt.Sprintf("tar -C %s  -cf - %s | cat ", filepath.Dir(copyCollector.ContainerPath), filepath.Base(copyCollector.ContainerPath))}
+	command := []string{"tar", "-C", filepath.Dir(copyCollector.ContainerPath), "-cf", "-", filepath.Base(copyCollector.ContainerPath)}
 	req := client.CoreV1().RESTClient().Post().Resource("pods").Name(pod.Name).Namespace(pod.Namespace).SubResource("exec")
 	scheme := runtime.NewScheme()
 	if err := corev1.AddToScheme(scheme); err != nil {
--- a/pkg/collect/copy.go
+++ b/pkg/collect/copy.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"path"
 	"path/filepath"
 
 	troubleshootv1beta1 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta1"
@@ -14,8 +13,8 @@ import (
 	"k8s.io/client-go/tools/remotecommand"
 )
 
+//Copy function gets a file or folder from a container specified in the specs.
 func Copy(c *Collector, copyCollector *troubleshootv1beta1.Copy) (map[string][]byte, error) {
-
 	client, err := kubernetes.NewForConfig(c.ClientConfig)
 	if err != nil {
 		return nil, err
@@ -49,7 +48,7 @@ func Copy(c *Collector, copyCollector *troubleshootv1beta1.Copy) (map[string][]b
 			}
 
 			for k, v := range files {
-				copyOutput[filepath.Join(bundlePath, path.Dir(copyCollector.ContainerPath), k)] = v
+				copyOutput[filepath.Join(bundlePath, filepath.Dir(copyCollector.ContainerPath), k)] = v
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


divolgin:
@manavellamnimble this should use `tar` command to retrieve the entire folder (or file, if the path is a file).  There are two main reasons for this.  Using `tar` allows copying all data in a single API call, leaving less chance for an error.  But also we need to preserve file metadata: permissions, ownership, and timestamps.   Metadata should be applied to files when the archive is extracted.

marccampbell:
@divolgin what should we do in the event hat `tar` is not installed in the container?

matias-manavella-nimble:
@divolgin @marccampbell those are both good points. I can figure out a way to test if tar is installed (checking if running `tar --help` for example, returns an error).  If it is not installed we go with the find+cat function, which is working (or at least it seems to be).The thing is, in case we use tar, should we leave the compressed file in /namespace/pod.name/ filename.tar.gz?
I should also investigate a if it is possible to run multiple commands (or concatenate commands) using the k8s client.

I assume we are not considering windows-based containers, is that correct?

Let me know what you think!

divolgin:
@marccampbell it is possible for tar to not installed.  But I'd argue that this is such a rare configuration that we can resolve it with documentation.  Even alpine image comes with tar.

matias-manavella-nimble:
Ok, I am working on a solution using tar too, and making only one call to k8s client, I hope to have it working by tomorrow, and then you can tell me which of the two solutions would you rather implement. Does that sounds ok?

emosbaugh:
I would recommend "borrowing" from this code 

~https://github.com/kubernetes/kubernetes/blob/382107e6c84374b229e6188207ef026621286aa2/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L268-L273~

https://github.com/kubernetes/kubernetes/blob/382107e6c84374b229e6188207ef026621286aa2/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L315

matias-manavella-nimble:
@marccampbell @divolgin @emosbaugh I was able to extract the folder or files as .tar and place the tar file inside the support bundle under <namespace>/<pod.name>/<filename.tar>. Nevertheless, it doesn't work when the folder being copied contains an image. I tracked the problem to the redactor (If I skip the redaction stage, everything works ok). My question is, should I work in finding  way to preserve images? Or are images not important and I can exclude them from the tar?

emosbaugh:
There is a known issue that redaction appends new lines. I wonder if this is the issue with images?

matias-manavella-nimble:
@emosbaugh It might be, yes! But you made me realize, it is not posible to redact any file in a tar file. Based on what you say, the redactor must be making a modification in the bytes of the tar file, corrupting the file. 

I think...maybe extracting the files in the tar prior to redaction may work and have a double benefit. We only make one request to the client, and we can apply the redactors to the files, which may solve issue #58 too.

emosbaugh:
@laverya you may have more of an opinion on the last comment from @manavellamnimble

marccampbell:
I think the redactors should be taking archives into account. If a tar or zip or tgz is found in a bundle, the Redactor phase must decompress, redact and recommpress.

pkg/collect/copy.go
```diff
@@ -109,9 +115,59 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 		return nil, errors
 	}
 
-	return map[string][]byte{
-		copyCollector.ContainerPath: output.Bytes(),
-	}, nil
+	for _, file := range filepaths {
+		if file != "" {
```
marccampbell: This would be better written has 

```
if file == "" {
  continue
}
```

The code is less conditionally indented and more readable this way

marccampbell: @CodeLingoBot capture continue in conditional in loop

pkg/collect/copy.go
```diff
@@ -109,9 +115,59 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
...
+			command := []string{"cat", file}
+
+			req := client.CoreV1().RESTClient().Post().Resource("pods").Name(pod.Name).Namespace(pod.Namespace).SubResource("exec")
```
marccampbell: This is unnecessarily verbose.

```
req := client.Corev1().Pods(pod.Namespace).Get(pod.Name)
```

Is cleaner and how we use the clientset

marccampbell: @CodeLingoBot capture unnecessary use of client-go restClient

matias-manavella-nimble:
@marccampbell  Good Morning Marc! You are right, it is way better as you propose in both cases, thank you for the observation! 

I will work in the following process for the copy collector (@divolgin and @emosbaugh I think your opinion about the process would be helpful too):

1- Compress file or folder in the pod and copy it (1 request to the client to get all, already working)
2- Decompress the tar file and send it through Redactor. In case one of the files in the selected folder is a tar, decompress it and redact it too.
3- Compress all again and save it in the bundle under <namespace>/<pod.name>/<folder/file>.tar

Considering that originally the copy collector used cat to copy a file, I would think this would be a great improvement. Let me know what you think!

emosbaugh:
that looks good. would you include gzipped files as well?

matias-manavella-nimble:
Yes, I was thinking about choosing a decompressor/compresor according to the extension of the compressed file in the folder.

divolgin:
@manavellamnimble I think it's not necessary to re-compress everything.  If fact, it's probably better to leave the result as a folder structure matching the original folder in the pod.

matias-manavella-nimble:
@divolgin taking into account what you said about keeping the file info of the copied files (for example, permissions), I found that the best way to keep this info was to make a tar file. 
I have a version already working, where both the tar with the folder and any tar inside the folder are decompressed, redacted, and compressed back. The tar file is, as you said, matching the original folder in the pod. 

In the image, for example, I copied two folders (httpd and httpdocs) and one file (cheddar.yaml). **And every file inside each tar is redacted** (and also each file inside a tar inside the tar...tar-inception).

Of course it may be removed but I think it is a robust solution! First thing tomorrow I will commit and we can discuss the code! Let me know what you think!

Screen Shot 2020-08-11 at 7 55 28 PM

matias-manavella-nimble:
@marccampbell @divolgin @emosbaugh Morning! I've made a new commit. The upgrades are:

- It compress and  extract a folder or file in a tar, making just one call to the client. 
- The redaction algorithm take the tar file, and every tar file inside it, decompress it, redact each file, and compress all back together.
- Copied folder or file is placed compressed into the bundle matching the original folder path in the pod. In case of copying multiple folders from the same directory, both compressed folder are placed under the same path (no duplicate folders)
- As a plus, it implies no changes or modifications to the copy collector function, or how it is specified in the .yml file. 
- FileInfo (including permissions) is saved.
- In my case, the files inside each tar where redacted (when it comes to yaml files)

Possible upgrades:
- extend functionality to decompress other tar-ish files (gzip, tar.gz,...)
- avoid redacting images or fix bug if there were one (still not working)

**Please let me know any thought you have! Have a great day!**

As an example, the image represents the output of the following yaml file:
```
collectors:
    - copy:
        selector: 
          - app=myhttpd
        namespace: default
        containerPath: /usr/local/apache2/htdocs
    - copy:
        selector: 
          - app=myapp
        namespace: default
        containerPath: /go/src/app/k8s/Ejemplos/cheddar.yaml
    - copy:
        selector: 
          - app=myapp
        namespace: default
        containerPath: /go/src/app/httpd

```
Screen Shot 2020-08-11 at 7 55 28 PM

pkg/collect/copy.go
```diff
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"path"
+
```
emosbaugh: remove this newline

pkg/collect/redact.go
```diff
@@ -1,20 +1,75 @@
...
+			if err != io.EOF {
+				return nil, nil, err
+			}
+			break
+		}
+		if header.FileInfo().IsDir() {
+			continue
+		}
+		file := new(bytes.Buffer)
+		io.Copy(file, tarReader)
```
emosbaugh: handle this error

pkg/collect/redact.go
```diff
@@ -1,20 +1,75 @@
...
-			result[k] = redacted
+			files, err := redactMap(unRedacted, additionalRedactors)
+			if err != nil {
+				return nil, err
+			}
+			tw := tar.NewWriter(buff)
+			for p, f := range files {
+				//File size must be recalculated in case the redactor added some bytes when redacting.
+				fileHeaders[p].Size = int64(binary.Size(f))
+				tw.WriteHeader(fileHeaders[p])
```
emosbaugh: handle this error

pkg/collect/redact.go
```diff
@@ -1,20 +1,75 @@
...
+				tw.Write(f)
```
emosbaugh: handle this error

pkg/collect/redact.go
```diff
@@ -1,20 +1,75 @@
...
+			unRedacted, fileHeaders, err := untarFile(tarFile)
 			if err != nil {
 				return nil, err
 			}
-			result[k] = redacted
+			files, err := redactMap(unRedacted, additionalRedactors)
+			if err != nil {
+				return nil, err
+			}
+			tw := tar.NewWriter(buff)
```
emosbaugh: could you make this its own function `tarFiles`?

pkg/collect/copy.go
```diff
@@ -60,8 +63,9 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 	if copyCollector.ContainerName != "" {
 		container = copyCollector.ContainerName
 	}
-
-	command := []string{"cat", copyCollector.ContainerPath}
+	//Command cd into the path directory, tars the target file or folder into a temp file, cat-copy it and removes the temp file.
+	//Not using a temp file may end up in error "Refusing to write archive contents to terminal"
+	command := []string{"sh", "-c", fmt.Sprintf("tar -C %v  -cf tmp.tar %v; cat tmp.tar; rm tmp.tar", path.Dir(copyCollector.ContainerPath), path.Base(copyCollector.ContainerPath))}
```
emosbaugh: what if the directory is not writeable? when does it end up in the error? how does this work?
https://github.com/kubernetes/kubernetes/blob/382107e6c84374b229e6188207ef026621286aa2/staging/src/k8s.io/kubectl/pkg/cmd/cp/cp.go#L315

emosbaugh (changes requested):

pkg/collect/copy.go
laverya: If you're really worried about the 'will refuse to write to terminal' bits, just pipe it to `| cat` - 'cat' doesn't care about being a binary file, and tar just sees that it's not going to the terminal.

So `tar -cf %s | cat`.

(Incidentally, you should probably use `%s` and not `%v` - that way we'll see improper type coercions)

matias-manavella-nimble: @laverya That actually did the trick! 👏 Thank you also for the observation about the format verbs, it's a good point!
@emosbaugh For what I've read, it depends on the version of tar. Some of them consider printing the file to stdout a security risk and prevent it from happening.

pkg/collect/redact.go
matias-manavella-nimble: Yes, in fact I was considering that approach too. I chose not to  make a function to save some lines of code, and because I did not had to use it in other part of the code, but I could totally make a function which receives the files and headers, and return a tar file, thinking in using it later too. Do you prefer it that way?

pkg/collect/copy.go
matias-manavella-nimble: Done!

matias-manavella-nimble:
Hi everyone! I've made a new commit with the changes suggested by @emosbaugh and @laverya. At your disposal for any new changes you want to make!

pkg/collect/copy.go
```diff
@@ -47,7 +49,7 @@ func Copy(c *Collector, copyCollector *troubleshootv1beta1.Copy) (map[string][]b
 			}
 
 			for k, v := range files {
-				copyOutput[filepath.Join(bundlePath, k)] = v
+				copyOutput[filepath.Join(bundlePath, path.Dir(copyCollector.ContainerPath), k)] = v
```
emosbaugh: this should be filepath.Dir

pkg/collect/copy.go
```diff
@@ -110,7 +110,7 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 	}
 
 	return map[string][]byte{
-		copyCollector.ContainerPath: output.Bytes(),
+		path.Base(copyCollector.ContainerPath) + ".tar": output.Bytes(),
```
emosbaugh: this should be filepath.Base

pkg/collect/redact.go
```diff
@@ -1,20 +1,96 @@
...
+			return nil, err
 		}
+		result[k] = redacted
 	}
 	return result, nil
 }
+
+func tarFiles(files map[string][]byte, fileHeaders map[string]*tar.Header) ([]byte, error) {
+	buff := new(bytes.Buffer)
+	tw := tar.NewWriter(buff)
```
emosbaugh: can you `defer tw.Close()` here to make sure its closed even on an error?

pkg/collect/redact.go
```diff
@@ -1,20 +1,96 @@
...
+		err = tw.WriteHeader(fileHeaders[p])
+		if err != nil {
+			return nil, err
+		}
+		_, err = tw.Write(f)
+		if err != nil {
+			return nil, err
+		}
+	}
+	err = tw.Close()
```
emosbaugh: this should be moved to a defer above

pkg/collect/redact.go
```diff
@@ -1,20 +1,96 @@
...
+	files := make(map[string][]byte)
+	for {
+		header, err := tarReader.Next()
+		if err != nil {
+			if err != io.EOF {
+				return nil, nil, err
+			}
+			break
+		}
+		if header.FileInfo().IsDir() {
```
emosbaugh: what happens if you re-compress the tar without the dirs?

pkg/collect/redact.go
```diff
@@ -1,20 +1,96 @@
...
+			continue
+		}
+		file := new(bytes.Buffer)
+		_, err = io.Copy(file, tarReader)
+		if err != nil {
+			return nil, nil, err
+		}
+		files[header.Name] = file.Bytes()
```
emosbaugh: and maps are not ordered. the tar will now be in a different order. im not sure how that affects things.

emosbaugh (changes requested):

pkg/collect/redact.go
matias-manavella-nimble: Actually folders in tar are just the header with the Dir info. Without that info when you decompress the tar, folders described in file paths of the files inside the tar are created with default perm. It is though an avoidable loss of info! I already did the modifications to keep it

pkg/collect/redact.go
matias-manavella-nimble: If I am not mistaken, I think order matters for very specific applications, for example when you want to give access to decompressed files while simultaneously decompressing the other ones in the tar. For the copy collector I would think order is not really important

pkg/collect/redact.go
matias-manavella-nimble: Yes, definitely. It is way more robust in the way you suggest

pkg/collect/copy.go
matias-manavella-nimble: Yep, indeed. Will change it!

pkg/collect/redact.go
matias-manavella-nimble: I think it would be better to leave this Close() too, in order to handle any closing error if there is one. Calling Close() twice (in this line and after that, when defer executes) wont cause any problem, and we still handle the error. Do you agree?

matias-manavella-nimble:
@emosbaugh I've made the changes we discussed!

pkg/collect/redact.go
```diff
@@ -1,20 +1,101 @@
...
+		err := tw.WriteHeader(tarHeaders[p])
+		if err != nil {
+			return nil, err
+		}
+		_, err = tw.Write(f)
+		if err != nil {
+			return nil, err
+		}
+	}
+	err := tw.Close()
```
emosbaugh, line 81: can you remove this

emosbaugh (changes requested):

emosbaugh (approved):

pkg/collect/redact.go
emosbaugh: ive never seen that pattern but im ok with it

pkg/collect/copy.go
matias-manavella-nimble: Hi Marc! I didn't have much contact with the client set yet, so I left that line as it was. The way you suggest returns another type of variable and requires another type of variable too instead of pod.Name. I didn't investigate it further because I am sure I will have to deal with the clientset in later PR's, and I didn't wanted to delay the copy collector. 
That said I can either change it later, once I've dealt with the client set, or look into it now, as you prefer!

laverya (approved):

pkg/collect/copy.go
```diff
@@ -60,9 +61,7 @@ func copyFiles(c *Collector, client *kubernetes.Clientset, pod corev1.Pod, copyC
 	if copyCollector.ContainerName != "" {
 		container = copyCollector.ContainerName
 	}
-
-	command := []string{"cat", copyCollector.ContainerPath}
-
+	command := []string{"sh", "-c", fmt.Sprintf("tar -C %s  -cf - %s | cat ", filepath.Dir(copyCollector.ContainerPath), filepath.Base(copyCollector.ContainerPath))}
```
divolgin: Why does it need to pipe into `cat`?   Can the pipe be removed and then `sh` is not needed, so it's just the `tar` command by itself.

pkg/collect/redact.go
```diff
@@ -1,20 +1,101 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return buff.Bytes(), nil
+
+}
+
+func untarFile(tarFile *bytes.Buffer) (map[string][]byte, map[string]*tar.Header, error) {
+	tarReader := tar.NewReader(tarFile)
```
divolgin: defer close

pkg/collect/redact.go
```diff
@@ -1,20 +1,101 @@
...
+			}
+			result[k], err = tarFiles(redacted, tarHeaders)
 			if err != nil {
 				return nil, err
 			}
-			result[k] = redacted
+			//Content of the tar file was redacted. Continue to next file.
+			continue
+		}
+		redacted, err := redact.Redact(v, k, additionalRedactors)
```
divolgin, line 41: This can corrupt `tgz` and `tar.gz`  files.

divolgin (changes requested):

pkg/collect/copy.go
matias-manavella-nimble: Hi Dimitriy! In some cases, depending on the version of tar installed in the container, we may get the error: 
tar: Refusing to write archive contents to terminal (missing -f option?)
tar: Error is not recoverable: exiting now

Talking about this matter with Andrew and Ethan we came to the conclusion that piping it would be the best way to avoid the error.

pkg/collect/redact.go
matias-manavella-nimble: Are you refering to the tar reader? I think it has not a Close() function. I've used defer for the tarWriter

pkg/collect/redact.go
matias-manavella-nimble: Yes! Only when the file is inside the folder. I suggested a possible upgrade to the code to decompress .tgz and .tar.gz files too. If you think it would be useful I can work in this functionality!

emosbaugh: we should address this in a separate pr if we choose to

matias-manavella-nimble:
@divolgin @emosbaugh Hi there! I've added the possibility to decompress and redact files from .tar.gz and .tgz files. It works on nested tar-ish files as well. I tried to use the same functions, to keep it simple.
Tomorrow is holiday in Argentina, and I thought about committing the code today so you can see it tomorrow morning!

divolgin:
@manavellamnimble does this copy collector leave folders as folders in the support bundle, or are they in tar format?

matias-manavella-nimble:
@divolgin The target folder (lets say, folder httpd in go/src/app/httpd) is place as tar in the bundle under <namespace>/<pod_name>/ go/src/app/htrpd.tar as tar. Every content in the tar is left as it was (folders as folders, files as files and compressed files (tar, tar.gz and tgz) as compressed files). 
That way we not only copy the folder in its entirety, but also keep all the info of it contents, and of the target-folder itself. Image as example, with the target folder decompressed: 
Screen Shot 2020-08-18 at 10 28 45 AM

pkg/collect/redact.go
```diff
@@ -1,20 +1,129 @@
...
+
+}
+
+func decompressFile(tarFile *bytes.Buffer, filename string) (map[string][]byte, map[string]*tar.Header, error) {
+	var tarReader *tar.Reader
+	var zr *gzip.Reader
+	var err error
+	if filepath.Ext(filename) != ".tar" {
+		zr, err = gzip.NewReader(tarFile)
+		defer zr.Close()
```
emosbaugh: this defer should come after checking the err != nil

matias-manavella-nimble: Yep, you are right! Done

emosbaugh (approved):

divolgin:
@manavellamnimble files and folders in the support bundle should match the source.  So in your example, `cheddar.yaml.tar` should be plain `cheddar.yaml` text file.  And `httpd.tar` should be `httpd` folder with the same structure as in the container.

matias-manavella-nimble:
@divolgin Hi Dimitriy! After considering many approaches, I think this is the cleaner one. It requires almost no modifications, and it doesnt spread changes through the application. It stores folders and files in the support bundle, keeping the permissions and matching the path in origin. 

It maintains all the upgrades developed until this point. 

I've tested it copying two folders (httpd and htdocs) and  a file (k8s/ingress.yaml), see image below:

Screen Shot 2020-08-19 at 4 29 14 PM

Issue #159, opened 2020-03-20
Should NodeResources analyzer allow filtering by Label?


dexhorthy:
Right now, NodeResources allows filtering by resources, but not by label:

https://github.com/replicatedhq/troubleshoot/blob/d9625477d2a552f7b5dc1289449527704fea1a95/pkg/apis/troubleshoot/v1beta1/analyzer_shared.go#L81-L96

### Use Case

I want to be able to require a certain number of nodes with certain labels as a preflight check, e.g. to run the Mongo Operator, the implementation today uses node labels as a scheduling requirement. 

https://github.com/jasonmimick/mongodb-k8s/blob/master/templates/ops-manager.yaml#L20-L27

Furthermore, the requirements I present for memory and CPU should only be applied to the nodes that are schedulable for my application, so filters seems like the right place to add this (rather than, say, a separate analyzer spec)

### Examples

Option 1, `matchLabels`

```yaml
    - nodeResources:
        checkName: Must have 1 node with 16 GB (available) memory and 5 cores (on a single node) with label kubernets.io/role=database-primary-replica
        filters:
          allocatableMemory: 16Gi
          cpuCapacity: "5"
          selector:
            matchLabels: 
               kubernetes.io/role=database-primary-replica
        outcomes:
          - fail:
              when: "count() < 1"
              message: Must have 1 node with 16 GB (available) memory and 5 cores (on a single node) with label kubernets.io/role=database-primary-replica
          - pass:
              message: This cluster has a node with enough memory.
```

Option 2, use the exact syntax for nodeSelectors from pod spec

```yaml
          - matchExpressions:
            - key: kubernetes.io/role 
              operator: In
              values:
              - database-primary-replica 
```

matias-manavella-nimble:
@marccampbell @dexhorthy Hi guys, I am working on the final details of this feature, I followed option 1 of dex's suggestion, but I used a dictionary instead of a string value, like this. If you think it's ok, I send the PR tomorrow morning

```yaml
filters:
  cpuCapacity: "6"
  selector: 
     matchLabel: 
         kubernetes.io/hostname: docker-desktop
```

matias-manavella-nimble:
@marccampbell @dexhorthy  Morning Marc, Dex! I think this issue may be closed!

PR #249, opened 2020-08-28
selector:MatchLabel

manavellamnimble:
@marccampbell @divolgin @dexhorthy Morning! I am sending a PR to solve issue #159 . I desing the selector label to accept further expansions in the future. Under matchLabel, multiple labels can be used to filter the nodes. The final structure would be something like this:
```YAML
- nodeResources:
        checkName: Must have at least 3 nodes in the cluster, with cpuCapacity of 2 and label kubernetes.io/hostname=docker-desktop
        filters:
            cpuCapacity: "2"
            selector: 
                matchLabel: 
                    kubernetes.io/hostname  : docker-desktop
        outcomes:
        - fail:
            when: "count() < 3"
            message: This application requires at least 3 nodes.
            uri: https://kurl.sh/docs/install-with-kurl/adding-nodes
        - warn:
            when: "count() < 5"
            message: This application recommends at last 5 nodes.
            uri: https://kurl.sh/docs/install-with-kurl/adding-nodes
        - pass:
            message: This cluster has enough nodes.
```

### DIFF
--- a/pkg/analyze/node_resources.go
+++ b/pkg/analyze/node_resources.go
@@ -319,6 +319,13 @@ func nodeMatchesFilters(node corev1.Node, filters *troubleshootv1beta1.NodeResou
 	}
 
 	// all filters must pass for this to pass
+	if filters.Selector != nil {
+		for k, v := range filters.Selector.MatchLabel {
+			if l, found := node.Labels[k]; !found || l != v {
+				return false, errors.New(fmt.Sprintf("failed to match label %s", k))
+			}
+		}
+	}
 
 	if filters.CPUCapacity != "" {
 		parsed, err := resource.ParseQuantity(filters.CPUCapacity)
--- a/pkg/analyze/node_resources.go
+++ b/pkg/analyze/node_resources.go
@@ -322,7 +322,7 @@ func nodeMatchesFilters(node corev1.Node, filters *troubleshootv1beta1.NodeResou
 	if filters.Selector != nil {
 		for k, v := range filters.Selector.MatchLabel {
 			if l, found := node.Labels[k]; !found || l != v {
-				return false, errors.New(fmt.Sprintf("failed to match label %s", k))
+				return false, errors.Errorf("failed to match label %s", k)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


marccampbell:
@manavellamnimble the `matchLabel` dictionary feels a little foreign to me, compared to the rest of the Kubernetes API.

If we can match the format of Kubernetes native objects for label selectors (https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) it will be easier for people to implement because we won't be asking them to change the format and they can re-use existing YAML selectors.

Is is possible to use the standard key=value types here, and use the `selector` package from `client-go` to parse so that we match the format and results the same?

matias-manavella-nimble:
@marccampbell In fact I used a map based on what you mentioned. K8s api uses mostly maps (I called it dictionary, my bad). In the docs there is an example for selectors, following the same structure:
```YAML
selector:
  matchLabels:
    component: redis
```
```matchLabels is a map of {key,value} pairs.```
src: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

Regarding using the client, that was also my first approach, but the matchFilter() func uses the info stored in the file nodes.json (created earlier by the collector), thats why I use node.Labels (which is also a map). 

Nevertheless, I can adjust the labels as you prefer, it is really not lots of work. Let me know what you think!

File: pkg/analyze/node_resources.go [hunk 1]
```diff
@@ -319,6 +319,13 @@ func nodeMatchesFilters(node corev1.Node, filters *troubleshootv1beta1.NodeResou
 	}
 
 	// all filters must pass for this to pass
+	if filters.Selector != nil {
+		for k, v := range filters.Selector.MatchLabel {
+			if l, found := node.Labels[k]; !found || l != v {
+				return false, errors.New(fmt.Sprintf("failed to match label %s", k))
```
marccampbell: `errors.Errorf`

marccampbell:
Thanks! This lgtm, 1 comment on cleaning up an error string before we can merge

matias-manavella-nimble:
@marccampbell Great! Done!

PR #250, opened 2020-08-31
Check for Default Storage Class when storageClassName is empty

### LINKED ISSUE #200: Analyzer for default storage class?
Right now we can check for a storage class with a given name, but what if I just want to check for a storage class that has the `isDefaultStorageClass` field set to true? In most cases I don't care what the end user is calling their storage class and I want to reduce friction by leaving `storageClassName` empty in my PVCs and StatefulSets, and just use a preflight check to ensure there is a default storageClass to provision PVs.

manavellamnimble:
@marccampbell @dexhorthy Good morning! I send you the PR to check for a storage class set as default when storageClassName is empty. I can add the parse function to parse the string "true" to bool, but it would also consider other  values as "True", "T", "TRUE"...and I think k8s only expect  the value ```true```
```
The default StorageClass has an annotation storageclass.kubernetes.io/is-default-class set to true. Any other value or absence of the annotation is interpreted as false.
```
Closes #200 

src: https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/#changing-the-default-storageclass (item 2)

### DIFF
--- a/pkg/analyze/storage_class.go
+++ b/pkg/analyze/storage_class.go
@@ -31,7 +31,8 @@ func analyzeStorageClass(analyzer *troubleshootv1beta1.StorageClass, getCollecte
 	}
 
 	for _, storageClass := range storageClasses {
-		if storageClass.Name == analyzer.StorageClassName {
+		val, found := storageClass.Annotations["storageclass.kubernetes.io/is-default-class"]
+		if (storageClass.Name == analyzer.StorageClassName) || (analyzer.StorageClassName == "" && found && val == "true") {
 			result.IsPass = true
 			for _, outcome := range analyzer.Outcomes {
 				if outcome.Pass != nil {
--- a/pkg/analyze/storage_class.go
+++ b/pkg/analyze/storage_class.go
@@ -21,7 +21,11 @@ func analyzeStorageClass(analyzer *troubleshootv1beta1.StorageClass, getCollecte
 
 	title := analyzer.CheckName
 	if title == "" {
-		title = fmt.Sprintf("Storage class %s", analyzer.StorageClassName)
+		if analyzer.StorageClassName != "" {
+			title = fmt.Sprintf("Storage class %s", analyzer.StorageClassName)
+		} else {
+			title = "Default Storage class"
+		}
 	}
 
 	result := AnalyzeResult{
@@ -40,6 +44,9 @@ func analyzeStorageClass(analyzer *troubleshootv1beta1.StorageClass, getCollecte
 					result.URI = outcome.Pass.URI
 				}
 			}
+			if analyzer.StorageClassName == "" && result.Message == "" {
+				result.Message = "Default Storage found"
+			}
 
 			return &result, nil
 		}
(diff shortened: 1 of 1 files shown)


dexhorthy:
Hi @manavellamnimble thanks for the PR! I think I'm in favor of only supporting exactly what Kubernetes supports (or what they document to support) -- if adding an annotation value of (string) `T` or `yes` causes the `kubectl get storageclass` output to change, then we should treat that as valid:

```
NAME             PROVISIONER               AGE
standard         kubernetes.io/gce-pd      1d
gold (default)   kubernetes.io/gce-pd      1d
```

But if Kubernetes only parses the `"true"` string for this then we should only allow that. Does that make sense?

pkg/analyze/storage_class.go
```diff
@@ -31,7 +31,8 @@ func analyzeStorageClass(analyzer *troubleshootv1beta1.StorageClass, getCollecte
 	}
 
 	for _, storageClass := range storageClasses {
-		if storageClass.Name == analyzer.StorageClassName {
+		val, _ := storageClass.Annotations["storageclass.kubernetes.io/is-default-class"]
+		if (storageClass.Name == analyzer.StorageClassName) || (analyzer.StorageClassName == "" && val == "true") {
```
dexhorthy, line 39: so the way I'm reading this, you could do

```
storageClass:
  name: "gold"
```

to check that a storage class exists with the name `gold`, or you could do 

```
storageClass: {}
```

to check that any default storage class exists. @markpundsack I'm okay with this overloading, but does that work for you, or do you want to do a separate analyzer `hasDefaultStorageClass` or something?

matias-manavella-nimble: Yes, exactly! When a name is specified, it doesn't check the is-default flag, and when no name is provided, it checks if there is at least one default sc. 
I could add a new optional field in the analyzer storageClass specs to check if a storage class with a given name is also marked as default. Something like:
```YAML
- storageClass:
        storageClassName: "gold"
        isDefault: true/false 
```

markpundsack: I'm OK with overloading since specifying `storageClass: {}` doesn't really have any other reasonable interpretation.

I don't think it makes sense to check if a specific storage class is default or not.

matias-manavella-nimble: @markpundsack @dexhorthy I could add a default fail/pass message in absence of an user-defined one. If storageClass were used as ```storageClass: {}``` add a message "There is/is not a default storage class". In addition a more explicative checkName could also be add, something like "Storage Class Marked as Default". What do you think?

markpundsack: Adding default message and name makes sense. Not sure about the wording. "Storage Class Marked as Default" is a bit misleading. It's more like, "Default storage class" and "No default storage class found".

matias-manavella-nimble:
@markpundsack @dexhorthy I've added default messages in case storageClassName is empty and no message is provided.

pkg/analyze/storage_class.go
markpundsack: Messages look good.

laverya (approved):

sgalsaleh (approved):

PR #252, opened 2020-09-03
Fix statefulset status

marccampbell:
The range would still override status


divolgin (approved):

pkg/analyze/deployment_status.go
```diff
@@ -24,7 +24,7 @@ func analyzeDeploymentStatus(analyzer *troubleshootv1beta2.DeploymentStatus, get
 	var status *appsv1.DeploymentStatus
 	for _, deployment := range deployments {
 		if deployment.Name == analyzer.Name {
-			status = &deployment.Status
```
marccampbell, line 27: @codelingobot Prevent reference assignment of local variable in a loop to a variable used outside of the loop

Issue #247, opened 2020-08-19
Support passing an imagePullSecret in a run collector


marccampbell:
A run collector can only use public images today. It would be useful if the run collector schema supported an optional imagePullSecret to be added to the podSpec when creating. 

Additionally, it would be useful if the run collector could include the secret data (useful to templated and generated credentials). In this case, the run collector in troubleshoot would have to dynamically create the secret in the namespace (giving it a random name), and then using that generated secret name as the imagePullSecret in the podSpec. The secret should be deleted when completed.

matias-manavella-nimble:
@marccampbell Hi Marc, I'm about to start working on this. I am thinking of defining a new set of specs for the run collector as follows:

A) If pod is going to use an existing secret: 
```YAML
- run:
        imagePullSecret:
            name: secrets_name
        collectorName: "run-ping"
        ...
```
B) If instead we want to create the secret (I used the example form: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/):
```YAML
- run:
        imagePullSecret:
            name: <random_if_not_specified>
            data: 
               .dockerconfigjson: UmVhbGx5IHJlYWxseSByZWVlZWVlZ...
            type: kubernetes.io/dockerconfigjson
        collectorName: "run-ping"
        ...
```

dexhorthy:
@marccampbell I think I care less about inline-provided creds for most of my use cases, although I see the value of that for standalone troubleshoot use-cases. It sounds like from the Issue that you wanna do both, but I care mostly about what @manavellamnimble has outlined in `A`

marccampbell:
thanks @dexhorthy There's probably value in (b) also though, since a KOTS app is not deployed when preflights run.  Having the ability to template an imagePullSecret with `.dockerconfigjson; repl{{ LicenseDockerCfg }}` feels like it will be necessary.

matias-manavella-nimble:
@dexhorthy @marccampbell Great! Yes, the idea is to add both, and make the collector able to know which is the use case based on the fields completed under imagePullSecrets (thus avoiding using separate specs). I'll send the PR as soon as it's ready

matias-manavella-nimble:
Hi! This issue may be closed

PR #253, opened 2020-09-04
Add or create ImagePullSecret  in a run collector 

manavellamnimble:
@marccampbell @dexhorthy regarding issue #247   I added the possibility to either add an existing secret :

```yaml
- run
     ...
     imagePullSecret:
         name: mysecret
```
or to create a secret. In this case there are various options:
- add an opaque secret, with one or many data fields in it
- optionally add a name to the secret. If none is provided a random name is generated 
- provide a type kubernetes.io/dockerconfigjson secret, in which case there are two options
   - provide the path to the config file. The client parses and encodes it automatically.
   - provide an already base64-encoded configfile. Though it is not actually an option when using the k8s client, I dealt with this because it is the suggested way in kubernetes official docs (but it only works when using yaml). What I did was to decode the string, indent it to rebuild the json, and put it into the secret. Base64 does not takes into consideration white spaces or tabs, making this algorithm independent of  the original config file's indentation. All this thinking in making it more familiar to the user.

I handle the secret's data as []byte because it is safer according to what I read.
```yaml
- run
      ...
      imagePullSecret:
          name: mysecret
          data: 
            .dockerconfigjson: path/.docker/config.json
          type: kubernetes.io/dockerconfigjson
```
```yaml
- run
      ...
      imagePullSecret:
          name: mysecret
          data: 
            .dockerconfigjson: woJICJhdXRocyI6IHsKCQkiaHR0cHM6Ly9pbmRleC5kb2NrZXIuaW8vdjEvIjoge30KCX0sCgkiSHR0cEhlYWRlcnMiOiB7CgkJIlVzZXItQWdlbnQiOiAiRG9ja2VyLUNsaWVudC8xOS4wMy4xMiAoZGFyd2luKSIKCX0sCgkiY3JlZHNTdG9yZSI6ICJkZXNrdG9wIiwKCSJleHBlcmltZW50YWwiOiAiZGlzYWJsZWQiLAoJInN0YWNrT3JjaGVzdHJhdG9yIjogInN3YXJtIgp9
          type: kubernetes.io/dockerconfigjson
```
                         

```yaml
- run
      ...
      imagePullSecret:
          data: 
            a: foo
            b: bar
```
Let me know what you think of the solution!

### DIFF
--- a/pkg/collect/run.go
+++ b/pkg/collect/run.go
@@ -183,7 +183,7 @@ func createSecret(ctx context.Context, client *kubernetes.Clientset, imagePullSe
 				//Client only accepts Json formated files as data, so we decode and indent it (indentation is required)
 				parsedConfig, err := base64.StdEncoding.DecodeString(v)
 				if err != nil {
-					return errors.Wrap(err, "Secret's config file not found or unable to parse encoded data.")
+					return errors.Wrap(err, "Secret's config file not found or unable to decode data.")
 				}
 				err = json.Indent(&out, parsedConfig, "", "\t")
 				if err != nil {
--- a/pkg/collect/run.go
+++ b/pkg/collect/run.go
@@ -152,7 +152,7 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
 	if runCollector.ImagePullSecret != nil {
 		err := createSecret(ctx, client, runCollector.ImagePullSecret, &pod)
 		if err != nil {
-			return nil, err
+			return nil, errors.Wrap(err, "failed to create secret")
 		}
 	}
 	created, err := client.CoreV1().Pods(namespace).Create(ctx, &pod, metav1.CreateOptions{})
@@ -183,11 +183,11 @@ func createSecret(ctx context.Context, client *kubernetes.Clientset, imagePullSe
 				//Client only accepts Json formated files as data, so we decode and indent it (indentation is required)
 				parsedConfig, err := base64.StdEncoding.DecodeString(v)
 				if err != nil {
-					return err
+					return errors.Wrap(err, "Secret's config file not found or unable to parse encoded data.")
 				}
 				err = json.Indent(&out, parsedConfig, "", "\t")
 				if err != nil {
-					return errors.Errorf("Secret's config file not found or unable to parse encoded data.")
+					return errors.Wrap(err, "Unable to parse encoded data.")
 				}
 				data[".dockerconfigjson"] = out.Bytes()
 			} else {
@@ -205,7 +205,7 @@ func createSecret(ctx context.Context, client *kubernetes.Clientset, imagePullSe
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name:         imagePullSecret.Name,
-				GenerateName: "tmpsecret",
+				GenerateName: "troubleshoot",
 				Namespace:    pod.Namespace,
 			},
 			Data: data,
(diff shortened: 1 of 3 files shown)


pkg/collect/run.go
```diff
@@ -137,10 +149,79 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
 		},
 	}
 
+	if runCollector.ImagePullSecret != nil {
+		err := createSecret(ctx, client, runCollector.ImagePullSecret, &pod)
+		if err != nil {
+			return nil, err
```
divolgin: needs `errors.Wrap`

pkg/collect/run.go
```diff
@@ -137,10 +149,79 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
...
+			if err != nil {
+				//If data is not a valid path, we assume data is a base64 encoded config.json file
+				//Client only accepts Json formated files as data, so we decode and indent it (indentation is required)
+				parsedConfig, err := base64.StdEncoding.DecodeString(v)
+				if err != nil {
+					return err
+				}
+				err = json.Indent(&out, parsedConfig, "", "\t")
+				if err != nil {
+					return errors.Errorf("Secret's config file not found or unable to parse encoded data.")
```
divolgin: needs errors.Wrap with the original error

pkg/collect/run.go
```diff
@@ -137,10 +149,79 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
...
+				return errors.Errorf("Secret type kubernetes.io/dockerconfigjson accepts only one argument \".dockerconfigjson\"")
+			}
+			//Then, if data is a path to a config file, it is opened
+			configFile, err := ioutil.ReadFile(v)
+			if err != nil {
+				//If data is not a valid path, we assume data is a base64 encoded config.json file
+				//Client only accepts Json formated files as data, so we decode and indent it (indentation is required)
+				parsedConfig, err := base64.StdEncoding.DecodeString(v)
+				if err != nil {
+					return err
```
divolgin: `errors.Wrap`

pkg/collect/run.go
```diff
@@ -137,10 +149,79 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
...
+			}
+		}
+		secret := corev1.Secret{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "Secret",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:         imagePullSecret.Name,
+				GenerateName: "tmpsecret",
```
divolgin: let's change `tmpsecret` to `troubleshoot`

pkg/collect/run.go
```diff
@@ -137,10 +149,79 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
...
+				Namespace:    pod.Namespace,
+			},
+			Data: data,
+			Type: corev1.SecretType(imagePullSecret.SecretType),
+		}
+		created, err := client.CoreV1().Secrets(pod.Namespace).Create(ctx, &secret, metav1.CreateOptions{})
+		if err != nil {
+			return err
```
divolgin: `errors.Wrap`

matias-manavella-nimble:
@divolgin Good morning Dimitriy! I made the modifications you suggested.

laverya:
Why not just add an entire pod or container spec here, instead of adding each component one by one?

matias-manavella-nimble:
@laverya It could be done, we could even add the possibility to use a link to a yaml file if necessary. 

Should I start working on that?

laverya:
no, I think that would be a seperate PR
let me give this a review

PR #259, opened 2020-09-10
fixEmptyKeySecret

### LINKED ISSUE #258: Secret collector always has empty key in result
Running the following Secret collector returns the json result below with key an empty string.

```yaml
kind: Collector
metadata:
  name: sample
spec:
  collectors:
    - secret:
        namespace: default
        name: my-secret
        key: password
        includeValue: true
```

```json
{
  "namespace": "default",
  "name": "my-secret",
  "key": "",
  "secretExists": true,
  "keyExists": true,
  "value": "mypass"
}
```

manavellamnimble:
@areed Hi Andrew, the 'key' field was missing in the foundsecret struct. I ran it locally and got:

```json
{
  "namespace": "default",
  "name": "mysec",
  "key": ".dockerconfigjson",
  "secretExists": true,
  "keyExists": true,
  "value": "{...}"
}
```

Closes #258

### DIFF
--- a/pkg/collect/secret.go
+++ b/pkg/collect/secret.go
@@ -83,6 +83,7 @@ func secret(ctx context.Context, client *kubernetes.Clientset, secretCollector *
 	secret := FoundSecret{
 		Namespace:    found.Namespace,
 		Name:         found.Name,
+		Key:          secretCollector.Key,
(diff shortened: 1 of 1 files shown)


manavellamnimble:
Closes #258

PR #268, opened 2020-09-17
Error Parsing Postgres Version 

### LINKED ISSUE #254: Postgres analyzer fails to parse postgres version
If you run the Postgres [support-bundle example](https://troubleshoot.sh/preflight/supported-postgres-version) against Postgres 10, the bundle includes the file `postgres/postgresql.json` with these contents:

```json
{"isConnected":true,"version":"10.14"}
```

The analyzer logs container the error `an analyzer failed to run: failed to compare postgres database conditional: failed to parse actual psotgres version: No Major.Minor.Patch elements found`

manavellamnimble:
@areed Hi! Regarding issue #254, the problem was that the semver version used only accepts version format major.minor.patch (eg. 10.2.3). This PR fix #254 

According to Postgres Versioning Policy, starting with version 10 on the version string will only have 2 parts, major.minor, that would make possible to simply add a 0 after the versions entered by the user and the one collected, but I think this solution is more robust.

I tested it against a dockerized Postgres DB and worked just fine.

### DIFF
--- a/pkg/analyze/database_shared.go
+++ b/pkg/analyze/database_shared.go
@@ -35,6 +35,18 @@ func compareDatabaseConditionalToActual(conditional string, result *collect.Data
 		return false, errors.New("unable to parse postgres connected analyzer")
 
 	case "version":
+		//semver requires major.minor.patch format to successqfully compare versions.
+		if compVer := strings.Split(parts[2], "."); len(compVer) == 2 {
+			parts[2] = fmt.Sprintf("%s.%s.0", compVer[0], compVer[1])
+		} else if len(compVer) == 1 {
+			parts[2] = fmt.Sprintf("%s.0.0", compVer[0])
+		}
+		if compVer := strings.Split(result.Version, "."); len(compVer) == 2 {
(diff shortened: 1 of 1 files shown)


pkg/analyze/database_shared.go
```diff
@@ -35,14 +35,26 @@ func compareDatabaseConditionalToActual(conditional string, result *collect.Data
 		return false, errors.New("unable to parse postgres connected analyzer")
 
 	case "version":
+		//semver requires major.minor.patch format to successfully compare versions.
+		if compVer := strings.Split(parts[2], "."); len(compVer) == 2 {
```
divolgin: Can we make use of [ParseTolerant](https://godoc.org/github.com/blang/semver/v4#ParseTolerant) function here instead of doing all this manually?

matias-manavella-nimble: I don't know how could I have missed it. The only thing with that function is that it doesn't recognize 'x' as a version wildcard. Most algorithms take the x as 0 when comparing, so I used Replace() to replace 'x' for '0'. Maybe it is not entirely necessary when parsing the actual version, but I think it wont do any harm to left it there. I could remove it if you want.

PR #272, opened 2020-09-30
Support loading spec from a secret

divolgin:


cmd/troubleshoot/cli/run.go
```diff
@@ -258,6 +267,47 @@ func loadSpec(v *viper.Viper, arg string) ([]byte, error) {
...
+		return nil, errors.Wrap(err, "failed to get spec from URL")
+	}
+	return spec, nil
+}
+
+func loadSpecFromSecret(v *viper.Viper, arg string) ([]byte, error) {
+	// format secret/namespace-name/secret-name
+	pathParts := strings.Split(arg, "/")
+
+	if len(pathParts) != 3 {
```
marccampbell: can this be in the caller, so that if it hits this code only if there are 3 parts?

marccampbell (approved):

PR #279, opened 2020-10-08
Avoid redaction in specific collectors

### LINKED ISSUE #262: ImagePullSecret analyzer fails with IP registryName
The ip is redacted before the analyzer runs.

manavellamnimble:
@areed @divolgin Hi Andrew, Dimitriy. Regarding #262, I think there is not an easy way to fix it. I've tried many approaches, and this, I think,is the cleanest one. I added two things:
- Pass analyzer specs to the redactor, for saving un-redacted files if needed. 
- Created a new variable, `protected`, which stores the un-redacted files separated from the results. This prevents not redacted data to be stored in the host, even as temporary files. 

This modifications would also be helpful in the future if there were a similar issue with another analyzer.

There are a couple of things that may be discussed, for example pasing `protected` as a pointer to the functions instead of returning it as a variable or avoid passing analyzers specs to the redactor and simply separating unredacted files whether there is an analyzer which will use it or not, but I would like you to check this approach before continuing working on this. 
Fixes #262.


xavpaice:
I think given the age of this PR, plus that it is clearer to ensure that the redactors specified redact in a predictable way (selectors for files, or all files), we can close this now.

PR #281, opened 2020-10-11
Run pod on all nodes

Arau:
Run collector: One Pod per node when using  Node Selectors

When node selectors are defined in a run collector, one Pod is created for each node that satisfies the nodeSelector terms.

In case of needing a Pod to run in one Node specifically, the user must define a nodeSelector that is only fulfilled by one node.

Logs from the Pod are gathered in one action, hence a timeout defined in the run collector applies for the gathering of data from all Pods.

Pods that can't start for entropy reasons, taints, cordons, etc are skipped.


Arau:
This PR is intended to go towards a personal fork of the repo as a proof of concept for a possible future contribution. The changes are an exploration of functionality, hence code is not ready or even proven to be of use. 

I'm closing the PR for the moment, I apologise for the inconvenience.

PR #286, opened 2020-10-16
fixed runCollector filepath

### LINKED ISSUE #189: Run collector and text analyzer don't work well together
Naming of collector output is unexpected, and doesn't match up well with what text analyzer expects. In particular https://troubleshoot.sh/reference/analyzers/regex/ is very wrong.

What I've found digging through the source is that the `Run` collector creates a file key for its output with `fmt.Sprintf("%s/%s.log", run.Name, run.CollectorName)` (this is the result of names working through creating a pod, calling `getPodLogs`, etc), while the text analyzer expects a name of `filepath.Join(textAnalyzer.CollectorName, textAnalyzer.FileName)`.

The only way I got this to work was with
```
collectors:
- run:
    name: foo
    collectorName: foo
    image: postgres:11.7-alpine
    command: ["echo"]
    args: ["hello"]
analyzers:
- textAnalyze:
    collectorName: foo
    fileName: foo.log
    regex: hello
    outcomes:
    - pass:
        message: "It worked!"
    - fail:
        message: "It didn't!"
```

manavellamnimble:
@laverya @dexhorthy @divolgin The use of the run collector together with  the text analyzer, the filepath handling and the fields of the run collector where not entirely clear. I will modify the docks once the PR is merged. I did some changes:

- `CollectorName` field was used as the pod's name and the `name` field was used as collector name, preceding the file path.
- Changed the `name` field to `PodName` and used it as the pod's name.
- Log file is stored under `runCollector.CollectorName/runCollector.PodName.log`, previously it was stored under `runCollector.Name/runCollector.CollectorName.log`. This change makes it compatible with the text analyzer. 
- As CollectorName and PodName are optional, I made PodName default to the image name, and collector name default to "run-collector-pod-"+PodName.

This change is only noticeable when using the text analyzer,  which will now work in a more logical way now, making it also easier to use. The text analyzer looks for the files under `collectorName/filename`, the other way around as how the run collector was saving the file, which made it tricky to use.

```yaml
kind: Preflight
metadata:
  name: example
spec:
  collectors:
    - run:
        collectorName: "run-ping"
        image: busybox:1
        podName: ping
        namespace: default
        command: ["ping"]
        args: ["-w", "10", "-c", "10", "-i", "0.3", "www.google.com"]
  analyzers:
    - textAnalyze:
        checkName: "run-ping"
        fileName: "ping.log"
        regexGroup: '(?P<Transmitted>\d+) packets? transmitted, (?P<Received>\d+) packets? received, (?P<Loss>\d+)(\.\d+)?% packet loss'
        outcomes:
          - pass:
              when: "Loss < 5"
              message: Solid connection to google.com
          - fail:
              message: High packet loss
```

Fix #189

### DIFF
--- a/pkg/collect/run.go
+++ b/pkg/collect/run.go
@@ -24,8 +24,12 @@ func Run(c *Collector, runCollector *troubleshootv1beta2.Run) (map[string][]byte
 		return nil, errors.Wrap(err, "failed to create client from config")
 	}
 	if runCollector.PodName == "" {
-		imageName := strings.Split(runCollector.Image, ":")
-		runCollector.PodName = imageName[0]
+		if runCollector.Image != "" {
+			imageName := strings.Split(runCollector.Image, ":")
+			runCollector.PodName = imageName[0]
+		} else {
+			return nil, errors.Errorf("failed to run pod: runcollector.Image field is required")
+		}
 	}
 
 	pod, err := runPod(ctx, client, runCollector, c.Namespace)
--- a/pkg/collect/run.go
+++ b/pkg/collect/run.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -21,6 +22,10 @@ func Run(c *Collector, runCollector *troubleshootv1beta2.Run) (map[string][]byte
 	client, err := kubernetes.NewForConfig(c.ClientConfig)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create client from config")
+	}
+	if runCollector.PodName == "" {
+		imageName := strings.Split(runCollector.Image, ":")
+		runCollector.PodName = imageName[0]
 	}
 
 	pod, err := runPod(ctx, client, runCollector, c.Namespace)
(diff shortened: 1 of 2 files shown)


matias-manavella-nimble:
A decision should be made on this change considering backwards compatibility. The user who opened the issue share its concern about this, and I don't know exactly if there is a policy concerning this situations. 
I personally think that the changes should be applied, as they would make the run collector clearer to use, and it would work better whit the text analyzer, let me know what you think

mrcrowl (dismissed):

pkg/collect/run.go
```diff
@@ -103,7 +113,10 @@ func runWithoutTimeout(ctx context.Context, c *Collector, pod *corev1.Pod, runCo
 	limits := troubleshootv1beta2.LogLimits{
 		MaxLines: 10000,
 	}
-	podLogs, err := getPodLogs(ctx, client, *pod, runCollector.Name, "", &limits, true)
+	if runCollector.CollectorName == "" {
+		runCollector.CollectorName = fmt.Sprintf("run-collector-pod-%s", pod.Name)
+	}
+	podLogs, err := getPodLogs(ctx, client, *pod, runCollector.CollectorName, "", &limits, true)
```
divolgin, line 119: @manavellamnimble can we generate file name here and pass it to `getPodLogs`?  and then update other places where `getPodLogs` is called as well.   This will allow you to keep everything backwards compatible.

marccampbell:
@manavellamnimble Will you provide an example of how this would work in a copy collector, when there are multiple replicas of the pod?

Say I have a copy collector that's copying log files from a pod generated by a deployment. And the deployment has 3 replicas. Will the new (or old?) solution work in this case?

manavellamnimble (commented):
@marccampbell this change does not affect the text analyzer, but only the run collector, which creates the path the other way around as the rest of collectors, and is therefore not very intuitive to use. The case you describe would work in both versions like this (note that a wildcard '*' is used to capture all the replicas, regardless the  replicaset and pod id strings), as the:
```YAML
 collectors:
    - copy:
        selector: 
          - app=myapp
        namespace: default
        containerPath: /internal/path/to/logfile/logfile.log
  analyzers:
    - textAnalyze:
        checkName: "run-ping"
        fileName: <namespace>/<deployment-name>-*/internal/path/to/logfile/logfile.log
        regexGroups: regexrule
        outcomes:
          - pass:
          - fail:
```

PR #287, opened 2020-10-19
added flag --since-time for Logs Collectors

### LINKED ISSUE #277: Support command line (and web UI) way of adjusting time period for collecting logs
Some customers produce so many logs that a support bundle may only contain 3 seconds of logs. A vendor can adjust the size of logs to collect, but with that kind of volume, it's not possible for a vendor to provide a single configuration that will work for all customer scenarios. It would be much better for the enterprise user to be able to narrow down the collection themselves, based on their knowledge of when a problem has occurred. 

I'm not positive about best design here, but can imagine a command line option (and in Kots, a web form) to specify a start and end time, or maybe just a start time and we collect as many logs as the vendor spec specifies. Or maybe they specify a time and we collect +/- around that time.

Related, but probably out of scope, might be to let a user specify a request ID that we could filter for. If they really generate a massive amount of logs, this kind of ID might be necessary to truly collect everything a vendor needs for debugging.

Also related, but probably out of scope, if a vendor includes some kind of advanced logging capabilities (e.g. Elasticsearch, Loki), could we tap into that capability to return aggregated and filtered logs?

manavellamnimble:
@markpundsack @divolgin Hi, this is the first approach for a flag to determine a point of time, from which the logs collector should start collecting logs. It overrides all the logs collectors, and keeps the maxLine parameter from the original specs. Let me know what you think of it. Fix #277 An example:
```SHELL
    $ kubectl support-bundle --since-time="2020-10-19T12:36:23Z" file_or_url 
```

### DIFF
--- a/cmd/preflight/cli/root.go
+++ b/cmd/preflight/cli/root.go
@@ -35,6 +35,7 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-image", "", "the full name of the collector image to use")
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
+	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
 
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
--- a/cmd/preflight/cli/root.go
+++ b/cmd/preflight/cli/root.go
@@ -36,7 +36,7 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
 	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
-	cmd.Flags().String("since", "", "forces pod's logs collectors to  return logs newer than a relative duration like 5s, 2m, or 3h.")
+	cmd.Flags().String("since", "", "forces pod's logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
 
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
--- a/cmd/preflight/cli/root.go
+++ b/cmd/preflight/cli/root.go
@@ -35,7 +35,7 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-image", "", "the full name of the collector image to use")
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
-	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
+	cmd.Flags().String("since-time", "", "force pod logs collectors to return logs after a specific date (RFC3339)")
 	cmd.Flags().String("since", "", "force pod logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
 
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
--- a/cmd/preflight/cli/root.go
+++ b/cmd/preflight/cli/root.go
@@ -36,6 +36,7 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
 	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
+	cmd.Flags().String("since", "", "forces pod's logs collectors to  return logs newer than a relative duration like 5s, 2m, or 3h.")
 
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
--- a/cmd/preflight/cli/root.go
+++ b/cmd/preflight/cli/root.go
@@ -36,7 +36,7 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
 	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
-	cmd.Flags().String("since", "", "forces pod's logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
+	cmd.Flags().String("since", "", "force pod logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
 
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
--- a/cmd/preflight/cli/run.go
+++ b/cmd/preflight/cli/run.go
@@ -181,12 +181,12 @@ func parseTimeFlags(v *viper.Viper, progressChan chan interface{}, collectors []
 		}
 		sinceTime, err = time.Parse(time.RFC3339, v.GetString("since-time"))
 		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("unable to parse date and time %s as YYYY-MM-DDTHH:MM:SSZHH:MM, e.g.:\"2006-01-02T15:04:05Z07:00\"", v.GetString("since-time")))
+			return errors.Wrap(err, "unable to parse --since-time flag")
 		}
 	} else {
 		parsedDuration, err := time.ParseDuration(v.GetString("since"))
 		if err != nil {
-			return errors.Wrap(err, fmt.Sprintf("unable to parse time duration %s", v.GetString("since")))
+			return errors.Wrap(err, "unable to parse --since flag")
 		}
 		now := time.Now()
 		sinceTime = now.Add(0 - parsedDuration)
--- a/cmd/preflight/cli/run.go
+++ b/cmd/preflight/cli/run.go
@@ -129,26 +129,10 @@ func runPreflights(v *viper.Viper, arg string) error {
 		KubernetesRestConfig:   restConfig,
 	}
 
-	if v.GetString("since-time") != "" {
-		if v.GetString("since") != "" {
-			progressChan <- errors.Errorf("Only one of since-time / since may be used. The flag since-time will be used.")
-		}
-		for _, collector := range preflightSpec.Spec.Collectors {
-			if collector.Logs != nil {
-				if collector.Logs.Limits == nil {
-					collector.Logs.Limits = new(troubleshootv1beta2.LogLimits)
-				}
-				collector.Logs.Limits.SinceTime = v.GetString("since-time")
-			}
-		}
-	} else if v.GetString("since") != "" {
-		for _, collector := range preflightSpec.Spec.Collectors {
-			if collector.Logs != nil {
-				if collector.Logs.Limits == nil {
-					collector.Logs.Limits = new(troubleshootv1beta2.LogLimits)
-				}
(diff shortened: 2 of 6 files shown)


markpundsack:
Great! I'd like to see it support `since` as well, since relative time is easier to manipulate than UTC time. Does `since-time` support local time?

matias-manavella-nimble:
@markpundsack Would you like to add the --since flag or do you want the --since-time flag to accept hours, `--since-time="720h"` for example. I think if you use the 'Z' at the end of the date-time, like in "2020-10-19T12:36:23Z", the local time is used.

markpundsack:
@manavellamnimble Both have value.

> I think if you use the 'Z' at the end of the date-time, like in "2020-10-19T12:36:23Z", the local time is used.

Hmm... I think 'Z' means "zero offset" which means UTC.

markpundsack:
Btw, we don’t need to block merging this; `since` handling can be a separate PR.

markpundsack:
@jgruica @GraysonNull We should consider adding this to the Kots web UI as well.

matias-manavella-nimble:
@markpundsack Hi! I added the --since flag, so both can be shipped (and documented) together.

markpundsack:
Awesome, thank you!

markpundsack:
@salahalsaleh Can you review this?

cmd/preflight/cli/root.go
```diff
@@ -35,6 +35,8 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-image", "", "the full name of the collector image to use")
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
+	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
+	cmd.Flags().String("since", "", "forces pod's logs collectors to  return logs newer than a relative duration like 5s, 2m, or 3h.")
```
sgalsaleh:
```suggestion
	cmd.Flags().String("since", "", "forces pod's logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
```

cmd/troubleshoot/cli/run.go
```diff
@@ -439,6 +439,23 @@ func runCollectors(v *viper.Viper, collectors []*troubleshootv1beta2.Collect, ad
...
+		}
+		for _, collectors := range cleanedCollectors {
+			if collectors.Collect.Logs != nil {
+				collectors.Collect.Logs.Limits.SinceTime = v.GetString("since-time")
+			}
+		}
+	} else if v.GetString("since") != "" {
+		for _, collectors := range cleanedCollectors {
+			if collectors.Collect.Logs != nil {
+				collectors.Collect.Logs.Limits.SinceTime = v.GetString("since-time")
```
sgalsaleh:
```suggestion
				collectors.Collect.Logs.Limits.Since = v.GetString("since")
```

cmd/preflight/cli/run.go
```diff
@@ -129,6 +129,23 @@ func runPreflights(v *viper.Viper, arg string) error {
...
+		if v.GetString("since") != "" {
+			progressChan <- errors.Errorf("Only one of since-time / since may be used. The flag since-time will be used.")
+		}
+		for _, collectors := range preflightSpec.Spec.Collectors {
+			if collectors.Logs != nil {
+				collectors.Logs.Limits.SinceTime = v.GetString("since-time")
+			}
+		}
+	} else if v.GetString("since") != "" {
+		for _, collectors := range preflightSpec.Spec.Collectors {
```
sgalsaleh: would be clearer to name this `collector` instead of `collectors`
```suggestion
		for _, collector := range preflightSpec.Spec.Collectors {
```

cmd/preflight/cli/run.go
```diff
@@ -129,6 +129,23 @@ func runPreflights(v *viper.Viper, arg string) error {
 		KubernetesRestConfig:   restConfig,
 	}
 
+	if v.GetString("since-time") != "" {
+		if v.GetString("since") != "" {
+			progressChan <- errors.Errorf("Only one of since-time / since may be used. The flag since-time will be used.")
+		}
+		for _, collectors := range preflightSpec.Spec.Collectors {
+			if collectors.Logs != nil {
+				collectors.Logs.Limits.SinceTime = v.GetString("since-time")
```
sgalsaleh: `Limits` can be `nil` here. Might wanna initialize it before assigning the value to `SinceTime`. Same for other occurrences.

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -31,8 +31,10 @@ type Secret struct {
 }
 
 type LogLimits struct {
-	MaxAge   string `json:"maxAge,omitempty" yaml:"maxAge,omitempty"`
-	MaxLines int64  `json:"maxLines,omitempty" yaml:"maxLines,omitempty"`
+	MaxAge    string `json:"maxAge,omitempty" yaml:"maxAge,omitempty"`
+	MaxLines  int64  `json:"maxLines,omitempty" yaml:"maxLines,omitempty"`
+	SinceTime string
```
sgalsaleh: why not add json and yaml tags for these two fields too?

pkg/collect/logs.go
```diff
@@ -117,7 +117,19 @@ func getPodLogs(ctx context.Context, client *kubernetes.Clientset, pod corev1.Po
...
+		if err != nil {
+			logger.Printf("unable to parse --since-time=%s\n", limits.SinceTime)
+		}
+
+		sinceTime := metav1.NewTime(t)
+		podLogOpts.SinceTime = &sinceTime
+
+	} else if limits != nil && (limits.MaxAge != "" || limits.Since != "") {
+		if limits.Since != "" {
+			limits.MaxAge = limits.Since
```
salahalsaleh, line 131: why do we need `Since` if it is the same functionality as `MaxAge`? that feels redundant.

pkg/collect/logs.go
```diff
@@ -117,7 +117,19 @@ func getPodLogs(ctx context.Context, client *kubernetes.Clientset, pod corev1.Po
...
 	}
 
-	if limits != nil && limits.MaxAge != "" {
+	if limits != nil && limits.SinceTime != "" {
+		t, err := time.Parse(time.RFC3339, limits.SinceTime)
+		if err != nil {
+			logger.Printf("unable to parse --since-time=%s\n", limits.SinceTime)
+		}
+
+		sinceTime := metav1.NewTime(t)
```
sgalsaleh: this should be in an `else` statement to handle the case when parsing `SinceTime` fails.

pkg/apis/troubleshoot/v1beta2/collector_shared.go
matias-manavella-nimble: I think it would be pointless to make since-time part of the specs, because I can't imagine a vendor fixing a date and time for the logs, and the user shouldn't modify the specs either. And in the case of since, because it overrides maxAge, so it would be redundant to add it to the specs. I could add them though.

cmd/preflight/cli/run.go
matias-manavella-nimble: I removed all the 's'

pkg/collect/logs.go
matias-manavella-nimble: Done!

cmd/preflight/cli/run.go
matias-manavella-nimble: Done!

pkg/apis/troubleshoot/v1beta2/collector_shared.go
sgalsaleh: Makes sense, no need to add them, just wanted to make sure we're excluding them intentionally.

sgalsaleh (approved):

cmd/preflight/cli/root.go
```diff
@@ -35,6 +35,8 @@ that a cluster meets the requirements to run an application.`,
 	cmd.Flags().String("collector-image", "", "the full name of the collector image to use")
 	cmd.Flags().String("collector-pullpolicy", "", "the pull policy of the collector image")
 	cmd.Flags().Bool("collect-without-permissions", false, "always run preflight checks even if some require permissions that preflight does not have")
+	cmd.Flags().String("since-time", "", "forces pod's logs collectors to return logs after a specific date (RFC3339)")
```
markpundsack:
```suggestion
	cmd.Flags().String("since-time", "", "force pod logs collectors to return logs after a specific date (RFC3339)")
```

cmd/preflight/cli/root.go
```diff
@@ -35,6 +35,8 @@ that a cluster meets the requirements to run an application.`,
...
+	cmd.Flags().String("since", "", "forces pod's logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
```
markpundsack:
```suggestion
	cmd.Flags().String("since", "", "force pod logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
```

cmd/preflight/cli/run.go
```diff
@@ -129,6 +129,29 @@ func runPreflights(v *viper.Viper, arg string) error {
 		KubernetesRestConfig:   restConfig,
 	}
 
+	if v.GetString("since-time") != "" {
+		if v.GetString("since") != "" {
+			progressChan <- errors.Errorf("Only one of since-time / since may be used. The flag since-time will be used.")
```
markpundsack: Is that how Kubectl handles it? Or does it abort because of invalid args?

cmd/troubleshoot/cli/root.go
```diff
@@ -39,6 +39,8 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 	cmd.Flags().StringSlice("redactors", []string{}, "names of the additional redactors to use")
 	cmd.Flags().Bool("redact", true, "enable/disable default redactions")
 	cmd.Flags().Bool("collect-without-permissions", false, "always generate a support bundle, even if it some require additional permissions")
+	cmd.Flags().String("since-time", "", "forces pods logs collectors to return logs after a specific date (RFC3339)")
```
markpundsack:
```suggestion
	cmd.Flags().String("since-time", "", "force pod logs collectors to return logs after a specific date (RFC3339)")
```

cmd/troubleshoot/cli/root.go
```diff
@@ -39,6 +39,8 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
...
+	cmd.Flags().String("since", "", "forces pod's logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
```
markpundsack:
```suggestion
	cmd.Flags().String("since", "", "force pod logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
```

cmd/preflight/cli/run.go
matias-manavella-nimble: Yes I had my doubts about this too. K8s aborts with the message `error: parsing time`. I followed the error handling for the limits in the logs collector (communicating it through the progress channel), but it would make sense to abort because if someone used the flag and it can't be parsed maybe there would be no point in running through all the collectors and analyzers

matias-manavella-nimble:
@markpundsack based on what we discussed, I've redesign the solution. Before running the collectors, the flags are parsed to time variable to check if it is all right, and throws an error if there is something wrong. All of this happens in a separate function  to make it more organized. This modification also requires just one extra field in log.limits (not two) and requires less work to implement in the log collector. @salahalsaleh Can you please check it?

cmd/preflight/cli/run.go
```diff
@@ -161,3 +169,35 @@ func runPreflights(v *viper.Viper, arg string) error {
...
 }
+
+func parseTimeFlags(v *viper.Viper, progressChan chan interface{}, collectors []*troubleshootv1beta2.Collect) error {
+	var (
+		sinceTime time.Time
+		err       error
+	)
+	if v.GetString("since-time") != "" {
+		if v.GetString("since") != "" {
+			progressChan <- errors.Errorf("Only one of since-time / since may be used. The flag since-time will be used.")
```
markpundsack: If `kubectl` errors on inclusion of both, I think we should do the same. If someone specifies both, it's a configuration error that they probably didn't mean to do. Is ignoring one of the arguments really the right thing?

matias-manavella-nimble: Makes sense, I changed it to return an error.

markpundsack:
Thanks for the changes.

BTW, do we not have tests on this project?

PR #288, opened 2020-10-28
Adding kubelet metrics collector

marccampbell:


xavpaice:
Closing: the kubelet metrics _should_ be heading into Prometheus, and have a collector for the Prometheus metrics added.

PR #290, opened 2020-10-31
Collector for RRD data

divolgin:
Example collector and analyzer

```
kind: Analyzer
metadata:
  name: collector-sample
spec:
  collectors:
    - collectd:
        collectorName: "my-rrd"
        namespace: 'default'
        image: 'alpine:3'
        name: my-rrd-name
        hostPath: "/var/lib/collectd/rrd"
        imagePullPolicy: IfNotPresent

  analyzers:
    - collectd:
        checkName: "load and stuff"
        outcomes:
          - fail:
              when: "load > 1"
              message: High load detected on cluster nodes.
          - pass:
              message: Load on cluster nodes does not exceed expected values
```

Load value is normalized using the number of CPUs found in the RRD data.


pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -123,6 +123,16 @@ type Database struct {
 	URI           string `json:"uri" yaml:"uri"`
 }
 
+type RRD struct {
```
divolgin: How about calling this collector something more generic.  It doesn't really care about the format of the files collected.

marccampbell: I think there's discoverability in "collectd" vs `rrd`.  Unless there's another common tool that overlaps with troubleshoot using rrd?

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -123,6 +123,16 @@ type Database struct {
...
+	CollectorMeta   `json:",inline" yaml:",inline"`
+	Namespace       string            `json:"namespace" yaml:"namespace"`
+	Image           string            `json:"image" yaml:"image"`
```
marccampbell, line 129: is this needed  or why would someone specify this ?

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -123,6 +123,16 @@ type Database struct {
...
+	ImagePullPolicy string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
+	ImagePullSecret *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
+	Timeout         string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	HostPath        string            `json:"hostPath" yaml:"hostPath"`
```
marccampbell, line 133: We can default to the kURL well-known path if this is empty, right?

divolgin: Yes, this will be injected by kots and added to our default spec.

pkg/apis/troubleshoot/v1beta2/collector_shared.go
divolgin: I don't want to hardcode any images in troubleshoot.  But it _will_ be optional. We'll inject either a kotsadm image here or use some public image in default static collector.  But in case someone wants to use this from CLI with their own spec, they could specify an image here.

pkg/apis/troubleshoot/v1beta2/collector_shared.go
divolgin: Makes sense.

sgalsaleh (approved):

PR #304, opened 2020-11-25
Prevent failure

manavellamnimble:
@marccampbell  @divolgin I would like your opinion on this. I kept working on preventing failure on unnecessary default collectors, either RBAC errors or errors of other kind. 

 - in case of RBAC errors, the not-allowed resource is checked against the analyzers in the spec. If no analyzer will use that resource, then it is not considered an error.
 - in case of failure of any of the collectors in the cluster resources, the collector (namespace, deployments, etc...) is checked against the analyzers. The error is only informed if  there is no analyzer using that collector, or an error is sent if the collector is required. 

Issues #239 #227

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -147,7 +147,7 @@ func runTroubleshoot(v *viper.Viper, arg string) error {
 		}
 	}()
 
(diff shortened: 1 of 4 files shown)


matias-manavella-nimble:
PD: collector_test would need some modifications to work, but I wanted your opinion bf continuing working on this.

PR #305, opened 2020-11-27
read uri from secret

### LINKED ISSUE #300: Postgres/Mysql/Redis analyzers should support a valueFrom for URI
Currently, the only way to pass the URI to one of the database analyzers is to hard code it and write it into a spec. Often this is handled via templating to make it portable.

The uri field should support referencing an existing secret also, much like env vars do. Ideally, we can just reuse existing K8s objects and make it a struct, like:

```
uri:
  value: <>
  valueFrom:
    secretKeyRef:
     ...
```

manavellamnimble:
@marccampbell I think this is the most effective way to read the uri from the secret. I tried using the functions from the secret collector but was a bit more messy. Fix #300


xavpaice:
the amount of rework required to merge in this change justifies revisiting the solution entirely and deciding if we want to take a different approach.  Closing for now.

PR #311, opened 2021-02-08
Host preflight

areed:


examples/preflight/host-disk-usage.yaml
```diff
@@ -0,0 +1,33 @@
...
+              message: /var/lib/kubelet has less than 10Gi of disk space available
+          - fail:
+              when: "used/total > 70%"
+              message: /var/lib/kubelet is more than 70% full
+          - warn:
+              when: "total < 40Gi"
+              message: /var/lib/kubelet has less than 40Gi of total space
+          - warn:
+              when: "used/total > 60%"
+              message: /var/lib/kubelet is more than 50% full
```
laverya: 60 and 50, pick one

examples/preflight/sample-host-preflight.yaml
```diff
@@ -0,0 +1,51 @@
...
+        timeout: 5000ms
+    - diskUsage:
+        collectorName: ephemeral
+        path: /var/lib/kubelet
+  analyzers:
+    - cpu:
+        outcomes:
+          - fail:
+              when: "< 4"
+              message: This server has less than 4 CPU cores, and we require 8, but recommend 16
```
laverya, line 20: I feel like this example used to have another option

pkg/analyze/host_cpu.go
```diff
@@ -0,0 +1,165 @@
...
+
+				return &result, nil
+			}
+		}
+	}
+
+	return &result, nil
+}
+
+func compareHostCPUConditionalToActual(conditional string, logicalCount int, physicalCount int) (res bool, err error) {
```
laverya, line 102: Could we get a unit test for this? There's one for doCompareHostCPU, which is a lot simpler

laverya (approved):

examples/preflight/sample-host-preflight.yaml
areed: I added examples/preflight/host-cpu.yaml with more examples

pkg/analyze/host_cpu.go
areed: Added

PR #314, opened 2021-02-09
Host HTTP request analyzer

areed:


examples/preflight/host-http.yaml
```diff
@@ -0,0 +1,22 @@
...
+          url: https://registry.replicated.com
+  analyzers:
+    - http:
+        collectorName: registry
+        outcomes:
+          - fail:
+              when: "error"
+              message: Error connecting to registry
+          - pass:
+              when: "statusCode == 404"
```
laverya, line 19: do we not have an endpoint that returns something besides 404?

laverya (approved):

examples/preflight/host-http.yaml
areed: There's https://registry.replicated.com/healthz but that gives 404 also

PR #317, opened 2021-02-11
Analyze TCP connection

areed:


pkg/collect/host_tcp_connect.go
```diff
@@ -0,0 +1,66 @@
...
+
+func attemptConnect(address string, timeout time.Duration) NetworkStatus {
+	stopAfter := time.Now().Add(timeout)
+
+	for {
+		if time.Now().After(stopAfter) {
+			return NetworkStatusConnectionTimeout
+		}
+
+		conn, err := net.DialTimeout("tcp", address, 50*time.Millisecond)
```
divolgin: Is there a drawback to using the timeout provided here?  What if network is slow and it always takes more than 50ms to connect?

pkg/collect/host_tcp_connect.go
```diff
@@ -0,0 +1,66 @@
...
+		if err != nil {
+			if strings.Contains(err.Error(), "i/o timeout") {
+				time.Sleep(time.Millisecond * 50)
+				continue
+			}
+			if strings.Contains(err.Error(), "connection refused") {
+				return NetworkStatusConnectionRefused
+			}
+			fmt.Println(err.Error())
```
divolgin: this should probably not print anything

pkg/collect/host_tcp_connect.go
divolgin: you could wrap this into a func and call `defer conn.Close()` on every iteration.  right now this can leak sockets

divolgin (approved):

PR #320, opened 2021-02-13
Analyze HTTP load balancer

areed:


pkg/collect/host_httploadbalancer.go
```diff
@@ -0,0 +1,150 @@
...
+		ForceAttemptHTTP2: true,
+	}
+	client := http.Client{
+		Transport: transport,
+	}
+
+	buf := bytes.NewBuffer(request)
+	req, err := http.NewRequestWithContext(ctx, "POST", address, buf)
+	if err != nil {
+		fmt.Println(err.Error())
```
areed, line 140: This would be a bug such as context is nil, which is why this one is logged but no other errors are.

pkg/collect/host_tcploadbalancer.go
```diff
@@ -9,16 +9,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-type ConnectionResult int
```
areed, line 12: unused code

laverya (approved):

PR #323, opened 2021-02-17
Analyze fs sequential write performance

areed:
The included example found P99 latency of 2.6ms, compared to 2.5ms reported by `fio --rw=write --ioengine=sync --fdatasync=1 --directory=/var/lib/etcd --size=220m --bs=2300`.


laverya (approved):

dexhorthy:
🤩

PR #325, opened 2021-02-19
feat(analyzer): rke2 and k3s distro support

DanStough:
Please confirm removing `# Will be supported in the future` is accurate 😬

### DIFF
--- a/examples/preflight/sample-preflight.yaml
+++ b/examples/preflight/sample-preflight.yaml
@@ -60,13 +60,18 @@ spec:
           - pass:
               when: "== aks"
               message: AKS is a supported distribution
-          # Will be supported in the future
           - pass:
               when: "== kurl"
               message: KURL is a supported distribution
           - pass:
               when: "== digitalocean"
               message: DigitalOcean is a supported distribution
+          - pass:
+              when: "== rke2"
+              message: RKE2 is a supported distribution
+          - pass:
+              when: "== k3s"
+              message: K3S is a supported distribution
           - warn:
               message: Unable to determine the distribution of Kubernetes
     - nodeResources:
--- a/pkg/analyze/distribution.go
+++ b/pkg/analyze/distribution.go
@@ -22,6 +22,8 @@ type providers struct {
 	aks           bool
 	ibm           bool
 	minikube      bool
+	rke2          bool
+	k3s           bool
 }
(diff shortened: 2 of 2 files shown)


pkg/analyze/distribution.go
```diff
@@ -77,6 +81,21 @@ func ParseNodesForProviders(nodes []corev1.Node) (providers, string) {
 				foundProviders.minikube = true
 				stringProvider = "minikube"
 			}
+			if k == "node.kubernetes.io/instance-type" && v == "k3s" {
+				foundProviders.k3s = true
+				stringProvider = "k3s"
+			}
+			if k == "beta.kubernetes.io/instance-type" && v == "k3s" {
+				foundProviders.k3s = true
+				stringProvider = "k3s"
```
laverya, line 90: this is a duplicate of the above

DanStough: To clarify, you're saying you just want to see it consolidated on the same line? The key is different.

laverya: 🤦

laverya (approved):

examples/preflight/sample-preflight.yaml
```diff
@@ -60,13 +60,18 @@ spec:
           - pass:
               when: "== aks"
               message: AKS is a supported distribution
-          # Will be supported in the future
           - pass:
               when: "== kurl"
```
markpundsack, line 64: @DanStough My assumption is that the `pass` logic stops at the first one that passes, in which case, if rke2 is installed by kURL, this preflight will always return kURL, even if it is also rke2. Is that your intended behavior? If not, would it make sense to put the more-specific distros above the kurl check?

DanStough: That is how I understand it works. My thought was somehow that kurl is a more specific/interesting case than rke or k3s alone. The vendor would define the kurl spec, so this would be known and higher priority than roll your own k3s or rke2. I see it works the other way though. I'm happy to change it if you think it's the other way around.

markpundsack: That's fine. I wanted to make sure you thought about it and it was intentional. It may cause confusion for someone, but the vendor can certainly change the order if they care more about that part.

PR #326, opened 2021-02-19
Analyze TLS certificate

areed:


pkg/collect/host_certificate.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		if strings.Contains(err.Error(), "no such file") {
+			result = KeyPairMissing
+		} else if strings.Contains(err.Error(), "PEM inputs may have been switched") {
+			result = KeyPairSwitched
+		} else if strings.Contains(err.Error(), "found a certificate rather than a key") {
+			result = KeyPairSwitched
+		} else if strings.Contains(err.Error(), "private key does not match public key") {
+			result = KeyPairMismatch
+		} else if strings.Contains(err.Error(), "failed to parse private key") {
```
laverya, line 31: I would greatly prefer to use `errors.Is` here for each of these `strings.Contains`

laverya (approved):
Approved, but again I would prefer to do type matching with `errors.Is` rather than string comparison

pkg/collect/host_certificate.go
areed: I agree but I don't think it's possible since the standard library is using errors.New(string) https://golang.org/src/crypto/tls/tls.go?s=8380:8451#L299

PR #330, opened 2021-02-26
add pv and pvcs to support bundle

jala-dx:


sgalsaleh (approved):

pkg/collect/cluster_resources.go
```diff
@@ -532,3 +550,39 @@ func convertToPolicyRule(status authorizationv1.SubjectRulesReviewStatus) []rbac
...
+		return nil, []string{err.Error()}
+	}
+
+	b, err := json.MarshalIndent(pv.Items, "", "  ")
+	if err != nil {
+		return nil, []string{err.Error()}
+	}
+	return b, nil
+}
+func pvcs(ctx context.Context, client *kubernetes.Clientset, namespaces []string) (map[string][]byte, map[string]string) {
```
divolgin, line 567: add line break between functions

pkg/collect/cluster_resources.go
```diff
@@ -532,3 +550,39 @@ func convertToPolicyRule(status authorizationv1.SubjectRulesReviewStatus) []rbac
...
+		if err != nil {
+			errorsByNamespace[namespace] = err.Error()
+			continue
+		}
+
+		pvcsByNamespace[namespace+".json"] = b
+	}
+
+	return pvcsByNamespace, errorsByNamespace
+
```
divolgin: remove empty line

divolgin (approved):

divolgin (approved):

markpundsack:
@jala-dx Cool! Does this behave gracefully with minimalRBAC? Like, will it return PVCs in namespace KOTS has access to, and gracefully skip other namespaces?

jala-dx:
> @jala-dx Cool! Does this behave gracefully with minimalRBAC? Like, will it return PVCs in namespace KOTS has access to, and gracefully skip other namespaces?

@markpundsack thanks:) ! yes it is a best effort. it walks through all the namespaces in kots and gets the PVCs. For some reason say RBAC, it cannot get, the code skips and continues and gets the next one.

markpundsack:
Thanks. Have you confirmed that with minimalRBAC, it does as expected, where it includes things in-namespace? I'm just slightly worried that it won't include any PVCs, if, for example, the account doesn't have permission to even list namespaces, but *does* in fact have access to at least one namespace with a PVC.

PR #343, opened 2021-04-07
Docker registry collector/analyzer

divolgin:
Adds a collector that, given a list of images and optional registry credentials, will produce a file that lists whether each image can be successfully pulled.

Example specs:

```
kind: Preflight
metadata:
spec:
  collectors:
    - registry:
        collectorName: "registry"
        namespace: test-ns
        imagePullSecret:
          type: kubernetes.io/dockerconfigjson
          name: troubleshoot-test
          data:
            .dockerconfigjson: ewogICJhdXRocyI6IHsK...g==
        images:
          - "alpine:3.9"
          - "private.registry.com/someorg/image:v1.0.0"

  analyzers:
    - registry:
        outcomes:
          - fail:
              when: "missing > 0"
              message: Images are missing from registry
          - warn:
              when: "errors > 0"
              message: Failed to check if images are present in registry
          - pass:
              message: All images are present in registry
```


marccampbell:
I'm not sure "registry" is the right name here. Unless there are more plans to expand it. Right now, it only checks that an image is present in a registry. If that's the extent, maybe we need a more descriptive name?

pkg/analyze/registry.go
```diff
@@ -0,0 +1,183 @@
...
+
+	title := analyzer.CheckName
+	if title == "" {
+		title = collectorName
+	}
+
+	result := &AnalyzeResult{
+		Title:   title,
+		IconKey: "kubernetes_registry_analyze",
+		IconURI: "https://troubleshoot.sh/images/analyzer-icons/registry-analyze.svg",
```
DanStough, line 54: PR TBD?

pkg/analyze/registry.go
```diff
@@ -0,0 +1,183 @@
...
+				result.IsFail = true
+				result.Message = outcome.Fail.Message
+				result.URI = outcome.Fail.URI
+
+				return result, nil
+			}
+
+			isMatch, err := compareRegistryConditionalToActual(outcome.Fail.When, numVerifiedImages, numMissingImages, numErrors)
+			if err != nil {
+				return result, errors.Wrap(err, "failed to compare redis database conditional")
```
DanStough:
```suggestion
				return result, errors.Wrap(err, "failed to compare registry conditional")
```

pkg/analyze/registry.go
```diff
@@ -0,0 +1,183 @@
...
+				result.IsWarn = true
+				result.Message = outcome.Warn.Message
+				result.URI = outcome.Warn.URI
+
+				return result, nil
+			}
+
+			isMatch, err := compareRegistryConditionalToActual(outcome.Warn.When, numVerifiedImages, numMissingImages, numErrors)
+			if err != nil {
+				return result, errors.Wrap(err, "failed to compare redis database conditional")
```
DanStough:
```suggestion
				return result, errors.Wrap(err, "failed to compare registry conditional")
```

pkg/analyze/registry.go
```diff
@@ -0,0 +1,183 @@
...
+				result.IsPass = true
+				result.Message = outcome.Pass.Message
+				result.URI = outcome.Pass.URI
+
+				return result, nil
+			}
+
+			isMatch, err := compareRegistryConditionalToActual(outcome.Pass.When, numVerifiedImages, numMissingImages, numErrors)
+			if err != nil {
+				return result, errors.Wrap(err, "failed to compare redis database conditional")
```
DanStough:
```suggestion
				return result, errors.Wrap(err, "failed to compare registry conditional")
```

DanStough (approved):
Some minor stuff, otherwise looks good. 

Is the output of which images are missing/couldn't be pulled exposed anywhere, or is it just the numbers? Would be useful for debugging to know which of the images have problems.

pkg/analyze/registry.go
divolgin: yeah

divolgin:
@DanStough yeah, this is a missing feature in preflights.   You can get the output in support bundle though as a workaround.

PR #351, opened 2021-04-12
Benchmark write latency with background IOPS

areed:
Add a background IOPS feature to the filesystem performance collector
that specifies separate read and write background IOPS to perform while
measuring latency. This allows for better assessment of whether etcd
will be stable when running alongside other workloads on the same
cluster.

Also add templating to the outcome message of the filesystem performance
analyzers to allow printing individual latency percentiles or the entire
table.

Remove the random IOPS benchmark since it was attempting to perform
unaligned direct I/O.

### DIFF
--- a/pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
+++ b/pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
@@ -73,13 +73,43 @@ type TCPConnect struct {
 	Timeout           string `json:"timeout,omitempty"`
 }
 
+// FilesystemPerformance benchmarks sequential write latency on a single file.
+// The optional background IOPS feature attempts to mimic real-world conditions by running read and
+// write workloads prior to and during benchmark execution.
 type FilesystemPerformance struct {
-	HostCollectorMeta  `json:",inline" yaml:",inline"`
+	HostCollectorMeta `json:",inline" yaml:",inline"`
+	// The size of each write operation performed while benchmarking. This does not apply to the
+	// background IOPS feature if enabled, since those must be fixed at 4096.
 	OperationSizeBytes uint64 `json:"operationSize,omitempty"`
-	Directory          string `json:"directory,omitempty"`
-	FileSize           string `json:"fileSize,omitempty"`
-	Sync               bool   `json:"sync,omitempty"`
-	Datasync           bool   `json:"datasync,omitempty"`
+	// The directory where the benchmark will create files.
+	Directory string `json:"directory,omitempty"`
+	// The size of the file used in the benchmark. The number of IO operations for the benchmark
+	// will be FileSize / OperationSizeBytes. Accepts valid Kubernetes resource units such as Mi.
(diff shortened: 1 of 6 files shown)


pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
```diff
@@ -73,13 +73,38 @@ type TCPConnect struct {
...
+	// The target write IOPS to run while benchmarking. This is a limit and there is no guarantee
+	// it will be reached.
+	BackgroundWriteIOPS int `json:"backgroundWriteIOPS"`
+	// The target read IOPS to run while benchmarking. This is a limit and there is no guarantee
+	// it will be reached.
+	BackgroundReadIOPS int `json:"backgroundReadIOPS"`
+	// Number of threads to use for background write IOPS.
+	BackgroundWriteIOPSJobs int `json:"backgroundWriteIOPSJobs"`
+	// Number of threads to use for background read IOPS.
+	BackgroundReadIOPSJobs int `json:"backgroundReadIOPSJobs"`
```
laverya, line 112: are iops split over the threads, or per-thread?

pkg/collect/host_filesystem_performance_linux.go
```diff
@@ -143,83 +181,116 @@ func collectHostFilesystemPerformance(hostCollector *troubleshootv1beta2.Filesys
...
+				blockOffset := rand.Intn(blocks)
+				offset := int64(blockOffset) * int64(opSize)
+				data := make([]byte, opSize)
+
+				if opts.read {
+					syscall.Pread(int(f.Fd()), data, offset)
+				} else {
+					if _, err := rand.Read(data); err != nil {
+						log.Printf("Failed to generate data for background IOPS write job: %v", err)
+						return
```
laverya, line 290: this breaks the 'done' chan count as far as I can tell

laverya: Wouldn't that make it so that the defer func can never complete if there's an error in this collector somewhere?

pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
areed: Total for all threads. I'll update the comment.

pkg/collect/host_filesystem_performance_linux.go
areed: Yes

pkg/collect/host_filesystem_performance_linux.go
```diff
@@ -230,6 +230,12 @@ func backgroundIOPS(ctx context.Context, opts backgroundIOPSOpts, done chan bool
 				log.Printf("Failed to create temp file for background IOPS job: %v", err)
 				return
```
laverya, line 232: IMO this needs a `done <- true` or `done <- false` too

laverya (approved):
instead of having a dedicated 'check iops' test, instead test latency with a given number of background iops going on.

PR #360, opened 2021-04-20
chore(support-bundle): refactor out support bundle pkg

DanStough:
Refactoring to support using `support-bundle` as a package within kotsadm. Still work to do on consolidating things with what's in preflights but this is a first step.

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -1,20 +1,14 @@
 package cli
 
 import (
-	"archive/tar"
-	"bytes"
-	"compress/gzip"
-	"context"
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
-	"io"
-	"io/ioutil"
 	"net/http"
+
 	"net/url"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


cmd/troubleshoot/cli/run.go
```diff
@@ -678,134 +372,50 @@ func getExpectedContentType(uploadURL string) string {
...
-	}
-
-	return nil
-}
-
-type CollectorFailure struct {
-	Collector *troubleshootv1beta2.Collect
-	Failure   string
-}
-
```
DanStough, line 780: As far as I can tell this isn't used in troubleshoot or kots

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,203 @@
...
+
+	var analyzeResults []*analyzer.AnalyzeResult
+
+	if len(spec.Analyzers) > 0 {
+
+		tmpDir, err := ioutil.TempDir("", "troubleshoot")
+		if err != nil {
+			return analyzeResults, errors.Wrap(err, "failed to make directory for analysis")
+		}
+		defer os.RemoveAll(tmpDir)
```
DanStough, line 197: new

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,203 @@
...
+
+		f, err := os.Open(archivePath)
+		if err != nil {
+			return analyzeResults, errors.Wrap(err, "failed to open support bundle for analysis")
+		}
+		defer f.Close()
```
DanStough, line 203: new

cmd/troubleshoot/cli/run.go
divolgin: indeed

cmd/troubleshoot/cli/run.go
```diff
@@ -147,71 +137,50 @@ func runTroubleshoot(v *viper.Viper, arg string) error {
...
+		ProgressChan:              progressChan,
+		SinceTime:                 sinceTime,
+	}
+
+	archivePath, err := supportbundle.CollectSupportBundleFromSpec(&supportBundle.Spec, additionalRedactors, createOpts)
 	if err != nil {
 		return errors.Wrap(err, "run collectors")
 	}
 
 	fmt.Printf("\r%s\r", cursor.ClearEntireLine())
```
divolgin: should this print to `c` instead of using the fmt package directly?

divolgin (approved):

cmd/troubleshoot/cli/run.go
```diff
@@ -55,15 +56,15 @@ func runTroubleshoot(v *viper.Viper, arg string) error {
 		httpClient = http.DefaultClient
 	}
 
-	collectorContent, err := loadSupportBundleSpec(v, arg)
+	collectorContent, err := supportbundle.LoadSupportBundleSpec(arg, httpClient)
```
divolgin: I find it odd that httpClient is a global variable, but is passed around as a one off argument for some functions.  How about making something similar to `k8sutil.GetRESTConfig()` but for http client?

DanStough: Yes, that's a great point.

cmd/troubleshoot/cli/run.go
DanStough: Not sure - this was existing. Proposing this instead?
```golang
c := color.New()
c.Println("\r%s\r", cursor.ClearEntireLine())
```

divolgin: yeah, seems like an oversight from before.  not related to your changes.

pkg/httputil/config.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func AddTransport(transport *http.Transport) {
+	httpTransport = transport
+}
+
+func GetHttpClient() *http.Client {
+
+	if httpTransport != nil {
+		client := &http.Client{Transport: httpTransport}
```
divolgin: This will leak TCP sockets.  Should reuse the same client, similar to `http.DefaultClient`

divolgin (approved):

PR #366, opened 2021-05-26
Longhorn collector for all CRDs

areed:
Also implement a single analyzer as a proof of concept. More analyzers
can be added using the collected CRDs.


go.mod
```diff
@@ -52,3 +51,28 @@ require (
 	k8s.io/client-go v0.20.2
 	sigs.k8s.io/controller-runtime v0.8.3
 )
+
+replace (
+	github.com/longhorn/longhorn-manager => github.com/replicatedhq/longhorn-manager v1.1.2-0.20210526001437-4003f67ebd06
```
areed, line 56: Until https://github.com/longhorn/longhorn/issues/2626 is done.

pkg/collect/longhorn.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, instanceManager := range instanceManagers.Items {
+		b, err := yaml.Marshal(instanceManager)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to marshal instancemanager %s", instanceManager.Name)
+		}
+		key := filepath.Join(dir, instanceManager.Name+".yaml")
+		final[key] = b
+	}
+
+	// collect backingimagemanagers.longhorn.io
```
areed, line 124: Default kurl installs with longhorn don't have any backingimagemanagers, backingimages, or sharemanagers, but I included them for completeness.

pkg/collect/longhorn.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, shareManager := range shareManagers.Items {
+		b, err := yaml.Marshal(shareManager)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to marshal sharemanager %s", shareManager.Name)
+		}
+		key := filepath.Join(dir, shareManager.Name+".yaml")
+		final[key] = b
+	}
+
+	// collect settings
```
areed, line 169: This is a bit different than the others. All settings are in settings.yaml to avoid looking through 50 different files to find a setting.

pkg/collect/longhorn.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "list sharemagemanagers.longhorn.io")
+	}
+	dir = GetLonghornShareManagersDirectory(ns)
+	for _, shareManager := range shareManagers.Items {
+		b, err := yaml.Marshal(shareManager)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to marshal sharemanager %s", shareManager.Name)
+		}
+		key := filepath.Join(dir, shareManager.Name+".yaml")
```
laverya, line 165: I can't help but think there's a chance that this naming strategy will lead to collisions 

What happens if there's a manager and a share manager with the same name?

Should we make every type of file go in a different directory?

areed: They actually do go in a different directory, but it's kind of hard to tell because I reused the same `dir` variable everywhere. There's a separate `GetLonghornXDirectory` function for every resource.

laverya (approved):

PR #367, opened 2021-05-27
Analyze longhorn replica

areed:


pkg/analyze/longhorn.go
```diff
@@ -67,3 +96,46 @@ func analyzeLonghornNodeSchedulable(node *longhornv1beta1.Node) *AnalyzeResult {
...
+		return result
+	}
+
+	result.IsPass = true
+	result.Message = fmt.Sprintf("Replica is %s", actual)
+
+	return result
+}
+
+func stripRedactedLines(yaml []byte) []byte {
```
areed, line 126: yaml Unmarshal fails on lines like `ip: ***HIDDEN***`

laverya: seems like this function will be a bit over aggressive, but it'll still work for what we need

laverya (approved):

PR #377, opened 2021-06-30
Modify Host Preflights To Return Multiple Analyze Results

murphybytes:
Reference issue [31673 Add Preflight Check for NTP](https://app.clubhouse.io/replicated/story/34340/clock-skew-or-ntp-analyzer-or-preflight) this change addresses a problem that can result in host preflights where, in an analyzer with multiple outcomes, a single outcome, the first one where the predicate evaluates to true is the only one that is returned. This is a problem because if a pass predicate is true, it is returned to the user, and any subsequent warnings or fails are missed.  Note that cluster preflights already return multiple outcomes so this change makes host preflights consistent.

### DIFF
--- a/pkg/analyze/host_time_test.go
+++ b/pkg/analyze/host_time_test.go
@@ -206,6 +206,12 @@ func TestAnalyzeHostTime(t *testing.T) {
 							Message: "Timezone is not set to UTC",
 						},
 					},
+					{
+						Pass: &troubleshootv1beta2.SingleOutcome{
+							When:    "timezone == UTC",
+							Message: "Timezone is set to UTC",
+						},
+					},
 				},
 			},
 			result: []*AnalyzeResult{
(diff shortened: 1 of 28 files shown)


pkg/analyze/host_time_test.go
```diff
@@ -200,10 +211,56 @@ func TestAnalyzeHostTime(t *testing.T) {
...
+			},
+			hostAnalyzer: &troubleshootv1beta2.TimeAnalyze{
+				Outcomes: []*troubleshootv1beta2.Outcome{
+					{
+						Fail: &troubleshootv1beta2.SingleOutcome{
+							When:    "timezone != UTC",
+							Message: "timezone is not set to UTC",
+						},
+						Pass: &troubleshootv1beta2.SingleOutcome{
+							When:    "timezone == UTC",
```
laverya: each outcome should only have one of pass/fail/warn, not multiple

I think that's in another test in this file, too

laverya (changes requested):
Mostly lgtm, but the tests should use different outcomes for each entry and not combine them

laverya (approved):

PR #378, opened 2021-07-01
Always use longhorn namespace

areed:
Also speed up replica checksum collector by using md5sum and parallelizing.

Also skip checksum if data directory does not yet exist or replica not yet assigned to node.

### DIFF
--- a/pkg/collect/longhorn.go
+++ b/pkg/collect/longhorn.go
@@ -265,80 +265,47 @@ func Longhorn(c *Collector, longhornCollector *troubleshootv1beta2.Longhorn) (ma
 }
 
 func GetLonghornNodesDirectory(namespace string) string {
-	if namespace != DefaultLonghornNamespace {
-		return fmt.Sprintf("longhorn/%s/nodes", namespace)
-	}
-	return "longhorn/nodes"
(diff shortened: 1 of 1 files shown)


pkg/collect/longhorn.go
```diff
@@ -360,7 +345,7 @@ func GetLonghornReplicaChecksum(clientConfig *rest.Config, replica longhornv1bet
 		Param("stdin", "true").
 		Param("command", "/bin/bash").
 		Param("command", "-c").
-		Param("command", fmt.Sprintf("sha256sum /host/var/lib/longhorn/replicas/%s/*", replica.Spec.DataDirectoryName))
+		Param("command", fmt.Sprintf("if [ -d %s ]; then md5sum %s/*; fi", dir, dir))
```
laverya, line 348: is md5sum actually meaningfully faster than sha256sum?

laverya: I'd expect to be more limited by disk than CPU

laverya (approved):

pkg/collect/longhorn.go
areed: I got 1m 21s for shas56sum and 25s for md5sum in my test.

PR #381, opened 2021-07-06
Add preflight host collector and analyzer for kernel modules

croomes:
We'd like the ability to check for required kernel modules before performing an installation.  In most cases modules are available but are not yet loaded.  We get the list of loadable modules by recursively listing `*.ko*` files under `/lib/modules$(uname -r)`, and then check `lsmod` to update the status if they're already loaded.

Outcome conditions succeed if at least one status condition matches, but if multiple modules are specified they must all match.  This allows:

```yaml
kind: HostPreflight
metadata:
  name: modules
spec:
  collectors:
    - kernelModules: {}
  analyzers:
    - kernelModules:
        outcomes:
          - pass:
              when: "target_core_mod,target_core_file,tcm_loop == loaded,loadable"
              message: The 'target_core_mod', target_core_file', and 'tcm_loop' kernel modules are loaded or loadable
```
Where it will pass if all of `target_core_mod`, `target_core_file`, and `tcm_loop` are either `loaded` or `loadable`.


croomes:
Added support for multiple results (#377)

croomes:
This isn't working as intended - the check only runs on the local node rather than the k8s nodes (I misunderstood what `HostPreflight` does).

I'm looking into using a runner to exec it remotely.  Any suggestions appreciated!

divolgin:
@croomes There isn't a way to do this with troubleshoot.  You'd have to wrap it into a script and run it remotely using ssh commands.  If you are talking about nodes that are all a part of a kubernetes cluster, you _could_ have a daemon set that runs a pod on each node, and then use the Exec collector to run lsmod inside the containers.  I've tried using lsmod inside the containers and it worked.  Not really sure if it works always.

croomes:
Replaced by https://github.com/replicatedhq/troubleshoot/pull/392

PR #382, opened 2021-07-07
Log progress when interactive=false

emosbaugh:
This will print progress to stderr. Example below:

```$ ./bin/preflight host-preflight.yaml --interactive=false 
[Kubernetes API Server Load Balancer] Running collector...
failed to run collector: Kubernetes API Server Load Balancer: failed to dial: dial tcp: lookup asdf@adf: no such host
[CPU Info] Running collector...
[Amount of Memory] Running collector...

   --- FAIL: Kubernetes API Server Load Balancer
      --- Analyzer Failed: analyze: failed to read collected file name: tcpLoadBalancer/Kubernetes API Server Load Balancer.json: file tcpLoadBalancer/Kubernetes API Server Load Balancer.json was not collected
--- FAIL   kurl-builtin
FAILED```

### DIFF
--- a/cmd/preflight/cli/run.go
+++ b/cmd/preflight/cli/run.go
@@ -139,7 +139,7 @@ func runPreflights(v *viper.Viper, arg string) error {
 					if !ok {
 						return
 					}
-					fmt.Fprintln(os.Stderr, msg)
+					fmt.Fprintf(os.Stderr, "%v\n", msg)
 				case <-finishedCh:
 					return
 				}
--- a/cmd/preflight/cli/run.go
+++ b/cmd/preflight/cli/run.go
@@ -135,10 +135,11 @@ func runPreflights(v *viper.Viper, arg string) error {
 		go func() {
 			for {
 				select {
-				case _, ok := <-progressCh:
+				case msg, ok := <-progressCh:
 					if !ok {
 						return
 					}
+					fmt.Fprintln(os.Stderr, msg)
 				case <-finishedCh:
 					return
 				}

### CODE REVIEW

**Review by laverya — APPROVED** [review]

File: cmd/preflight/cli/run.go [hunk 1]
```diff
@@ -135,10 +135,11 @@ func runPreflights(v *viper.Viper, arg string) error {
 		go func() {
 			for {
 				select {
-				case _, ok := <-progressCh:
+				case msg, ok := <-progressCh:
 					if !ok {
 						return
 					}
+					fmt.Fprintln(os.Stderr, msg)
```
murphybytes: progressCh should probably be a string type.  Since it's an interface this could panic if you sent it a nil. Maybe use fmt.Fprintf(os.Stderr, "%v" , msg) instead

murphybytes (commented):
This looks good although there are some problems with the way that channels are used here.  First, finishedCh shouldn't be a buffered channel and you should wait for all goroutines to exit cleanly before going on and exiting the program.  Having said all that, it will work the way that it is.  I don't think progressCh should be an interface though
```golang
var wait sync.WaitGroup
finishedCh := make(chan struct{})  
// progressCh shouldn't be an interface{} it could be if you needed it to be but it's dangerous // for a lot of different reasons, and looking at the code, I think a string would suffice
progressCh := make(chan string)
defer func() {
  close(finishedCh)
  // wait for goroutines to exit
  wait.Wait()
  close(progressCh)
}()

wait.Add()
go func() {
  defer wait.Done()
  select {
  case msg <- progressCh:  // don't need ok 
    fmt.Fprint
  case <-finishedCh:
    return
 }
 ```

cmd/preflight/cli/run.go
emosbaugh: updated

emosbaugh:
> This looks good although there are some problems with the way that channels are used here. First, finishedCh shouldn't be a buffered channel and you should wait for all goroutines to exit cleanly before going on and exiting the program. Having said all that, it will work the way that it is. I don't think progressCh should be an interface though
> 
> ```go
> var wait sync.WaitGroup
> finishedCh := make(chan struct{})  
> // progressCh shouldn't be an interface{} it could be if you needed it to be but it's dangerous // for a lot of different reasons, and looking at the code, I think a string would suffice
> progressCh := make(chan string)
> defer func() {
>   close(finishedCh)
>   // wait for goroutines to exit
>   wait.Wait()
>   close(progressCh)
> }()
> 
> wait.Add()
> go func() {
>   defer wait.Done()
>   select {
>   case msg <- progressCh:  // don't need ok 
>     fmt.Fprint
>   case <-finishedCh:
>     return
>  }
> ```

and i would welcome a PR!

murphybytes (approved):

PR #383, opened 2021-07-08
Fixed bugs introduced by handling multiple results in host preflights

murphybytes:
Fixed bug caused by host preflights not handling empty when clauses, this cropped up because we now handle multiple host preflight results. Also expanded test coverage and added integration test script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ vendor
 
 dist
 try.sh
+
+.vscode/
+workspace.*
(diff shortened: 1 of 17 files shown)


areed (commented):
Why do memory and cpu only return 1 result but others return multiple?

murphybytes:
> Why do memory and cpu only return 1 result but others return multiple?

@areed because in both these cases you have a single mutually exclusive outcome, for example, in the memory case FAIL you don't have enough memory; WARN you barely have enough; PASS you have enough.  The same sort of reasoning applies to CPU.  In the other cases it makes sense to have multiple outcomes, warnings for example as in the case of time where your clock isn't synchronized and the timezone isn't UTC

areed (approved):

PR #384, opened 2021-07-08
ConfigMap collector and secrets can be collected by selectors

emosbaugh:
1. Add configmap collector by name or selector
2. Port configmap collector code back to secret collector
3. Make sure that the file tree matches the documentation on troubleshoot.sh


pkg/analyze/configmap.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	result := AnalyzeResult{
+		Title:   title,
+		IconKey: "kubernetes_analyze_configMap",
+		IconURI: "https://troubleshoot.sh/images/analyzer-icons/configMap.svg?w=13&h=16",
+	}
+
+	var failOutcome *troubleshootv1beta2.Outcome
+	for _, outcome := range analyzer.Outcomes {
```
laverya, line 43: could we use the standard analyzer outcomes list code instead, with support for more outcomes than "fail"?

laverya: or am I just misunderstanding what this analyzer is supposed to be able to analyze

pkg/collect/configmap.go
```diff
@@ -0,0 +1,147 @@
...
+	Namespace       string `json:"namespace"`
+	Name            string `json:"name"`
+	Key             string `json:"key"`
+	ConfigMapExists bool   `json:"configMapExists"`
+	KeyExists       bool   `json:"keyExists"`
+	Value           string `json:"value,omitempty"`
+}
+
+func ConfigMap(c *Collector, configMapCollector *troubleshootv1beta2.ConfigMap) (map[string][]byte, error) {
+	client, err := kubernetes.NewForConfig(c.ClientConfig)
```
laverya: we should REALLY refactor troubleshoot at some point to pass in a client instead of generating one each time

(this would also allow for better testing)

laverya (approved):

pkg/analyze/configmap.go
emosbaugh: good feedback although i didnt write this code. i copied it from secret analyzer. it works as documented now so i think its good enough for this pr.

pkg/collect/configmap.go
emosbaugh: yep

laverya: It would also result in not having the `wall of shame` when using doctl:
```
W0708 13:07:14.096840    5597 exec.go:282] constructing many client instances from the same exec auth config can cause performance problems during cert rotation and can exhaust available network connections; 3211 clients constructed calling "doctl"
```

murphybytes (approved):

laverya (approved):

laverya (approved):

PR #386, opened 2021-07-13
XFS ftype collector and analyzer

areed:


pkg/collect/host_xfs_info_linux.go
```diff
@@ -0,0 +1,113 @@
+package collect
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"syscall"
+	"unsafe"
```
laverya, line 9: I'm really not a fan of the use of unsafe here
Can we just try execing `xfs_info` or something?

PR #387, opened 2021-07-13
Feature/validate tcp load balancer address

kwsorensen:


kwsorensen:
Test file used

```
kind: HostPreflight
metadata:
  name: kurl-builtin
spec:
  collectors:
    - tcpLoadBalancer:
        collectorName: "Kubernetes API Server Load Balancer"
        port: 16443
        address: asdf@adf
        exclude: 'false'
  analyzers:
    - tcpLoadBalancer:
        checkName: "Kubernetes API Server Load Balancer"
        collectorName: "Kubernetes API Server Load Balancer"
        exclude: 'false'
        outcomes:
          - fail:
              when: "invalid-address"
              message: The address {{kurl .Installer.Spec.Kubernetes.LoadBalancerAddress }} is not valid.
          - warn:
              when: "connection-refused"
              message: Connection to {{kurl .Installer.Spec.Kubernetes.LoadBalancerAddress }} via load balancer was refused.
          - warn:
              when: "connection-timeout"
              message: Timed out connecting to {{kurl .Installer.Spec.Kubernetes.LoadBalancerAddress }} via load balancer. Check your firewall.
          - warn:
              when: "error"
              message: Unexpected port status
          - warn:
              when: "address-in-use"
              message: Port 6443 is unavailable
          - pass:
              when: "connected"
              message: Successfully connected to {{kurl .Installer.Spec.Kubernetes.LoadBalancerAddress }} via load balancer
          - warn:
              message: Unexpected port status
```

Makes output

```./bin/preflight sample-troubleshoot.yaml --interactive=false
[Kubernetes API Server Load Balancer] Running collector...
[CPU Info] Running collector...
[Amount of Memory] Running collector...

   --- FAIL: Kubernetes API Server Load Balancer
      --- The address {{kurl .Installer.Spec.Kubernetes.LoadBalancerAddress }} is not valid.
   --- WARN: Kubernetes API Server Load Balancer
      --- Unexpected port status
--- FAIL   kurl-builtin
FAILED```

File: pkg/collect/host_network.go [hunk 1]
```diff
@@ -20,13 +23,67 @@ const (
...
+		return false
+	}
+
+	// Checking for uppercase letters
+	if strings.ToLower(hostAddress) != hostAddress {
+		return false
+	}
+
+	// Checking if it's all numbers and .
+	if isValidIPCandidate(hostAddress) {
```
areed: I don't think you need this candidate function. You can do
```
if len(validation.IsValidIP(hostAddress)) == 0 {
  return true
}
```
and then go on to validate as hostname if not.

kwsorensen: That sounds like a great idea!

kwsorensen: Forgot that without that validation check it would actually fail the unit tests we created.

I created a branch and PR with the suggested changes as proof. https://github.com/replicatedhq/troubleshoot/pull/388

Basically the k8s api validation check is too loose.

areed: In that case what about using `net.ParseIP`?

kwsorensen: The issue is that https://github.com/replicatedhq/troubleshoot/blob/feature/validate-tcp-load-balancer-address/pkg/collect/host_network.go#L76 returns cases of false positives. 

If it goes to that function it will pass if it contains all digits and '.' characters. The only valid configuration for that is if it is an IP address. So that's why the inside of the IF statement returns a fail if it's not an IP but should be.

pkg/collect/host_network.go
```diff
@@ -57,17 +113,23 @@ func checkTCPConnection(progressChan chan<- interface{}, listenAddress string, d
 	}()
 
 	stopAfter := time.Now().Add(timeout)
+	fmt.Printf("Timeout")
```
areed: debug

pkg/collect/host_network.go
```diff
@@ -57,17 +113,23 @@ func checkTCPConnection(progressChan chan<- interface{}, listenAddress string, d
...
 
 	for {
 		if time.Now().After(stopAfter) {
+			fmt.Printf("Timeout")
```
areed: debug

pkg/collect/host_network.go
```diff
@@ -57,17 +113,23 @@ func checkTCPConnection(progressChan chan<- interface{}, listenAddress string, d
...
+
 			return NetworkStatusConnectionTimeout, nil
 		}
 
 		conn, err := net.DialTimeout("tcp", dialAddress, 50*time.Millisecond)
 		if err != nil {
+			fmt.Printf("Error: %s", err)
```
areed: debug

pkg/collect/host_network.go
```diff
@@ -20,13 +23,67 @@ const (
...
+	NetworkStatusInvalidAddress       = "invalid-address"
 )
 
 type NetworkStatusResult struct {
-	Status NetworkStatus `json:"status"`
+	Status  NetworkStatus `json:"status"`
+	Message string        `json:"message"`
+}
+
+func isValidIPCandidate(address string) bool {
```
areed: Why not use a regex?

kwsorensen: Mostly because of an unfamiliarity with regex. Not sure if we prefer regex vs code/algorithms.

pkg/collect/host_network.go
```diff
@@ -63,7 +54,8 @@ func isValidLoadBalancerAddress(address string) bool {
 	}
 
 	// Checking if it's all numbers and .
-	if isValidIPCandidate(hostAddress) {
+	var ipRegexp = regexp.MustCompile(`^[0-9.]+$`)
```
areed: This should be at the top of the file just after the constants so it fails immediately. If it's in here you'll have to use `Compile` and check the returned error is nil.

areed (approved):

PR #397, opened 2021-07-28
Stop longhorn false positives on no results.

kwsorensen:


areed (approved):
Want to add a test case here? https://github.com/replicatedhq/troubleshoot/blob/master/pkg/analyze/longhorn_test.go#L386

PR #398, opened 2021-07-28
Default result only when no other result exists for http preflight

murphybytes:
Addresses [this issue](https://app.clubhouse.io/replicated/story/36192/one-host-preflight-still-shows-pass-and-warn-for-the-same-check)


laverya (approved):
This will now only ever return a single result instead of multiple results

PR #405, opened 2021-08-12
fix: allow passing empty namespace again for clusterResources

dkoshkin:
[A recent commit](https://github.com/replicatedhq/troubleshoot/pull/391/commits/2889e1ffe79ce3db4d3e5b48726aae5240cbbe8e) broke previous behavior of being able to collect `clusterResources: {}` from all namespaces.

From the [docs](https://troubleshoot.sh/docs/collect/cluster-resources/)
> This will attempt to collect information from all namespaces

Tested with:
```
kind: SupportBundle
metadata:
  name: namespace-bug
spec:
  collectors:
    - clusterInfo: {}
    - clusterResources: {}
```

Ran it with this fix (see output snippet):
```
➜ tree support-bundle-2021-08-12T16_45_04/cluster-resources
support-bundle-2021-08-12T16_45_04/cluster-resources
├── deployments
│   ├── calico-system.json
│   ├── default.json
│   ├── kube-node-lease.json
│   ├── kube-public.json
│   ├── kube-system.json
│   ├── node-feature-discovery.json
│   └── tigera-operator.json
```

And with latest release
```
➜ ./support-bundle version
Replicated Troubleshoot 0.13.7
➜  tree support-bundle-2021-08-12T16_46_11/cluster-resources/
support-bundle-2021-08-12T16_46_11/cluster-resources/
├── deployments
│   └── default.json
```

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -48,10 +48,6 @@ func runTroubleshoot(v *viper.Viper, arg string) error {
 	}
 
 	namespace := v.GetString("namespace")
-	if namespace == "" {
-		kubeconfig := k8sutil.GetKubeconfig()
-		namespace, _, _ = kubeconfig.Namespace()
-	}
 
 	var sinceTime *time.Time
 	if v.GetString("since-time") != "" || v.GetString("since") != "" {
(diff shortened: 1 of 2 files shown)


xavpaice:
This has been over a year since submission, and I see that there are now several similar fixes in place already (though, not for the clusterResources collector).

I'll close this out now, if the issue is still current we should raise a clear bug and it's a simpler fix than this one.

PR #414, opened 2021-09-06
Add signed SBOM to troubleshoot

murphybytes:
See [Clubhouse Issue](https://app.clubhouse.io/replicated/story/37032/when-troubleshoot-is-released-an-sbom-will-be-generated-and-included-in-the-release-notes) This change will generate a signed software bill of materials and add it to the repository release archives when the project is released.    Note: before merge a repository owner will need to add two secrets to the repository by a repository owner. These secrets are `COSIGN_KEY` and `COSIGN_PASSWORD`.  The `COSIGN_KEY` is a base 64 encoded private key, and the `COSIGN_PASSWORD` is the password used to generate the key.  [cosign](https://github.com/sigstore/cosign) is used to generate these as follows:
```
cosign generate-key-pair
cat cosign.key | base64 | pbcopy
```

### DIFF
--- a/.github/workflows/build-test-deploy.yaml
+++ b/.github/workflows/build-test-deploy.yaml
@@ -114,7 +114,9 @@ jobs:
         with:
           go-version: "1.14"
 
-      - uses: sigstore/cosign-installer@v1.1.0
+      - uses: sigstore/cosign-installer@main
+        with:
+          cosign-release: "v1.2.1"
 
       - name: Generate SBOM
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/build-test-deploy.yaml
```diff
@@ -113,6 +113,17 @@ jobs:
       - uses: actions/setup-go@v1
         with:
           go-version: "1.14"
+
+      - uses: sigstore/cosign-installer@main
+        with:
+          cosign-release: "v1.2.1"
```
DanStough, line 119: Do we care what version we use? No strong opinions, but I guess I'd rather have it use latest to know if there are upcoming issues.

DanStough (approved):

PR #419, opened 2021-09-16
make non-interactive `support-bundle` output more machine readable

danbudris:
when using the `interactive=false` flag of `support-bundle`, the spinner would still spin and the archive path and analysis output were kind of smooshed together with the logs. This made it harder to parse the output programatically. RE: https://github.com/replicatedhq/troubleshoot/issues/418

Now, if `--interactive=false`, `support-bundle` only prints each received collector callback message once, and the spinner does not spin 🎡 

Also,  this adds a key to the archivePath and analyzerOutput that are returned, `archivePath` and `analyzerOutput`, for easier programatic parsing. Note that `archivePath` is also appended to the output for `--interactive=true`.

### Sample Output
The analyzer results are abbreviated for clarity, but I think it tells the whole story. This basically just prints each message the collector sends to the callback channel, so for each collection operation it performs, from my reading (hence the 13 printings of "Collecting support bundle logs").

```bash
dan@dansMachine:~/git/supportbundle$ support-bundle bundle.yaml --kubeconfig ./cluster.kubeconfig --interactive=false
Collecting support bundle cluster-info
Collecting support bundle cluster-resources
W0916 00:56:37.916499  442923 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
Collecting support bundle secret
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs
Collecting support bundle logs

analyzerResults=[
    {
        "name": "capv.controller.manager.status",
        "labels": {
            "desiredPosition": "1",
            "iconKey": "kubernetes_deployment_status",
            "iconUri": "https://troubleshoot.sh/images/analyzer-icons/deployment-status.svg?w=17\u0026h=17"
        },
        "insight": {
            "name": "capv.controller.manager.status",
            "labels": {
                "iconKey": "kubernetes_deployment_status",
                "iconUri": "https://troubleshoot.sh/images/analyzer-icons/deployment-status.svg?w=17\u0026h=17"
            },
            "primary": "capv-controller-manager Status",
            "detail": "capv-controller-manager is running.",
            "severity": "debug"
        },
        "severity": "debug",
        "analyzerSpec": ""
    },
    {
[...]

Issue #418, opened 2021-09-15
`support-bundle --interactive=false` returns a more machine-readable output


danbudris:
**Describe the rationale for the suggested feature.**
the option `interactive=false` allows me to output the analysis result for my cluster in json. This could be useful for running the tool and parsing the output with automation.

However, the colored text and progress spinner (e.g. Collecting support bundle ⠏) are currently included in the output, and there is no easily parsable/programatic separation of the output analysis and archive path. This makes it unpredictable to parse the output from `support-bundle --interactive=false`. 

**Describe the feature**

`support-bundle --interactive=false`, or some other flag to `support-bundle`, returns clearly delineated, machine readable sections containing the analysis JSON, archive path, and the progress of collectors, without the animated progress spinner.

**Describe alternatives you've considered**

parsing the output myself -- the progress spinner makes it challenging to use the output systematically, especially when there is no clear, easily parsable delineation between the progress output and the analysis output.

danbudris:
https://github.com/replicatedhq/troubleshoot/pull/419

PR #422, opened 2021-09-21
fix(support-bundle): no client-go warnings or control chars if noninteractive.

DanStough:
Follow up to #419 and #418 .

- Removed control characters for spinner if `interactive=false`
- Remove warnings from client-go calls if `interactive=false`

Test case:
```
./bin/support-bundle ./examples/support-bundle/sample-supportbundle.yaml --interactive=false | jq
```

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -4,12 +4,10 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


divolgin (approved):

DanStough:
Non-interactive test case, use shell script as follows: 
```bash test.sh
echo $(./bin/support-bundle https://kots.io)
```

PR #423, opened 2021-09-21
Bump controller-gen to v0.7.0, which uses v1 CRD schema instead of v1beta1

divolgin:


jala-dx (approved):

sgalsaleh:
duplicates https://github.com/replicatedhq/troubleshoot/pull/428

PR #431, opened 2021-09-24
add an unique id to each host preflights

jala-dx:


pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
```diff
@@ -143,6 +143,7 @@ type RegistryImagesAnalyze struct {
 }
 
 type AnalyzeMeta struct {
+	ID        *string                `json:"id,omitempty" yaml:"id,omitempty"`
```
divolgin: Why a pointer and not just `string`?

divolgin (changes requested):
This is missing new schemas

pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
jala-dx: This is an implementation detail. if it is defined as pointer, then we can do spec.CPU.ID == nil, instead of spec.CPU.ID == "" . Empty string is still a valid string, so chose nil. @murphybytes can add more if i missed anything.

murphybytes: Because in Go a nil or non nil pointer unambiguously indicates whether or not a particular field is assigned, and if we supply the omitempty struct tag will cause the encoded output to not include `id: ""` if it is not assigned, which is what we want in this case.  If id is not a pointer it will show up in the output no matter what because and empty string is still a string.

murphybytes (approved):

jala-dx:
> This is missing new schemas

thank you. Added now

pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
divolgin: I understand what pointers do.  My concern is that this is the only string that's a pointer in this project and it's bound to be a cause of nil pointer panics in the future.

jala-dx: @divolgin @murphybytes and i discussed and we decided to keep it consistent with the rest of the spec.

PR #440, opened 2021-09-29
Better preflight validation

divolgin:


.github/workflows/build-test-deploy.yaml
```diff
@@ -56,7 +56,32 @@ jobs:
           path: bin/
       - uses: engineerd/setup-kind@v0.5.0
       - run: chmod +x bin/preflight
-      - run: ./bin/preflight --interactive=false --format=json https://preflight.replicated.com
+      - run: |
+          ./bin/preflight --interactive=false --format=json https://preflight.replicated.com > result.json
+          cat result.json
```
sgalsaleh, line 61: why is this needed?

.github/workflows/build-test-deploy.yaml
```diff
@@ -56,7 +56,32 @@ jobs:
...
+
+          EXIT_STATUS=0
+          if grep -q "was not collected" result.json; then
+            echo "Some files were not collected"
+            EXIT_STATUS=1
+          fi
+
+          if (( `jq '.pass | length' result.json` < 1 )); then
+            echo "No passing prelifhgts found"
```
sgalsaleh:
```suggestion
            echo "No passing preflights found"
```

.github/workflows/build-test-deploy.yaml
```diff
@@ -56,7 +56,32 @@ jobs:
...
+            EXIT_STATUS=1
+          fi
+
+          if (( `jq '.warn | length' result.json` < 1 )); then
+            echo "No warnings found"
+            EXIT_STATUS=1
+          fi
+
+          if (( `jq '.fail | length' result.json` < 1 )); then
+            echo "No failed prelifhgts found"
```
sgalsaleh:
```suggestion
            echo "No failed preflights found"
```

.github/workflows/build-test-deploy.yaml
divolgin: so we can see what's in that file.  because this test is kinds janky.

sgalsaleh (approved):

PR #442, opened 2021-09-30
Allow memory writers

divolgin:


pkg/collect/result.go
```diff
@@ -98,9 +102,10 @@ func (r CollectorResult) GetReader(bundlePath string, relativePath string) (io.R
 	return f, nil
 }
 
-func (r CollectorResult) GetWriter(bundlePath string, relativePath string) (io.WriteCloser, error) {
-	if r[relativePath] != nil {
```
divolgin, line 102: This if was bad, it would never evaluate to `true`.

sgalsaleh (approved):

PR #443, opened 2021-10-03
collector/analyzer for host operating system

jala-dx:
[SC](https://app.shortcut.com/replicated/story/37537/create-hostpreflight-collector-and-analyzer-for-operating-system)


pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
```diff
@@ -140,6 +143,7 @@ type HostCollect struct {
 	FilesystemPerformance *FilesystemPerformance `json:"filesystemPerformance,omitempty" yaml:"filesystemPerformance,omitempty"`
 	Certificate           *Certificate           `json:"certificate,omitempty" yaml:"certificate,omitempty"`
 	HostServices          *HostServices          `json:"hostServices,omitempty" yaml:"hostServices,omitempty"`
+	HostOS                *HostOS                `json:"hostos,omitempty" yaml:"hostos,omitempty"`
```
areed: The example shows `hostOS`

pkg/apis/troubleshoot/v1beta2/hostanalyzer_shared.go
```diff
@@ -107,4 +110,6 @@ type HostAnalyze struct {
 	Certificate *CertificateAnalyze `json:"certificate,omitempty" yaml:"certificate,omitempty"`
 
 	HostServices *HostServicesAnalyze `json:"hostServices,omitempty" yaml:"hostServices,omitempty"`
+
+	HostOS *HostOSAnalyze `json:"hostos,omitempty" yaml:"hostos,omitempty"`
```
areed: The example shows `hostOS`

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,206 @@
...
+	if err != nil {
+		return []*AnalyzeResult{&result}, errors.Wrap(err, "failed to get collected file")
+	}
+
+	var osInfo collect.HostOSInfo
+	if err := json.Unmarshal(contents, &osInfo); err != nil {
+		return []*AnalyzeResult{&result}, errors.Wrap(err, "failed to unmarshal host os info")
+	}
+
+	for _, outcome := range hostAnalyzer.Outcomes {
```
areed: All the other analyzers support warn. I think users would be surprised that warn isn't supported.

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,206 @@
...
+				result.URI = outcome.Fail.URI
+
+				return []*AnalyzeResult{&result}, nil
+			}
+			_, err := compareHostOSConditionalToActual(outcome.Fail.When, osInfo)
+			if err != nil {
+				result.IsFail = true
+				result.Message = outcome.Fail.Message
+				result.URI = outcome.Fail.URI
+				return []*AnalyzeResult{&result}, errors.Wrap(err, "unsupported distribution")
```
areed:
```suggestion
				return []*AnalyzeResult{&result}, errors.Wrap(err, "failed to compare host os with actual")
```

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	return false, fmt.Errorf("expected distribution does not match the supported %q", expectedVersion)
+}
+
+/*
+- fail
+    when: "ubuntu-16.04-kernel < 4.15"
+*/
+func bailIfUnsupportedKernel(operator string, expectedVersion string, actualVersion string) (bool, error) {
```
areed: This needs to be generic, not hard coded to check 4.15.

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,206 @@
...
+/*
+- pass
+    when: "ubuntu == 16.04"
+- pass
+    when: "ubuntu == 18.04"
+- fail
+    when: "ubuntu-16.04-kernel < 4.15"
+*/
+
+func bailIfUnsupportedOS(platform string, operator string, expectedVersion string, actualVersion string) (bool, error) {
```
areed: This needs to be generic, not have the kurl supported versions hard-coded.

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,206 @@
...
+    when: "ubuntu == 18.04"
+*/
+func compareHostOSConditionalToActual(conditional string, hostInfo collect.HostOSInfo) (bool, error) {
+	parts := strings.Split(conditional, " ")
+	if len(parts) != 3 {
+		return false, fmt.Errorf("Expected exactly 3 parts, got %d", len(parts))
+	}
+	// only for ubuntu 16.04, check if the kernel version is < 4.15
+	//https://kurl.sh/docs/install-with-kurl/system-requirements
+	if parts[0] == "ubuntu-16.04-kernel" {
```
areed: This needs to be generic so kernel versions can be checked for any distribution version. Also, the checks shouldn't be prefixed with `bail` since it's returning a boolean, not exiting.

pkg/analyze/host_os_info.go
jala-dx: Made changes to look for the source of truth from the spec (basically removed kurl specific dependencies)

pkg/analyze/host_os_info.go
jala-dx: Added a different error message "version out of range"  Let me know if this error message does not make sense.

examples/preflight/host/host-os.yaml
```diff
@@ -0,0 +1,130 @@
...
+  name: hostos
+spec:
+  collectors:
+    - hostOS : {}
+  analyzers:
+    - hostOS:
+        outcomes:
+          - warn:
+              when: "ubuntu == 16.04"
+              message: supported distribution
```
divolgin, line 19: ~this is deprecated, right?~ .... nm, this is just an example

jala-dx: yeah, thanks, that is why it is "warn"

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,143 @@
...
+		expectedVer := fixVersion(parts[2])
+		toleratedVer, err := semver.ParseTolerant(expectedVer)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to parse semver range %v", expectedVer)
+		}
+
+		when = fmt.Sprintf("%s %v", parts[1], toleratedVer)
+		whenRange, err := semver.ParseRange(when)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to parse semver range: %v", toleratedVer)
```
areed:
```suggestion
			return nil, errors.Wrapf(err, "failed to parse semver range: %s", when)
```

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,143 @@
...
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed to parse semver range: %v", fixedKernelVer)
+				}
+
+				result.Message = message
+				result.URI = uri
+				if whenRange(toleratedKernelVer) {
+					return []*AnalyzeResult{&result}, nil
+				} else {
+					return []*AnalyzeResult{&result}, fmt.Errorf("kernel versions mismatch: desired: %v actual: %v", platform, kernelInfo)
```
divolgin: This error is odd.  This looks like the outcome does not match, so it should not error out, but move on to the next one.

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+			result.Message = message
+			result.URI = uri
+			if whenRange(toleratedDistVer) {
+				return []*AnalyzeResult{&result}, nil
+			} else {
+				return []*AnalyzeResult{&result}, fmt.Errorf("distribution version out of range desired: %v, actual: %v", when, toleratedDistVer)
+			}
+		}
+		return []*AnalyzeResult{&result}, fmt.Errorf("unknown distribution %v : %v", platform, when)
```
divolgin: This doesn't look like it belongs here

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, errors.Wrapf(err, "failed to parse semver range %v", expectedVer)
+		}
+
+		when = fmt.Sprintf("%s %v", parts[1], toleratedVer)
+		whenRange, err := semver.ParseRange(when)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to parse semver range: %v", toleratedVer)
+		}
+
+		kernelInfo := fmt.Sprintf("%s-%s-kernel", osInfo.Platform, osInfo.PlatformVersion)
```
areed, line 100: Does the collector only return major.minor in `PlatformVersion`? It would never be something like `18.04.5`?

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,143 @@
...
+		if len(strings.Split(platform, "-")) == 3 && strings.Split(platform, "-")[2] == "kernel" {
+			if platform == kernelInfo {
+				fixedKernelVer := fixVersion(osInfo.KernelVersion)
+				toleratedKernelVer, err := semver.ParseTolerant(fixedKernelVer)
+				if err != nil {
+					return nil, errors.Wrapf(err, "failed to parse semver range: %v", fixedKernelVer)
```
areed: I don't think this is a range

pkg/analyze/host_os_info.go
areed: This should be a `continue` to the next outcome, not an error

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,143 @@
...
+			toleratedDistVer, err := semver.ParseTolerant(fixedDistVer)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to parse semver range")
+			}
+			result.Message = message
+			result.URI = uri
+			if whenRange(toleratedDistVer) {
+				return []*AnalyzeResult{&result}, nil
+			} else {
+				return []*AnalyzeResult{&result}, fmt.Errorf("distribution version out of range desired: %v, actual: %v", when, toleratedDistVer)
```
areed: This should be a `continue` to the next outcome, not a return

pkg/analyze/host_os_info.go
jala-dx: yeah it returns only major.minor

```

Platform infoStat {"hostname":"repldev-jalaja","uptime":23088112,"bootTime":1610903487,"procs":404,"os":"linux","platform":"ubuntu","platformFamily":"debian","platformVersion":"18.04","kernelVersion":"5.4.0-1034-gcp","kernelArch":"x86_64","virtualizationSystem":"","virtualizationRole":"guest","hostid":"8b4e7a59-17f1-433b-48a7-5eb2a46468ba"}

```

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,138 @@
...
+			if whenRange(toleratedDistVer) {
+				return []*AnalyzeResult{&result}, nil
+			}
+		}
+	}
+
+	return []*AnalyzeResult{&result}, nil
+}
+
+func fixVersion(versionStr string) string {
```
areed, line 129: Wouldn't this change `5.4.0-1034-gcp` to `5.4.-1034-gcp`? And `5.4.0` to `5.4.`?

jala-dx: yes, that was a bug:(. Wrote a new [func](https://github.com/replicatedhq/troubleshoot/pull/443/commits/0b4ca99b1603d35ed374c13c297b14a7b4e23ed0#diff-45a5637f6e8a6d0bcc48d007083c1db4d0b7e0a792659935e6f94c7504f96703R134-R145) to handle kernel version separately.  Also added an unit testcase to handle this case

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return strings.Join(splitStr, ".")
+}
+
+func fixKernelVersion(versionStr string) string {
+	splitStr := strings.Split(versionStr, ".")
+	for i := 0; i < len(splitStr); i++ {
+		splitStr[i] = strings.TrimPrefix(splitStr[i], "0")
+	}
+	if len(splitStr) >= 2 {
```
divolgin: This needs a comment with an example that show's what's happening here

pkg/collect/host_os_info.go
```diff
@@ -0,0 +1,52 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get os info")
+	}
+	hostInfo := HostOSInfo{}
+	hostInfo.Platform = infoStat.Platform
+	hostInfo.KernelVersion = infoStat.KernelVersion
+	hostInfo.Name = infoStat.Hostname
+	hostInfo.PlatformVersion = infoStat.PlatformVersion
+
+	fmt.Println("----------- Platform Version", infoStat.PlatformVersion)
```
divolgin: debug print outs

pkg/collect/host_os_info.go
```diff
@@ -0,0 +1,52 @@
...
+	fmt.Println("----------- Platform infoStat", infoStat)
+	b, err := json.Marshal(hostInfo)
```
divolgin: `MarshalIndent` would be better here

examples/preflight/host/host-os.yaml
```diff
@@ -0,0 +1,135 @@
...
+          - pass:
+              when: "ol == 8.3"
+              message: supported distribution
+          - pass:
+              when: "ol == 8.4"
+              message: supported distribution
+          - fail:
+              when: "ol > 8.4"
+              message: unsupported distribution
+          - fail:
```
areed, line 134: I know this example is only meant to document the usage, but these could never be true because you have `ubuntu == 16.04` and `ubuntu == 18.04` outcomes aboe.

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,146 @@
...
+func fixDistVersion(versionStr string) string {
+	splitStr := strings.Split(versionStr, ".")
+	for i := 0; i < len(splitStr); i++ {
+		splitStr[i] = strings.TrimPrefix(splitStr[i], "0")
+	}
+
+	return strings.Join(splitStr, ".")
+}
+
+func fixKernelVersion(versionStr string) string {
```
areed: What about `6.0.2`. Wouldn't that end up as `6.`? I think there are enough edge cases here to warrant a separate unit test.

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return []*AnalyzeResult{&result}, nil
+}
+
+func fixDistVersion(versionStr string) string {
+	splitStr := strings.Split(versionStr, ".")
+	for i := 0; i < len(splitStr); i++ {
+		splitStr[i] = strings.TrimPrefix(splitStr[i], "0")
```
areed: CentOS 8.0 would become `8.`.

pkg/analyze/host_os_info.go
jala-dx: thanks andrew for the regex pointers. fixed now

pkg/analyze/host_os_info_test.go
```diff
@@ -165,6 +196,60 @@ func TestAnalyzeHostOS(t *testing.T) {
...
+				Name:            "my-host",
+				KernelVersion:   "4.14",
+				PlatformVersion: "16.04",
+				Platform:        "ubuntu",
+			},
+			hostAnalyzer: &troubleshootv1beta2.HostOSAnalyze{
+				Outcomes: []*troubleshootv1beta2.Outcome{
+					{
+						Fail: &troubleshootv1beta2.SingleOutcome{
+							When:    "ubuntu-16.04-kernel < 4.14",
```
areed: If this test is passing there is a bug

pkg/analyze/host_os_info.go
```diff
@@ -122,25 +123,15 @@ func analyzeOSVersionResult(osInfo collect.HostOSInfo, outcomes []*troubleshootv
...
 		splitStr[i] = strings.TrimPrefix(splitStr[i], "0")
 	}
-	// kernel version can be suffixed, for ex., 5.4.0-1034-gcp, parse only the major and minor
-	if len(splitStr) >= 2 {
-		splitStr = splitStr[:2]
-		return strings.Join(splitStr, ".")
-	}
-
-	return ""
+	fixTrailZero := strings.Join(splitStr, ".")
```
areed, line 137: Could you add some tests for these versions: `0.1.2`, `1.0.2`, and `1.2.0`.

pkg/analyze/host_os_info_test.go
jala-dx: thank you, fixed it.

pkg/analyze/host_os_info_test.go
```diff
@@ -0,0 +1,347 @@
...
+			},
+			result: []*AnalyzeResult{
+				{
+					Title:   "Host OS Info",
+					IsFail:  true,
+					Message: "None matched, centos-18.04-kernel >= 4.15,ubuntu-18.04-kernel >= 4.15, supported distribution",
+				},
+			},
+		},
+		/*
```
areed: I'd either uncomment these or delete them

pkg/analyze/host_os_info.go
```diff
@@ -0,0 +1,141 @@
...
+			if err != nil {
+				return []*AnalyzeResult{&result}, errors.Wrapf(err, "failed to parse tolerant: %v", fixedDistVer)
+			}
+			if whenRange(toleratedDistVer) {
+				return []*AnalyzeResult{&result}, nil
+			}
+		}
+	}
+
+	return []*AnalyzeResult{nil}, nil
```
areed:
```suggestion
	return []*AnalyzeResult{}, nil
```

pkg/analyze/host_os_info_test.go
jala-dx: uncommented (meant to be:)

areed (approved):

PR #445, opened 2021-10-05
feat: add more detail to the ceph analyzer output

morningvera:
Adds some more detail to the ceph analyzer output. Specifically, the disk usage and whether it's classified as "nearly full" (>85%) or "full" (>95%). Previously it was only returning healthy, warning, or error.

[SH38120](https://app.shortcut.com/replicated/story/38120/improve-ceph-analyzer)


pkg/analyze/ceph.go
```diff
@@ -173,3 +227,18 @@ func compareCephStatus(actual, when string) (bool, error) {
...
+func detailedMessage(msg string, status CephStatus) string {
+	if status.OsdMap.OsdMap.Full {
+		msg = fmt.Sprintf("%s: OSD is full", msg)
+	} else if status.OsdMap.OsdMap.NearFull {
+		msg = fmt.Sprintf("%s: OSD is nearly full", msg)
+	}
+
+	if status.PgMap.TotalBytes > 0 {
+		pgUsage := 100 * float64(status.PgMap.UsedBytes) / float64(status.PgMap.TotalBytes)
+		msg = fmt.Sprintf("%s: PG usage is %.1f%%", msg, pgUsage)
```
morningvera: @jeffreygolden i basically went with what we discussed here but maybe we should change this to something a bit more human readable (e.g. "disk is nearly full: disk usage is 85%"). also lmk if you want some additional info in here.

pkg/analyze/ceph.go
```diff
@@ -173,3 +227,18 @@ func compareCephStatus(actual, when string) (bool, error) {
 		return false, errors.New("unknown operator")
 	}
 }
+
+func detailedMessage(msg string, status CephStatus) string {
+	if status.OsdMap.OsdMap.Full {
+		msg = fmt.Sprintf("%s: OSD is full", msg)
```
morningvera: i might change the `:` because i don't see any other cases where use that delimiter. `.` might make more sense.

morningvera:
finally got this working on my embedded cluster 🎉 

ceph-status

pkg/analyze/ceph.go
jeffreygolden: I'm torn on this. OSD and PG are the rook-ceph proprietary terms in their docs, giving somebody the best chance to look up the error and find more information.  Making it easy to understand up front is also important, and `disk` and `storage` are far more familiar terms.

morningvera: maybe `OSD storage is full` and `PG disk usage is 85%`?

pkg/analyze/ceph.go
morningvera: ended up making this change

pkg/analyze/ceph.go
```diff
@@ -173,3 +227,18 @@ func compareCephStatus(actual, when string) (bool, error) {
 		return false, errors.New("unknown operator")
 	}
 }
+
+func detailedMessage(msg string, status CephStatus) string {
+	if status.OsdMap.OsdMap.Full {
+		msg = fmt.Sprintf("%s. OSD disk is full", msg)
```
morningvera, line 204: this sort of thing could always result in messages like `Ceph is unhealthy.. OSD disk is full` but it's not a big deal imo. i could add a check to make it look cleaner if you think it's important though

jeffreygolden: The tests should catch this if somebody changes it unintentionally. I think we're OK without checking the string.

pkg/analyze/ceph.go
morningvera: yeah i'm thinking more for custom messages but i'm not sure how often that feature is used

pkg/analyze/ceph.go
morningvera: yeah i'm thinking more for custom messages but i'm not sure how often that feature is used

jeffreygolden (approved):

pkg/analyze/ceph.go
```diff
@@ -64,6 +65,30 @@ var CephStatusDefaultOutcomes = []*troubleshootv1beta2.Outcome{
...
+
+type HealthStatus struct {
+	Status string `json:"status"`
+}
+
+type OsdMap struct {
+	NumOsd   int  `json:"num_osds"`
+	NumUpOsd int  `json:"num_up_osds"`
+	Full     bool `json:"full"`
+	NearFull bool `json:"nearfull"`
```
marccampbell, line 84: should this be `near_full` for consistency?

morningvera: ideally, yes but we pull data via ceph directly so we have to follow their json tags. we used rook ceph as a reference here: https://github.com/rook/rook/blob/master/pkg/daemon/ceph/client/status.go#L108

PR #447, opened 2021-10-07
feat: Collect custom resources

### LINKED ISSUE #446: Collect custom resources as part of support bundle.
Hi Folks,
There is no way to collect the custom resources deployed on the k8s cluster using troubleshoot. This feature would be a nice to have since the community is making heavy use of extended objects to deploy operators and other applications.

rishabh96b:
Resolves: https://github.com/replicatedhq/troubleshoot/issues/446

## Overview
This change introduces the custom resources to be collected on a separate directory under the support bundle. Custom resources, once collected looks something like this
```
support-bundle-2021-10-07T02_47_49/custom-resources
├── crontabs.stable.example.com
│   └── test.json
└── elasticsearches.elasticsearch.k8s.elastic.co
    └── dev.json
    └── default.json
```

## Testing
The changes are tested manually by creating the support bundle from the `KIND` cluster having some custom resources deployed.

Co-authored-by: Martin Hrabovcin(@mhrabovcin) <mhrabovcin@users.noreply.github.com>


divolgin:
This is a cool addition.  I'm a little worried about this being enabled by default because custom resources could contain sensitive data.  It might be better to have this as a separate optional collector.  Or add a flag, `includeCRs`, to this collector and default it to `false`.  The flag seems like the better option.

divolgin:
On second thought, I think we will explore adding more default redactors and keeping this enabled by default.  We just can't merge this just yet.

mhrabovcin:
Hi @divolgin, thanks for your thoughts. Would you be willing to merge an adjusted patch that would include a boolean flag to optionally enable the custom resources in the diagnostics bundle with the default set to `false`?

areed:
I converted the output to yaml so that we can use the yamlPath redactor on these. Example:
```
kind: Redactor
metadata:
  name: rook-object-store-user
spec:
  redactors:
  - name: rook-object-store-user
    fileSelector:
      file: cluster-resources/custom-resources/cephobjectstoreusers.ceph.rook.io/*
    removals:
      yamlPath:
      - "*.spec.store"
 ```

Other changes:

- prefer apiextensions/v1 client over v1beta1
- rebase and fix conflicts
- move output into cluster-resources directory

How does everyone feel about merging this as a yaml collector enabled by default since it's compatible with the yamlPath redactor?

divolgin:
@areed that sounds good to me.  We should also include any additional default redactors for this if needed for kurl and OpenShift clusters at least.

areed:
The only CR I found in kurl or OpenShift clusters that contain sensitive data was `installers.cluster.kurl.sh`. I addded some default redactors for that. OpenShift CRs contain a lot of references to secret names, but never have secrets embedded directly in the CR.

divolgin (approved):

areed:
Thanks @rishabh96b!

PR #456, opened 2021-10-25
ClusterPodStatuses analyzer

sgalsaleh:


pkg/analyze/cluster_pod_statuses.go
```diff
@@ -0,0 +1,141 @@
...
+				r.URI = outcome.Pass.URI
+				results = append(results, &r)
+				whens = append(whens, outcome.Pass.When)
+			}
+		}
+
+		fmt.Println("whens", whens)
+
+		if len(results) == 0 {
+			return nil, errors.New("empty outcomes")
```
divolgin: this should be not an error, right?

pkg/analyze/cluster_pod_statuses.go
```diff
@@ -0,0 +1,141 @@
...
+			}
+			pods = append(pods, nsPods...)
+		}
+	}
+
+	results := []*AnalyzeResult{}
+
+	for _, pod := range pods {
+		whens := []string{}
+		results := []*AnalyzeResult{}
```
syntastical: Don't think you meant to shadow the variable in the outer scope

divolgin (approved):

schemas/collector-troubleshoot-v1beta2.json
```diff
@@ -432,33 +432,6 @@
                   }
                 }
               },
-              "kernelModules": {
```
syntastical, line 435: I take it that this is just some older unrelated stuff that needed to be cleaned up

PR #457, opened 2021-10-25
Include pod logs for pods that are failing

divolgin:


troubleshoot.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: troubleshoot.sh/v1beta2
```
sgalsaleh: what is this file

sgalsaleh (approved):

PR #474, opened 2021-10-28
38798: Adding node status check

syntastical:


pkg/longhorn/util/iscsi.go
```diff
@@ -1,3 +1,4 @@
+//go:build !windows
```
divolgin, line 1: what requires this change?  pretty sure `+build` works...

divolgin (approved):

pkg/longhorn/util/iscsi.go
syntastical: So this was auto generated and I actually had to search out the docs to figure out why and here's the most important line from the docs, "Go versions 1.16 and earlier used a different syntax for build constraints, with a "// +build" prefix. The gofmt command will add an equivalent //go:build constraint when encountering the older syntax. "

PR #476, opened 2021-10-28
a function to get a pod details from the support bundle

sgalsaleh:


pkg/supportbundle/parse.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			break
+		}
+	}
+
+	return &podDetails, nil
+}
+
+// GetFilesContents will return the file contents for filenames matching the filenames parameter.
+func GetFilesContents(bundleArchive string, filenames []string) (map[string][]byte, error) {
```
sgalsaleh, line 75: this is copied from https://github.com/replicatedhq/kots, kots will be updated to use this instead

divolgin (approved):

PR #506, opened 2021-12-07
host system packages collector/analyzer

sgalsaleh:


pkg/collect/host_system_package.go
```diff
@@ -0,0 +1,91 @@
...
+		var cmd *exec.Cmd
+		var stdout, stderr bytes.Buffer
+
+		switch releaseID {
+		case "ubuntu":
+			cmd = exec.Command("dpkg", "-s", p)
+		case "centos", "rhel", "amzn", "ol":
+			cmd = exec.Command("yum", "list", "installed", p)
+		default:
+			return nil, errors.New("unsupported distribution")
```
divolgin: add `releaseID` to error

pkg/analyze/host_system_packages.go
```diff
@@ -0,0 +1,164 @@
...
+				r.Message = outcome.Warn.Message
+				r.URI = outcome.Warn.URI
+				when = outcome.Warn.When
+			} else if outcome.Pass != nil {
+				r.IsPass = true
+				r.Message = outcome.Pass.Message
+				r.URI = outcome.Pass.URI
+				when = outcome.Pass.When
+			} else {
+				fmt.Println("error: found an empty outcome in a systemPackages analyzer") // don't stop
```
divolgin: probably shouldn't print here

sgalsaleh: it's for debugging purposes, otherwise we'd never know what happened or why there were no results in this case

pkg/analyze/host_system_packages.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func compareHostPackagesConditionalToActual(conditional string, isInstalled bool) (res bool, err error) {
+	if conditional == "" {
+		return true, nil
+	}
+
+	parts := strings.Split(conditional, " ")
+	if len(parts) != 1 {
+		return false, fmt.Errorf("invalid 'when' format: expected exactly 1 part, got %d", len(parts))
```
divolgin: this is weird... how about just adding a `default:` clause to the switch below to return this error since there are exactly two valid cases

divolgin (approved):

divolgin (approved):

divolgin (approved):

divolgin (dismissed):

pkg/analyze/host_system_packages.go
emosbaugh: can this print to stderr? otherwise this may make machine parseable output impossible

emosbaugh (changes requested):

pkg/analyze/host_system_packages.go
```diff
@@ -0,0 +1,159 @@
...
+
+			// template the title
+			titleTmpl, err := tmpl.Parse(a.Title())
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create new title template")
+			}
+			var t bytes.Buffer
+			err = titleTmpl.Execute(&t, info)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to execute template")
```
emosbaugh:
```suggestion
				return nil, errors.Wrap(err, "failed to execute title template")
```

pkg/analyze/host_system_packages.go
```diff
@@ -0,0 +1,159 @@
...
+
+			// template the message
+			msgTmpl, err := tmpl.Parse(r.Message)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create new title template")
+			}
+			var m bytes.Buffer
+			err = msgTmpl.Execute(&m, info)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to execute template")
```
emosbaugh:
```suggestion
				return nil, errors.Wrap(err, "failed to execute message template")
```

pkg/analyze/host_system_packages.go
```diff
@@ -0,0 +1,159 @@
...
+			err = titleTmpl.Execute(&t, info)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to execute template")
+			}
+			r.Title = t.String()
+
+			// template the message
+			msgTmpl, err := tmpl.Parse(r.Message)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create new title template")
```
emosbaugh:
```suggestion
				return nil, errors.Wrap(err, "failed to create new message template")
```

pkg/analyze/host_system_packages.go
```diff
@@ -0,0 +1,159 @@
...
+			// add to results, break and check the next pod
+			allResults = append(allResults, &r)
+			break
+		}
+	}
+
+	return allResults, nil
+}
+
+func templateString(value string, info collect.SystemPackageInfo) (string, error) {
```
emosbaugh: is this unused?

examples/preflight/host/system-packages.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: HostPreflight
+metadata:
+  name: packages
+spec:
+  collectors:
+    - systemPackages:
```
emosbaugh, line 7:
```
when: "OS=rhel8 and OS=centos8 and OS=ol8 and Name=blah"
```

config/crds/troubleshoot.sh_hostpreflights.yaml
```diff
@@ -951,6 +992,89 @@ spec:
...
+                          type: array
+                        amzn-2:
+                          items:
+                            type: string
+                          type: array
+                        centos:
+                          items:
+                            type: string
+                          type: array
+                        centos-7:
```
emosbaugh: this should be snake case `centos7`

emosbaugh (dismissed):

pkg/analyze/host_system_packages.go
```diff
@@ -115,33 +119,98 @@ func (a *AnalyzeHostSystemPackages) Analyze(getCollectedFileContents func(string
 	return allResults, nil
 }
 
-func isPackageInstalled(info collect.SystemPackageInfo) bool {
-	if info.Error != "" {
+func getSystemPackageTemplateMap(pkg collect.SystemPackage, osName string, osVersion string) map[string]interface{} {
+	m := map[string]interface{}{
+		"OS":          osName,
+		"OSVersion":   osVersion,
```
emosbaugh: how about OSVersionMajor so you can do int comparison?

pkg/analyze/host_system_packages.go
```diff
@@ -115,33 +119,98 @@ func (a *AnalyzeHostSystemPackages) Analyze(getCollectedFileContents func(string
...
+	}
+
+	var when bytes.Buffer
+	err = conditionalTmpl.Execute(&when, templateMap)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to execute when template")
 	}
 
-	return false, fmt.Errorf("invalid 'when' format: %s", conditional)
+	return when.String() == "true", nil
```
emosbaugh: how about parsebool?

pkg/textutil/util.go
```diff
@@ -0,0 +1,17 @@
+package textutil
+
+import (
+	"strings"
+	"unicode"
+)
+
+func StripSpaces(str string) string {
```
emosbaugh: can you use strings.ReplaceAll instead?

https://pkg.go.dev/strings#ReplaceAll

pkg/analyze/host_system_packages.go
```diff
@@ -115,33 +119,98 @@ func (a *AnalyzeHostSystemPackages) Analyze(getCollectedFileContents func(string
...
+			continue
+		}
+
+		// key shouldn't start with a space
+		if unicode.IsSpace([]rune(key)[0]) {
+			continue
+		}
+
+		// sanitize the key and value
+		key = textutil.StripSpaces(key) // remove all spaces, even the ones in between
```
emosbaugh: Why are you not just using replaceall here as well?

pkg/textutil/util.go
sgalsaleh: because i wanna strip all white space including `\t`, `\n`, etc.. and i couldn't find a way to convert `unicode.Space` to a string/[]rune

pkg/analyze/host_system_packages.go
sgalsaleh: i can add that too, but not in place of `OSVersion`

pkg/textutil/util.go
emosbaugh: ok

pkg/analyze/host_system_packages.go
emosbaugh: agreed

pkg/textutil/util.go
sgalsaleh: maybe it's not really necessary, i'll test some things out

pkg/analyze/host_system_packages.go
```diff
@@ -120,13 +120,23 @@ func (a *AnalyzeHostSystemPackages) Analyze(getCollectedFileContents func(string
 }
 
 func getSystemPackageTemplateMap(pkg collect.SystemPackage, osName string, osVersion string) map[string]interface{} {
+	osVersionParts := strings.Split(osVersion, ".")
```
emosbaugh, line 123: `strings.SplitN(osVersion, ".", 3)`

pkg/analyze/host_system_packages.go
```diff
@@ -120,13 +120,23 @@ func (a *AnalyzeHostSystemPackages) Analyze(getCollectedFileContents func(string
...
+	osVersionMajor, _ := strconv.ParseInt(osVersionParts[0], 10, 64)
+
+	var osVersionMinor int64 = -1
+	if len(osVersionParts) > 1 {
+		osVersionMinor, _ = strconv.ParseInt(osVersionParts[1], 10, 64)
```
emosbaugh, line 128: does this work for rhel based distros?

emosbaugh (approved):

pkg/analyze/host_system_packages.go
sgalsaleh: from the docs:
```
n > 0: at most n substrings; the last substring will be the unsplit remainder.
```
so how would that help?

pkg/analyze/host_system_packages.go
sgalsaleh: why wouldn't it? here's a sample map
```
map[Error: ExitCode:0 IsInstalled:true Name:iscsi-initiator-utils OS:rhel OSVersion:8.4 OSVersionMajor:8 OSVersionMinor:4]
```

sgalsaleh: one more:
```
map[Error: ExitCode:0 IsInstalled:true Name:libzstd OS:centos OSVersion:8 OSVersionMajor:8 OSVersionMinor:-1]
```

PR #515, opened 2021-12-17
Allow specifying namespaces when analyzing cluster resources

divolgin:


pkg/analyze/job_status.go
```diff
@@ -68,29 +68,37 @@ func analyzeOneJobStatus(analyzer *troubleshootv1beta2.JobStatus, getFileContent
 }
 
 func analyzeAllJobStatuses(analyzer *troubleshootv1beta2.JobStatus, getFileContents func(string) (map[string][]byte, error)) ([]*AnalyzeResult, error) {
-	var fileName string
+	fileNames := make([]string, 0)
```
syntastical, line 71: Would be nice to see some of these code expressed a bit more generically so we don't need to repeat it for each of these analyzers, but it out of the scope of this work.

syntastical (approved):

PR #522, opened 2022-01-10
detect ceph warning due to no redundancy

jala-dx:


pkg/analyze/ceph.go
```diff
@@ -204,6 +214,8 @@ func detailedCephMessage(msg string, status CephStatus) string {
 		msg = fmt.Sprintf("%s. OSD disk is full", msg)
 	} else if status.OsdMap.OsdMap.NearFull {
 		msg = fmt.Sprintf("%s. OSD disk is nearly full", msg)
+	} else if chkMsg, ok := status.Health.Checks["POOL_NO_REDUNDANCY"]; ok {
```
divolgin, line 227: How about instead of hardcoding this specific check, we iterate over all checks in health is not OK.

pkg/analyze/ceph.go
```diff
@@ -204,6 +214,8 @@ func detailedCephMessage(msg string, status CephStatus) string {
...
+		msg = fmt.Sprintf("%s. Pool no redundancy: %s", msg, chkMsg.Summary.Message)
```
divolgin, line 228: Line 223 overrides this message if condition is true.

diamonwiggins:
@jala-dx This was addressed with https://github.com/replicatedhq/troubleshoot/pull/575. Will go ahead and close this one out.

PR #530, opened 2022-02-02
Message templating and example

jeffreygolden:


xavpaice:
Hi, just following up on if this is still current?  If so, can you give us a bit of info on what the PR is about and if there's a related issue?  Just for background so we can review it properly.

PR #532, opened 2022-02-03
add --output-path flag to support-bundle

### LINKED ISSUE #407: The output path for Preflight (and Support-Bundle) Should be configurable
Right now the output of troubleshoot project CLIs go to the users home directory, regardless of where they run the tool. This is confusing and should also be configurable.

Thinking we should instead set the default to the current working directory and also add an output flag `-o/--output` flag to specify the path/filename.

### LINKED ISSUE #697: SupportBundle always saves to $HOME from the results screen
**Bug Description**

When running preflights from the CLI, at the results screen at the end, the user hits `s` to save and the results file is written to $CWD.  When running supportbundle from the CLI, however, it seems the results file is written to $HOME instead.  

**Expected Behavior**

The interfaces should be consistent in their behaviour, which should probably be to emit to $CWD.

**Steps To Reproduce**

run a minimal SupportBundle spec such as 

```
kind: SupportBundle
metadata:
  name: ping
spec:
  collectors:
    - runPod:
        name: "run-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["sh"]
            args: ['-c', 'echo SMTP server is reachable , PASS , blah blah message']
            imagePullPolicy: IfNotPresent
  analyzers:
    - textAnalyze:
        checkName: "run-ping"
        fileName: run-ping/run-ping.log
        regexGroups: 'SMTP server is reachable , (?P<Result>.*?) , (?P<Message>.*$)'
        outcomes:
          - pass:
              when: "Result == PASS"
              message: Passed
          - fail:
              when: "Result == FAIL"
              message: 'Result: {{ .Result }}'
```

from the CLI, and at the results screen, hit `s` to save results.  Observe the file gets written to `$HOME/support-bundle-results.txt`

**Additional Context**
```
~/.krew/bin/kubectl-support_bundle version
Replicated Troubleshoot 0.40.0
```

laverya:
fixes half of https://github.com/replicatedhq/troubleshoot/issues/407

### DIFF
--- a/pkg/supportbundle/supportbundle.go
+++ b/pkg/supportbundle/supportbundle.go
@@ -28,6 +28,7 @@ type SupportBundleCreateOpts struct {
 	SinceTime                 *time.Time
 	Redact                    bool
 	FromCLI                   bool
+	OutputPath                string
 }
 
 type SupportBundleResponse struct {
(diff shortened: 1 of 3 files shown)


OGtrilliams (approved):
saw it happen live, lgtm

pkg/supportbundle/supportbundle.go
```diff
@@ -57,9 +58,22 @@ func CollectSupportBundleFromSpec(spec *troubleshootv1beta2.SupportBundleSpec, a
...
-		basename = filepath.Join(os.TempDir(), basename)
+	basename := ""
+	if opts.OutputPath != "" {
+		// use override output path
+		overridePath, err := filepath.Abs(opts.OutputPath)
+		if err != nil {
+			return nil, errors.Wrap(err, "")
+		}
+
+		basename = filepath.Base(overridePath)
```
emosbaugh, line 69: this looks like it strips off the path and will always write to the current directory on line 129

https://github.com/replicatedhq/troubleshoot/pull/532/files#diff-04328572fa60f5b4cdaaa40835e6d53387d06e87f3e4becf95ab471678323bb6R129

xavpaice:
@laverya is this one still needed?

xavpaice:
Assigned @drohnow to look into this next week

laverya:
@xavpaice I think the original issue is still relevant, so yes

xavpaice:
Looks like this was completed with #538

PR #538, opened 2022-02-17
improvement: added --output flag for preflight and support bundle

### LINKED ISSUE #407: The output path for Preflight (and Support-Bundle) Should be configurable

cbodonnell:
Resolves: #407 

NOTE: Created a new file in `pkg/convert/output.go` with a func used by both preflight and support bundle.  Not sure if this is the right place for it.

### DIFF
--- a/cmd/preflight/cli/interactive_results.go
+++ b/cmd/preflight/cli/interactive_results.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


cmd/preflight/cli/interactive_results.go
```diff
@@ -217,8 +217,20 @@ func estimateNumberOfLines(text string, width int) int {
...
+	if outputPath != "" {
+		// use override output path
+		overridePath, err := convert.ValidateOutputPath(outputPath)
+		if err != nil {
+			return "", errors.Wrap(err, "override output file path")
+		}
+		filename = overridePath
+	} else {
+		// use default output path
+		filename = fmt.Sprintf("%s-results.txt", preflightName)
```
DanStough: Personally, I think it would be cool to use a timestamp in the filename to prevent overwrites, but this is just a suggestion.

DanStough (dismissed):

DanStough (approved):

PR #539, opened 2022-02-21
Add `strict` flag to Analyzers and ResultAnalyzers

pavansokkenagaraj:
Docs PR: https://github.com/replicatedhq/troubleshoot.sh/pull/364

- add `strict` flag to Analyzers.
```yaml
kind: Preflight
metadata:
spec:
  analyzers:
  - clusterVersion:
      strict: true # new field
      exclude: ""
      outcomes: 
        - fail: 
            when: "< 1.25.0"
            message: 1.21.0
            uri: https://www.kubernetes.io  
```

- add `strict` flag to ResultAnalyzers.
```go
type AnalyzeResult struct {
	IsPass bool
	IsFail bool
	IsWarn bool
	Strict bool  # new field
	...
}
```

-  Update stdout for `human` format
```

   --- FAIL: Required Kubernetes Version
      --- Strict: true # mapped new field
      --- 1.21.0
--- FAIL   example-preflight-checks
FAILED
```
-  Update stdout for `json` format
```json
{
  "fail": [
    {
      "title": "Required Kubernetes Version",
      "message": "1.21.0",
      "uri": "[https://www.kubernetes.io]()",
      "strict": true  # mapped new field
    }
  ]
}
```

- interactive result:
image

- downloaded result:
```
Example Preflight Checks Preflight Checks

Check FAIL
Title: Required Kubernetes Version
Message: 1.21.0
URI: https://www.kubernetes.io
Strict: true

------------

```
[example-preflight-checks-results-2022-02-21T15_27_22.txt](https://github.com/replicatedhq/troubleshoot/files/8111917/example-preflight-checks-results-2022-02-21T15_27_22.txt)


cmd/preflight/cli/stdout_results.go
```diff
@@ -90,6 +92,7 @@ func outputResult(analyzeResult *analyzerunner.AnalyzeResult) bool {
 		fmt.Printf("      --- %s\n", analyzeResult.Message)
 	} else if analyzeResult.IsFail {
 		fmt.Printf("   --- FAIL: %s\n", analyzeResult.Title)
+		fmt.Printf("      --- Strict: %t\n", analyzeResult.Strict)
```
laverya: maybe only output this if strict is true?
something like
` --- THIS FAILURE BLOCKS INSTALLATION` instead of ` --- Strict: true`

That also means we wouldn't change the current output without people having updated

pkg/analyze/analyzer.go
```diff
@@ -192,6 +200,9 @@ func Analyze(analyzer *troubleshootv1beta2.Analyze, getFile getCollectedFileCont
 		if err != nil {
 			return nil, err
 		}
+		for _, r := range results {
+			r.Strict = analyzer.DeploymentStatus.Strict.BoolVal
```
laverya: This doesn't work
You need to use the index instead, see https://goplay.tools/snippet/h_Zmfb3Faz1
```suggestion
		for idx, _ := range results {
			range[idx].Strict = analyzer.DeploymentStatus.Strict.BoolVal
```

cmd/preflight/cli/stdout_results.go
pavansokkenagaraj: Since `strict` can be consumed in any manner,
We (kots) are blocking the deployments when preflight's result `strict` is true
but for any other clients consuming support-bundle, `strict` can mean it different.

We would need a better message for the flag

laverya (dismissed):

cmd/preflight/cli/stdout_results.go
laverya: I still think "don't change the current output" is a better option

pavansokkenagaraj: Changed it. strict flags are displayed when strict is only `true`

laverya (approved):

garcialuis (approved):

PR #542, opened 2022-02-24
Fix `strict` flag mapping

pavansokkenagaraj:
fix `strict` flag mapping
- if `strict` is valid BoolString value, map accordingly
- else, map false as default and ignore parse error

### DIFF
--- a/pkg/multitype/boolstring.go
+++ b/pkg/multitype/boolstring.go
@@ -120,7 +120,10 @@ func (boolstr *BoolOrString) Fuzz(c fuzz.Continue) {
 
 // BoolOrDefaultFalse returns bool val, if strValu is parsed returns parsed value  else false as default when parse error
 func (boolstr *BoolOrString) BoolOrDefaultFalse() bool {
-	val, _ := boolstr.Bool()
+	val, err := boolstr.Bool()
+	if err != nil {
+		return false
+	}
 	return val
 }
 
--- a/pkg/multitype/boolstring.go
+++ b/pkg/multitype/boolstring.go
@@ -5,6 +5,7 @@ package multitype
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	fuzz "github.com/google/gofuzz"
 )
@@ -115,4 +116,22 @@ func (boolstr *BoolOrString) Fuzz(c fuzz.Continue) {
 		boolstr.BoolVal = false
 		c.Fuzz(&boolstr.StrVal)
 	}
+}
+
+// BoolOrDefaultFalse returns bool val, if strValu is parsed returns parsed value  else false as default when parse error
+func (boolstr *BoolOrString) BoolOrDefaultFalse() bool {
+	val, _ := boolstr.Bool()
+	return val
+}
+
+// Bool returns bool val, if strValu is parsed returns parsed value else false with parse error
+func (boolstr *BoolOrString) Bool() (bool, error) {
+	if boolstr.Type == Bool {
+		return boolstr.BoolVal, nil
+	}
+	parsed, err := strconv.ParseBool(boolstr.StrVal)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse bool string(err: %v)", err)
+	}
+	return parsed, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/multitype/boolstring_test.go
```diff
@@ -0,0 +1,94 @@
...
+		}, {
+			name:    "expect false when StrVal is 'true'",
+			fields:  fields{Type: String, StrVal: "true"},
+			want:    true,
+			wantErr: false,
+		}, {
+			name:    "expect false when StrVal is 'true'",
+			fields:  fields{Type: String, StrVal: "true"},
+			want:    true,
+			wantErr: false,
```
garcialuis: These are duplicated, one should be "expect true when StrVal is 'true'"?

pkg/multitype/boolstring.go
```diff
@@ -116,3 +117,21 @@ func (boolstr *BoolOrString) Fuzz(c fuzz.Continue) {
 		c.Fuzz(&boolstr.StrVal)
 	}
 }
+
+// BoolOrDefaultFalse returns bool val, if strValu is parsed returns parsed value  else false as default when parse error
+func (boolstr *BoolOrString) BoolOrDefaultFalse() bool {
+	val, _ := boolstr.Bool()
+	return val
+}
```
laverya:
```suggestion
func (boolstr *BoolOrString) BoolOrDefaultFalse() bool {
	val, err := boolstr.Bool()
	if err != nil {
		return false
	}
	return val
}
```
you shouldn't rely on the behavior of something returning non-nil error

garcialuis (changes requested):

pkg/multitype/boolstring_test.go
```diff
@@ -0,0 +1,94 @@
...
+		}, {
+			name:    "expect false when BoolVal is false",
+			fields:  fields{Type: Bool, BoolVal: false},
+			want:    false,
+			wantErr: false,
+		}, {
+			name:    "expect false when StrVal is 'true'",
+			fields:  fields{Type: String, StrVal: "true"},
+			want:    true,
+			wantErr: false,
```
laverya:
```suggestion
			name:    "expect false when StrVal is 'false'",
			fields:  fields{Type: String, StrVal: "false"},
			want:    false,
			wantErr: false,
```

pkg/multitype/boolstring_test.go
```diff
@@ -0,0 +1,94 @@
...
+			if (err != nil) != tt.wantErr {
+				t.Errorf("BoolOrString.Bool() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if got != tt.want {
+				t.Errorf("BoolOrString.Bool() = %v, want %v", got, tt.want)
+			}
+			if got := boolstr.BoolOrDefaultFalse(); got != tt.want {
+				t.Errorf("BoolOrString.BoolOrDefaultFalse() = %v, want %v", got, tt.want)
+			}
```
laverya:
```suggestion
			req := require.New(t)
			boolstr := &BoolOrString{
				Type:    tt.fields.Type,
				BoolVal: tt.fields.BoolVal,
				StrVal:  tt.fields.StrVal,
			}
			got, err := boolstr.Bool()
			if tt.wantErr {
				req.Error(err)
				return
			}
			req.NoError(err)
			req.Equal(tt.want, got)
			req.Equal(tt.want, boolstr.BoolOrDefaultFalse())
```

laverya: this uses testify/require, which you'd need to import, but we use elsewhere in the project

pkg/multitype/boolstring_test.go
pavansokkenagaraj: nice catch, thanks

pkg/multitype/boolstring_test.go
```diff
@@ -0,0 +1,89 @@
...
+			fields:  fields{Type: Bool, BoolVal: false},
+			want:    false,
+			wantErr: false,
+		}, {
+			name:    "expect false when StrVal is 'false'",
+			fields:  fields{Type: String, StrVal: "false"},
+			want:    false,
+			wantErr: false,
+		}, {
+			name:    "expect false when StrVal is 'true'",
```
garcialuis:
```suggestion
			name:    "expect true when StrVal is 'true'",
```

garcialuis (approved):

PR #551, opened 2022-03-23
map strict flag to analyze result

pavansokkenagaraj:
**map strict flag to analyze result when there is an error**
observe and ignore marshalling error when checking for strict flag in analyzer

### DIFF
--- a/pkg/preflight/analyze.go
+++ b/pkg/preflight/analyze.go
@@ -77,8 +77,10 @@ func doAnalyze(allCollectedData map[string][]byte, analyzers []*troubleshootv1be
 	for _, analyzer := range analyzers {
 		analyzeResult, err := analyze.Analyze(analyzer, getCollectedFileContents, getChildCollectedFileContents)
 		if err != nil {
+			strict, _ := HasStrictAnalyzer(analyzer)
 			analyzeResult = []*analyze.AnalyzeResult{
 				{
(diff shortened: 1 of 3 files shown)


sgalsaleh (approved):

pkg/preflight/analyze.go
```diff
@@ -77,8 +77,10 @@ func doAnalyze(allCollectedData map[string][]byte, analyzers []*troubleshootv1be
 	for _, analyzer := range analyzers {
 		analyzeResult, err := analyze.Analyze(analyzer, getCollectedFileContents, getChildCollectedFileContents)
 		if err != nil {
+			strict, _ := HasStrictAnalyzer(analyzer)
```
emosbaugh, line 80: If you have a reason to explicitly ignore this error can you please justify it with a comment in the code. Also it may be best to log the error. Thanks

pavansokkenagaraj: since we have already parsed and saved the preflight spec, when we unmarshall/process the spec again, we wouldn't be having parse errors.
moreover we would want to find if the spec.analyzers[].analyze[].analyzemeta.strict is true(atleast one is true), so we can ignore the errors here.

Yes, it would be better to log.

The other way to not address erroneous conditions is to use the traverse through the Preflight struct which is very complex.

PR #552, opened 2022-03-23
fix: return true when one of analyzers strict field is true

pavansokkenagaraj:
fix: return true when one of analyzers strict is true

### DIFF
--- a/pkg/preflight/analyze.go
+++ b/pkg/preflight/analyze.go
@@ -78,9 +78,9 @@ func doAnalyze(allCollectedData map[string][]byte, analyzers []*troubleshootv1be
 	for _, analyzer := range analyzers {
 		analyzeResult, err := analyze.Analyze(analyzer, getCollectedFileContents, getChildCollectedFileContents)
 		if err != nil {
-			strict, err := HasStrictAnalyzer(analyzer)
-			if err != nil {
-				logger.Printf("failed to determine if analyzer %v is strict: %s", analyzer, err)
+			strict, strictErr := HasStrictAnalyzer(analyzer)
+			if strictErr != nil {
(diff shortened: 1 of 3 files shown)


pkg/preflight/util.go
```diff
@@ -25,8 +25,11 @@ func HasStrictAnalyzers(preflight *troubleshootv1beta2.Preflight) (bool, error)
 	}
 
 	// analyzerMap will ignore empty Analyzers and loop around Analyzer with data
+	// hasStrictAnalyers:=false
```
sgalsaleh: cleanup?

sgalsaleh (dismissed):

pkg/preflight/analyze.go
```diff
@@ -77,7 +78,11 @@ func doAnalyze(allCollectedData map[string][]byte, analyzers []*troubleshootv1be
 	for _, analyzer := range analyzers {
 		analyzeResult, err := analyze.Analyze(analyzer, getCollectedFileContents, getChildCollectedFileContents)
 		if err != nil {
-			strict, _ := HasStrictAnalyzer(analyzer)
+			strict, err := HasStrictAnalyzer(analyzer)
```
emosbaugh: does this overwrite the error on 79 if nil?

emosbaugh: does this overwrite the error on 79 if nil?

emosbaugh (changes requested):

pkg/preflight/analyze.go
pavansokkenagaraj: @emosbaugh `Yes, it will.
but, now that I think of we should preserve analyzer error and return it as part of analyzeResult.
Updating the err checks.

emosbaugh (approved):

PR #556, opened 2022-04-01
Fix missing longhorn logs

emosbaugh:
Longhorn logs are always empty. Fix copy code.


pkg/collect/redact.go
```diff
@@ -26,9 +26,7 @@ func redactResult(bundlePath string, input CollectorResult, additionalRedactors
 				}
 				return errors.Wrap(err, "failed to get reader")
 			}
-			if r, ok := r.(io.ReadCloser); ok {
-				defer r.Close()
-			}
+			defer r.Close()
```
divolgin, line 29: this will panic. not all readers are closers

divolgin (approved):

PR #558, opened 2022-04-19
Allow entire PodSpec to be defined in Run Collector

diamonwiggins:


diamonwiggins:
Should we still support the creation of an image pull secret with this method?

dexhorthy:
Awesome! 🚀

pkg/collect/run.go
```diff
@@ -140,38 +140,55 @@ func runPod(ctx context.Context, client *kubernetes.Clientset, runCollector *tro
...
 			},
-		},
-	}
+		}
 
-	if runCollector.ImagePullSecret != nil && runCollector.ImagePullSecret.Data != nil {
-		secretName, err := createSecret(ctx, client, pod.Namespace, runCollector.ImagePullSecret)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to create secret")
+		if runCollector.ImagePullSecret != nil && runCollector.ImagePullSecret.Data != nil {
```
emosbaugh, line 183: is this useful when there is a podspec?

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -75,6 +76,7 @@ type Run struct {
 	ImagePullPolicy    string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
 	ImagePullSecret    *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
 	ServiceAccountName string            `json:"serviceAccountName,omitempty" yaml:"serviceAccountName,omitempty"`
+	PodSpec            corev1.PodSpec    `json:"podSpec,omitempty" yaml:"podSpec,omitempty"`
```
emosbaugh, line 79: The only thing i question is whether or not we should create an entirely new collector like RunPod... Did you consider that? Do you think reusing this is better?

diamonwiggins: I think reusing is worth it if it isn't confusing for users to discern which properties are used based on when you use `podSpec` or not. I do question that. 🤔

pkg/collect/run.go
diamonwiggins: when `podSpec` isn't used you'd still be able to use this feature, but planned to make it clear in the docs that with `podSpec` you'd always have to specify an existing secret name in the spec

Issue #555, opened 2022-03-29
MySQL Collector and Analyzer support checking for MySQL properties


rocktavious:
We have a business requirement that we need to run a preflight check to make sure a MySQL DB our app connects to has these parameters set

```
"character_set_server" = "utf8mb4"
"collation_server" = "utf8mb4_unicode_ci"
"init_connect" = "SET collation_connection = utf8mb4_unicode_ci,NAMES utf8mb4"
"innodb_file_format" = "Barracuda"
"innodb_large_prefix" = "1"
"innodb_strict_mode" = "1"
"log_bin_trust_function_creators" = "1"
```

It seems like we could adjust the https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/mysql.go to run an addtional query when configured of `SHOW VARIABLES` that it then stuffs in the json.

Then we could adjust the https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/mysql.go to be given a list of expected "key" == "value" stanza's which it could run through and ensure they all match or fail.

There probably some more to it but at first glance it seems quite easy to implement.  Would love any feedback - we are keen to have this so we could dedicate an engineer to implement this.

OGtrilliams:
Hello @rocktavious ! Thanks for reaching out. The issue has been added to the troubleshoot.sh [community agenda](https://hackmd.io/yZbotEHdTg6TfRZBzb8Tcg) for discussion. If you're available, it would be wonderful if you could join us so that we can discuss your goals & develop a plan to implement the desired changes. We will be meeting at 11AM PST on this Zoom link: https://replicated.zoom.us/j/83977996608?pwd=SmNobzF4V0puUTA5a0xZb1prdmthZz09

eochoa260:
Hi @OGtrilliams! @rocktavious is tentative for the meeting but I will be able to attend in his place to discuss if he's not able to make it.

emosbaugh:
We would welcome the contribution!

Capturing a discussion we had in the troubleshoot community meeting.

Adding additional data to the mysql collector makes sense.

Rather than reusing the mysql analyzer it may be best to use the text analyzer or adding a new key value analyzer.

It may make sense to add proposed spec changes here before implementation.

Thanks!

rocktavious:
The PR has been merged and we've tested it works correctly for our usecase.  Thanks everyone.

PR #562, opened 2022-04-27
Add Mysql variables to collector

eochoa260:
This change adds output of `SHOW VARIABLES` query to the mysql collector to solve issue #555.  Tested locally with the text analyzer and it's working as expected.


pkg/collect/mysql.go
```diff
@@ -27,6 +27,38 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
 			databaseConnection.IsConnected = true
 			databaseConnection.Version = version
 		}
+
+		rows, err := db.Query("SHOW VARIABLES")
```
rocktavious: This is more for replicated - but should this be "optional"?
If so how do we do that?

pkg/collect/mysql.go
```diff
@@ -27,6 +27,38 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
...
+		scanArgs := make([]interface{}, len(values))
+		for i := range values {
+			scanArgs[i] = &values[i]
+		}
+
+		variables := make(map[string]string)
+		for rows.Next() {
+			err = rows.Scan(scanArgs...)
+			if err != nil {
+				databaseConnection.Error = err.Error()
```
rocktavious, line 38: we should exit the loop here with `break`

pkg/collect/mysql.go
rocktavious: Also @eochoa260 - i think `rows` is an object that we need to defer cleanup on
```
defer rows.Close()
```

pkg/collect/mysql.go
```diff
@@ -27,6 +27,38 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
...
+
+		rows, err := db.Query("SHOW VARIABLES")
+		if err != nil {
+			databaseConnection.Error = err.Error()
+		}
+		columns, err := rows.Columns()
+		if err != nil {
+			databaseConnection.Error = err.Error()
+		}
+		values := make([]sql.RawBytes, len(columns))
```
rocktavious: I feel like we could clean this up.  I think `.Scan` can probably take a `map[string]interface{}`.

I'll do some tests tonight and report back.

rocktavious (changes requested):
Just a few things to take care of.  Good work overall.

rocktavious (dismissed):
LGTM now

pkg/collect/mysql.go
```diff
@@ -20,6 +20,8 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
 	} else {
 		query := `select version()`
 		row := db.QueryRow(query)
+		defer db.Close()
```
rocktavious: @eochoa260 - move this before the `QueryRow` to ensure its hooked up incase something crashes the deferred call will still happen and clean up

pkg/collect/mysql.go
```diff
@@ -15,18 +15,43 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
...
 		}
+
+		rows, err := db.Query("SHOW VARIABLES")
+		defer rows.Close()
+
+		if err != nil {
+			databaseConnection.Error = err.Error()
+		}
+
+		if err != nil {
```
divolgin: Line 36 is already doing this check

pkg/collect/mysql.go
```diff
@@ -15,18 +15,43 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
...
 		version := ""
 		if err := row.Scan(&version); err != nil {
 			databaseConnection.Error = err.Error()
 		} else {
 			databaseConnection.IsConnected = true
 			databaseConnection.Version = version
 		}
+
+		rows, err := db.Query("SHOW VARIABLES")
+		defer rows.Close()
```
divolgin: `defer rows.Close()` should happen after error check

pkg/collect/mysql.go
```diff
@@ -15,18 +15,43 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
...
+		if err != nil {
+			databaseConnection.Error = err.Error()
+		}
+
+		if err != nil {
+			databaseConnection.Error = err.Error()
+		}
+
+		variables := map[string]string{}
+		for rows.Next() {
```
divolgin: none of this should be executed if there is an error

pkg/collect/mysql.go
eochoa260: Oh yep, cleaned up some code here and must have left that one in there. Removing it now.

divolgin:
I think this is a great addition, but variables may contain sensitive data (like `report_password` below?).

```
mysql> SHOW VARIABLES like "%pass%";
+----------------------------------------------+-----------------+
| Variable_name                                | Value           |
+----------------------------------------------+-----------------+
| caching_sha2_password_auto_generate_rsa_keys | ON              |
| caching_sha2_password_digest_rounds          | 5000            |
| caching_sha2_password_private_key_path       | private_key.pem |
| caching_sha2_password_public_key_path        | public_key.pem  |
| default_password_lifetime                    | 0               |
| disconnect_on_expired_password               | ON              |
| generated_random_password_length             | 20              |
| mysql_native_password_proxy_users            | OFF             |
| password_history                             | 0               |
| password_require_current                     | OFF             |
| password_reuse_interval                      | 0               |
| report_password                              |                 |
| sha256_password_auto_generate_rsa_keys       | ON              |
| sha256_password_private_key_path             | private_key.pem |
| sha256_password_proxy_users                  | OFF             |
| sha256_password_public_key_path              | public_key.pem  |
+----------------------------------------------+-----------------+
16 rows in set (0.00 sec)
```

The default redactors may not pick this up.  One option is to make this data optional, for example

```
  collectors:
    - mysql:
        collectorName: mysql
        uri: ...
        includeVariables: true
```

rocktavious:
@divolgin 
Thats a great point.  Could we also configure it so you have to specify which variable names you want to pull back?  We aren't 100% up to speed on how to add extra configuration like this from the collector definition the user fills out.

Maybe something like
```
  collectors:
    - mysql:
        collectorName: mysql
        uri: ...
        withVariables:
        - character_set_server
        - collation_server
        - innodb_file_format
```

pkg/collect/database_shared.go
```diff
@@ -1,7 +1,13 @@
 package collect
 
+type MySQLVariable struct {
```
divolgin: There is also no need for this type

pkg/collect/mysql.go
```diff
@@ -18,15 +18,37 @@ func Mysql(c *Collector, databaseCollector *troubleshootv1beta2.Database) (Colle
...
+		rows, err := db.Query("SHOW VARIABLES")
+
+		if err != nil {
+			databaseConnection.Error = err.Error()
+		} else {
+			defer rows.Close()
+
+			variables := map[string]string{}
+			for rows.Next() {
+				variable := MySQLVariable{}
```
divolgin: This should just be `var key, value string`

eochoa260: Oh nice, thanks for the suggestion.  Just made that update.

eochoa260:
> I think this is a great addition, but variables may contain sensitive data (like `report_password` below?).
> 
> ```
> mysql> SHOW VARIABLES like "%pass%";

> +----------------------------------------------+-----------------+
> | Variable_name                                | Value           |
> +----------------------------------------------+-----------------+
> | caching_sha2_password_auto_generate_rsa_keys | ON              |
> | caching_sha2_password_digest_rounds          | 5000            |
> | caching_sha2_password_private_key_path       | private_key.pem |
> | caching_sha2_password_public_key_path        | public_key.pem  |
> | default_password_lifetime                    | 0               |
> | disconnect_on_expired_password               | ON              |
> | generated_random_password_length             | 20              |
> | mysql_native_password_proxy_users            | OFF             |
> | password_history                             | 0               |
> | password_require_current                     | OFF             |
> | password_reuse_interval                      | 0               |
> | report_password                              |                 |
> | sha256_password_auto_generate_rsa_keys       | ON              |
> | sha256_password_private_key_path             | private_key.pem |
> | sha256_password_proxy_users                  | OFF             |
> | sha256_password_public_key_path              | public_key.pem  |
> +----------------------------------------------+-----------------+
> 16 rows in set (0.00 sec)
> ```
> 
> The default redactors may not pick this up. One option is to make this data optional, for example
> 
> ```
>   collectors:
>     - mysql:
>         collectorName: mysql
>         uri: ...
>         includeVariables: true
> ```

@divolgin I made an update to only pull mysql variables that are in defined as parameters in the preflight manifest, e.g. - 
```spec:
  collectors:
    - mysql:
        collectorName: mysql
        uri: 'root:my-secret-pw@tcp(localhost:3306)/mysql'
        parameters:
          - character_set_server
          - collation_server
          - init_connect
          - innodb_file_format
          - innodb_large_prefix
          - innodb_strict_mode
          - log_bin_trust_function_creators
```
Look forward to any feedback you have on this though.

divolgin (approved):

emosbaugh:
I tested this with the provided yaml. It works with the following output. Merging.

```
$ cat support-bundle-2022-05-04T16_51_16/mysql/mysql.json 
{"isConnected":true,"version":"8.0.29","variables":{"character_set_server":"utf8mb4","collation_server":"utf8mb4_0900_ai_ci","init_connect":"","innodb_strict_mode":"ON","log_bin_trust_function_creators":"OFF"}}
```

PR #564, opened 2022-05-02
RunPod Collector

diamonwiggins:
Description

* Adding new `runPod` Collector that consumes a `corev1.PodSpec` instead of individual PodSpec properties
* Refactored old `run` collector to use new `runPod` collector code
* We currently document both `CollectorName` and `Name` for configuring the collector name, so I've added handling for both. Not sure if that's something we'd like to do, but currently if you use only `Name` for the `Run` collector it doesn't work. See - https://troubleshoot.sh/docs/collect/run/

Questions

* If a user specifies a secret to be created with `runPod.ImagePullSecret` as well as an existing secret name in `runPod.PodSpec`, I think the existing secret will be deleted in the cleanup also. Can't see a scenario where you'd need both, but should we protect against that just in case?
* The existing `run` collector is easier to write and simpler. Do we want to truly deprecate it, or just push those who need more than what it has implemented to `runPod`? 

Example

```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - runPod:
        collectorName: "ping"
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
            args: ["-w", "5", "www.github.com"]
```


emosbaugh (approved):
This is great. Thanks Diamon! Docs will need to be updated.

PR #574, opened 2022-05-12
Host Collectors in Support Bundles

diamonwiggins:
Proposal doc here: https://docs.google.com/document/d/1hgmySUTtLKS4SdEnwrWc8VfU35SFFzRb0TLt85UwlWc/edit

Only collectors and not analyzers are being added in this PR

example:
```yaml
kind: SupportBundle
metadata:
  name: collector-sample
spec:
  collectors: []
  hostCollectors:
    - diskUsage:
        path: /var/lib
    - cpu: {}
    - memory: {}
    - http:
        get:
          url: https://registry.replicated.com
    - ipv4Interfaces: {}
    - time: {}
    - blockDevices: {}
```


pkg/apis/troubleshoot/v1beta2/hostanalyzer_shared.go
```diff
@@ -1,13 +1,15 @@
 package v1beta2
 
 type CPUAnalyze struct {
-	AnalyzeMeta `json:",inline" yaml:",inline"`
-	Outcomes    []*Outcome `json:"outcomes" yaml:"outcomes"`
+	AnalyzeMeta   `json:",inline" yaml:",inline"`
+	CollectorName string     `json:"collectorName,omitempty" yaml:"collectorName,omitempty"`
```
divolgin, line 5: This change is a little unexpected.  How did this work without collector name?  Or is this a new dependency because of path changes?

sgalsaleh: looks like it was only defined for the analyzers that actually used this field, and now these analyzers are also updated to do that same.

sgalsaleh (approved):

divolgin (approved):

pkg/apis/troubleshoot/v1beta2/hostanalyzer_shared.go
diamonwiggins: Yeah, i was conflicted on this one honestly. Anywhere I added `CollectorName` it still isn't used, but was just thinking it was better to make all of the specs consistent since we usually document `CollectorMeta` and `AnalyzeMeta` as being default.

The collectors/analyzers where `CollectorName` was previously omitted are ones like `cpu`, `time`, `memory`, etc. These all will return the same result no matter how many times you run them so no point in specifying them multiple times hence needing to distinguish them by the `CollectorName`. 

If we think it's best to still let the user override what the filename could be when using the `CollectorName`, let me know. I don't have a strong opinion one way or the other.

PR #577, opened 2022-05-18
Template message in outcome block when using textAnalyzer

nachtmaar:
I started playing with the `kubectl trouble-shoot` plugin because I saw it yesterday on kubecon+cloudnativecon 😎 Thanks for presenting it 😃

One feature that was missing for me when using the `textAnalyzer` to parse some log files, is to use the **reg ex groups** also in the **message** which will be displayed to the user.

This is how the feature looks like (the words `wrong-eventtypeprefix` and `default` are coming from a reg ex group match):
Screenshot 2022-05-18 at 10 59 20 AM


mnp:
I came here to the issues looking for someone hacking on results.

It would be nice if the resource analyzer outcome messages could be templated.  So instead of "The cluster must contain at least 32 cores allocatable" you could see "The cluster must contain at least 32 cores allocatable, found 31.1".

nachtmaar:
> I came here to the issues looking for someone hacking on results.
> 
> It would be nice if the resource analyzer outcome messages could be templated. So instead of "The cluster must contain at least 32 cores allocatable" you could see "The cluster must contain at least 32 cores allocatable, found 31.1".

@mnp I think this can be achieved exactly the same way as in my PR except that the data input is coming from [corev1.Node{} struct](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/node_resources.go#L27)

The function `compareNodeResourceConditionalToActual` would have to return the node instead of a bool then.

alexanderjophus:
Hey @nachtmaar , thanks so much for your contribution, I'll take a look at this :)

> The function compareNodeResourceConditionalToActual would have to return the node instead of a bool then.

I believe this is on our radar

alexanderjophus (approved):
lgtm

nachtmaar:
@trelore What do I need to do to get the PR merged ? Do you have the permission to merge the PR ?

alexanderjophus:
> @trelore What do I need to do to get the PR merged ? Do you have the permission to merge the PR ?

@nachtmaar  Nothing, it's on me to merge it (which I plan on doing today). I had a minor concern that if someone uses template syntax (`{{ .something }}`) but didn't actually pass in the variable `something` in this case, it would break.

I spent a bit of time last week racking my brain to see if there's anything we can do about it, but I think we just merge it and go for it.

Thanks again for the contribution - apologies about taking my time.

alexanderjophus (approved):

alexanderjophus:
@nachtmaar I think you may need to sign a CLA for this to go in, looks like it's the only thing left :)

nachtmaar:
> @nachtmaar I think you may need to sign a CLA for this to go in, looks like it's the only thing left :)

Mhh, I did sign it before, but did it again :)

nachtmaar:
https://cla-assistant.io/check/replicatedhq/troubhelshoot?pullRequest=577

> > @trelore What do I need to do to get the PR merged ? Do you have the permission to merge the PR ?
> 
> @nachtmaar Nothing, it's on me to merge it (which I plan on doing today). I had a minor concern that if someone uses template syntax (`{{ .something }}`) but didn't actually pass in the variable `something` in this case, it would break.
> 
> I spent a bit of time last week racking my brain to see if there's anything we can do about it, but I think we just merge it and go for it.
> 
> Thanks again for the contribution - apologies about taking my time.

Good point. Did not think about this case yet 🤔 Maybe we could handle the error (not sure if there is a specific error for the case that a variable is missing) and return a more readable error for the user and tell him what the problem is.

nachtmaar:
@trelore Are you sure that you signed the CLA as well ? You merged a commit into the branch, so you are a co-author now :)
Screenshot 2022-06-07 at 1 41 02 PM

alexanderjophus:
> @trelore Are you sure that you signed the CLA as well ? You merged a commit into the branch, so you are a co-author now :) Screenshot 2022-06-07 at 1 41 02 PM

Well, that was embarrassing :D

PR #596, opened 2022-06-14
do not return that a strict analyzer is present in an application if …

kcboyle:
…a strict analyzer is excluded

If we have an application with a preflight like the following 

<details>
  <summary>Preflight Spec</summary>

```
kind: Preflight
metadata:
  name: supported-mysql-version
spec:
  collectors:
    - mysql:
        exclude: true
        collectorName: mysql
        uri: '{{repl ConfigOption "db_username"}}:{{repl ConfigOption "db_password"}}@tcp({{repl "db_hostname"}}:{{repl ConfigOption "db_port"}})/{{repl ConfigOption "db_name"}}'
  analyzers:
    - mysql:
        exclude: true
        strict: true
        checkName: Must be MySQL 8.x or later
        collectorName: mysql
        outcomes:
          - fail:
              when: connected == false
              message: Cannot connect to the MariaDB server
          - pass:
              message: The MariaDB server is ready
```
</details>

We are unable to deploy the application because the check still thinks that there is a strict analyzer, even though it is excluded. This PR adds a simple check to not include the analyzer if it is excluded. 

Screenshots of the behavior during deploy vv
Screen Shot 2022-06-14 at 11 18 43 AM
Screen Shot 2022-05-30 at 1 22 12 PM

### DIFF
--- a/pkg/preflight/util.go
+++ b/pkg/preflight/util.go
@@ -64,6 +64,9 @@ func hasStrictAnalyzer(analyzerMap map[string]interface{}) (bool, error) {
 		if err != nil {
 			return false, errors.Wrap(err, "error while un-marshalling marshalledAnalyzers")
 		}
+		if analyzeMeta.Exclude.BoolOrDefaultFalse() {
+			continue
+		}
 		if analyzeMeta.Strict.BoolOrDefaultFalse() {
 			return true, nil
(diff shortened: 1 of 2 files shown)


diamonwiggins (approved):
lgtm

PR #598, opened 2022-06-15
feat: adds new yamlCompare and jsonCompare analyzers

cbodonnell:
This PR adds two new types of analyzers `yamlCompare` and `jsonCompare` that can be used to compare an expected value against collected data.  The `when` condition functions similarly to the [text analyzer](https://troubleshoot.sh/docs/analyze/regex/) and takes a boolean input to determine whether the outcome has been satisfied (if `when` is not included, the default behavior is to pass when there is a match).  Additionally, there is a `path` field in both that allows the user to drill down to a particular field in the collected data structure before performing the comparison.

This PR also adds an `annotations` field to the analyzer metadata so that implementations can utilize them for various purposes (more on annotations [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)).

Example yamlCompare spec:
```
kind: Preflight
metadata:
  name: example-yaml-compare
spec:
  collectors:
    - data:
        name: example.yaml
        data: |
          foo: bar
          stuff:
            foo: bar
            bar: foo
          morestuff:
          - foo:
              bar: baz
  analyzers:
    - yamlCompare:
        checkName: Compare some yaml
        fileName: example.yaml
        path: "morestuff.[0].foo"
        value: |
          bar: baz
        outcomes:
          - warn:
              when: "false"
              message: The collected data does not match the value
              title: Not a match
              uri: https://troubleshoot.sh
          - pass:
              when: "true"
              message: The collected data matches the value
              title: Match
              uri: https://troubleshoot.sh
```

Example jsonCompare spec:
```
kind: Preflight
metadata:
  name: example-json-compare
spec:
  collectors:
    - data:
        name: example.json
        data: |
          {
            "foo": "bar",
            "stuff": {
              "foo": "bar",
              "bar": true
            },
            "morestuff": [
              {
                "foo": {
                  "bar": 123
                }
              }
            ]
          }
  analyzers:
    - jsonCompare:
        checkName: Compare some json
        fileName: example.json
        path: "morestuff.[0].foo.bar"
        value: |
          123
        outcomes:
          - warn:
              when: "false"
              message: The collected data does not match the value
              title: Not a match
              uri: https://troubleshoot.sh
          - pass:
              when: "true"
              message: The collected data matches the value
              title: Match
              uri: https://troubleshoot.sh
```


divolgin:
Could we utilize `when` on outcomes to make it also possible to fail when values match?  Like this, but less ugly

```
        outcomes:
          - error:
               when: < 5
          - warn:
               when: < 10
          - pass:
               ## default
```

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #611, opened 2022-06-28
Proposal provide extendable API for bundle information

chris-sanders:
I've talked this through with @emosbaugh a bit and want to make some changes and notes based on his thoughts. I'm opening this PR as a draft to keep the discussion open and start a history of the design discussion.


chris-sanders:
After discussion today with @emosbaugh there are a few changes we discussed and some discovery to do.

Summary:
* Neither of us are clear how much work using the aggregate API is but this could increase the level of effort to write a collector. (Not a collector spec which would presumably be the same once a collector is done)
* It's also possible this would reduce the burden of writing an analyzer b/c of the documented API to access data for analysis. Depending on the frequency with which Collectors and Analyzers are written this could offset the extra cost of writing a collector.
* There are likely many fields that we can't "push" into an API server as originally proposed. One very important example is pod status which is a field the API server would likely reject.
  * Because of this it might be more desirable rather than starting an api server for collecting to instead just start an etcd instance and push the key/values direcly into etcd. This etcd would still be used as a backend for an api-server when consuming data.
  * It's interesting to note this is very similar to what Velero is doing today, it's worth evaluating if Velero could be used as a collector if it were modified to write data to etcd rather than an object store. It's possible a "Velero Collector" could handle all native API server data using it's existing API to specify what to collect.
* Using the aggregation layer seems like a pretty big win for HostCollectors. An additional use case that came up was that by doing this you are very close to making any 'HostCollector' an installable feature on clusters so you could poll the collector on a live cluster anytime you wanted to check host stats.

I'll make some modifications to the design proposal to reflect the above from this conversation.

emosbaugh:
>   * It's interesting to note this is very similar to what Velero is doing today, it's worth evaluating if Velero could be used as a collector if it were modified to write data to etcd rather than an object store. It's possible a "Velero Collector" could handle 

it looks like velero uses api server to read and write data to etcd. it clears metadata and status on restore which makes sense but is unfortunate
https://github.com/vmware-tanzu/velero/blob/fb029fdd47af2042ee3dbfbd435aa24a04f6816e/pkg/restore/restore.go#L1128-L1129

Another idea. I wonder if its possible to leverage controller-runtime [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) which spins up a local instance of etcd and the kube-apiserver.

chris-sanders:
> it looks like velero uses api server to read and write data to etcd

I might be splitting hairs here, but it doesn't read/write to etcd but reads from the api-server, [stores in a file](https://github.com/vmware-tanzu/velero/blob/fad2df8740dc50e68dcf6b4a0d9a0a597f916abd/pkg/backup/backup.go#L345-L362), and writes back to the api-server after cleaning/modifying the prevoius json. Right?

The distinction I guess I'm making is that etcd is never directly talked to with Velero. Which means it can't set fields like Status that we are assuming the API server doesn't allow you to push to it. That might still make Velero an interesting collector, b/c it has a defined API to generically collect anything, but we would still need address pushing directly to etcd to get around any restrictions on what we can push, right?

> Another idea. I wonder if its possible to leverage controller-runtime [envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/envtest) which spins up a local instance of etcd and the kube-apiserver.

I agree this looks like a good starting point, probably easier and less maintenance for us than trying to do our own single binary compilation of the components. I was refreshing my memory on how [microk8s does it](https://github.com/canonical/microk8s/blob/master/build-scripts/patches/0000-Kubelite-integration.patch) and was looking to see how k3s does it and `envtest` looks cleaner than either other option at first glance.

I feel like a small experiment to answer:
* How easy is it to start an api-server backed by etcd with `envtest`.
* Can you in fact push metadata like "Status" into an api-server or do we have to write directly to etcd?
  * If we can't push to the api-server is just writing the information directly into etcd straight forward?

chris-sanders:
There are several options above for working with api-servers. I ran across the [k0s](https://docs.k0sproject.io/v1.23.8+k0s.0/architecture/) project which I notice is shipping the upstream binaries statically compiled with no dependencies. This lets it run a full control plane with "no" OS level dependencies and the binary is ~176MB all in. Considering we would need even less of the upstream than what they are including it might be a worthwhile option to consider doing something similar with troubleshoot.

emosbaugh (approved):
Any concerns I have have been left in previous comments or irl discussion. Lgtm

Issue #599, opened 2022-06-16
`postgres` analyzer fails to display `message:` field when the postgres pod can't be reached


adamancini:
**Bug Description**

in a preflight, the `postgres` analyzer doesn't include `outcome.Fail.Message` if the analyzer fails to connections

https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/postgres.go#L42-L55

**Expected Behavior**

my expectation for `isConnected` is that for any reason why the analyzer fails, I still get the message printed - for example on a first installation, if postgres isn't already available, the analyzer only dumps the connection failure message which may be confusing - I would expect to be able to catch the connection error message and also print `outcome.Fail.Message`

<img width="1126" alt="Screen Shot 2022-06-16 at 16 11 23" src="https://user-images.githubusercontent.com/292598/174156664-de2d841b-a450-4d72-a048-3392090178be.png">

**Steps To Reproduce**

Given:

```yaml
# preflight.yaml
    - postgres:
        checkName: Future versions of Sentry Enterprise will require PostgreSQL version 13 or greater.
        collectorName: postgresql
        outcomes:
          - fail:
              when: connected == false
              message: Cannot connect to PostgreSQL server. If you are installing for the first time this is normal.  If this is an upgrade, check that the pod is running and responding to DNS.
          - warn:
              when: version < 13.x
              message: >-
                {{repl if ConfigOptionEquals "postgres_type" "external_postgres_inline"}}
                  To prepare for the coming release of Sentry Enterprise 2.0, Please upgrade your PostgreSQL instance ASAP.
                {{repl else}}
                  This release uses a new embedded postgres 13 instance. Your data will be migrated to it automatically. Please allow several minutes after deployment for the migration to complete.
                {{repl end}}
          - pass:
              message: Your Postgres server is already compliant.
```

on a first installation, the analyzer is expected to fail, and I would cover that with messaging

adamancini:
fixed by #612

PR #612, opened 2022-06-28
include the db connection error message with the outcome

adamancini:
make sure to return `outcome.Message` even if the database connection fails, and include the connection error message with the outcome


adamancini:
fixes #612

adamancini:
fixes #599

diamonwiggins (dismissed):

diamonwiggins (approved):

PR #614, opened 2022-07-01
Redact Host Collectors in Support Bundles

diamonwiggins:


pkg/supportbundle/collect.go
```diff
@@ -20,7 +20,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func runHostCollectors(opts SupportBundleCreateOpts, hostCollectors []*troubleshootv1beta2.HostCollect, bundlePath string) (collect.CollectorResult, error) {
+func runHostCollectors(additionalRedactors *troubleshootv1beta2.Redactor, opts SupportBundleCreateOpts, hostCollectors []*troubleshootv1beta2.HostCollect, bundlePath string) (collect.CollectorResult, error) {
```
emosbaugh: can we make the order of params match this function? 

```
func runCollectors(collectors []*troubleshootv1beta2.Collect, additionalRedactors *troubleshootv1beta2.Redactor, bundlePath string, opts SupportBundleCreateOpts) (collect.CollectorResult, error) {
```

https://github.com/replicatedhq/troubleshoot/pull/614/files#diff-ef0fda343f4b7fa25eae0a9644502e8403b266b231d844b85954d8fcba7d9a24R74

emosbaugh (changes requested):

pkg/supportbundle/collect.go
diamonwiggins: Ah, yup. Meant to go back and do that but forgot. Will do.

emosbaugh (approved):

PR #621, opened 2022-07-15
fix: use kube-aware version selector for crs collector

mhrabovcin:
Fix the way of selecting the API version of CRDs when a CRD contains multiple versions. This will ensure that the "latest" version is stored in the bundle.

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"path" // this code uses 'path' and not 'path/filepath' because we don't want backslashes on windows
 	"path/filepath"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/collect/cluster_resources.go
```diff
@@ -795,10 +809,15 @@ func crsV1(ctx context.Context, client dynamic.Interface, config *rest.Config, n
...
-			version = crd.Spec.Versions[0].Name
-		}
-		if len(crd.Status.StoredVersions) > 0 {
-			version = crd.Status.StoredVersions[0]
+			versions := []string{}
+			for _, v := range crd.Spec.Versions {
+				versions = append(versions, v.Name)
+			}
+
+			version = versions[0]
```
divolgin, line 817: This can just be `version = selectCRDVersionByPriority(versions)`.   There is no need for the check below.  But ok.

divolgin (approved):

PR #626, opened 2022-07-19
add instructions to cross-compile with Go

adamancini:


xavpaice (approved):
Looks good, tested the updated process and ran the binary on Centos to confirm.

PR #643, opened 2022-08-02
Fix the saving log process

YoyoTT:
With this change, log error message could be saved to disk.


YoyoTT:
Could anyone please review this pr and feel free to leave comments! Thanks!

divolgin:
@YoyoTT thanks for the submission.  Can you give an example of what this fixes?  Looking at that code block, my first thought is that we should remove the nested `if` statement and ignore the error from saving.

YoyoTT:
> @YoyoTT thanks for the submission. Can you give an example of what this fixes? Looking at that code block, my first thought is that we should remove the nested `if` statement and ignore the error from saving.
 
Yeah, I think both of us are trying to fix the same error.  Without this change, it returns and this loop is done so that that podLog isn't save into `output` map and we could not get either log or error message.

divolgin:
OK, that's what I thought.  I'll merge it.

divolgin (approved):

PR #644, opened 2022-08-03
Allow custom messages to be passed if a deployment/statefulset is absent

### LINKED ISSUE #657: Add analyzer that checks for deployment name
## Problem to solve (Required)
- Warn users when ECKO is not installed. related issue: https://app.shortcut.com/replicated/story/47998/analyzer-spec-warn-when-ekco-is-not-installed-in-a-kurl-cluster

## Proposal

- The current [deployment status](https://troubleshoot.sh/docs/analyze/deployment-status/) configures outcomes based on replicaSet count. 
  - When an deployment doesn't exist, then it shows an error.
  - The error is not a configurable outcome. A warning is preferred to resolve the related issue: https://app.shortcut.com/replicated/story/47998/analyzer-spec-warn-when-ekco-is-not-installed-in-a-kurl-cluster
- Ideas
  - Add a new analyzer that checks if a resource exists by specific attributes like names, labels, or annotations 
  - Modify deployment status analyzer to allow handling "doesExist" condition 

## Further details

- Deployment status analyzer [source code](https://github.com/replicatedhq/troubleshoot/blob/af1afd09ca79a45b5d826f3afd7dc3b753da1ac8/pkg/analyze/deployment_status.go#L48)
  - links to the failure condition that's triggered when a deployment doesn't exist 

## Impact (importance/urgency) (Required)

* Replicated impact: Reduced support overhead 
* Vendor impact: Faster issue resolution 
* End-user impact: Faster issue resolution 

## What does business success look like, and how can we measure that?

- 50% reduction in issues caused by lack of ECKO operator 

## Links / references

- Slack [discussion](https://replicated.slack.com/archives/C0PHDRHJ9/p16535859031497390)

danj-replicated:
This is an attempt to address internal shortcut 49195. 

We would like to be able to add custom messages when a deployment (e.g EKCO) is absent from the cluster to better explain how to resolve it's absence.

This change introduces a new special keyword "absent" that can be used in the "When" statement in a deploymentStatus or statefulSetStatus analyzer to allow absent deployments and statefulsets to be handled with custom messages.

for example:
```yaml
kind: SupportBundle
metadata:
  name: example
spec:
  analyzers:
    - deploymentStatus:
        name: ekc-operator
        namespace: kurl
        outcomes:
          - fail:
              when: "absent"
              message: "The ekc-operator deployment is not instaled. Without it, this cluster may not behave as intended"
```

While the change acheives this objective, I do have some reservations that I'd like to discuss:

- currently the `when: "absent"` statement would need to be the first outcome for an analyzer due to how I'm catching it. comments on how I can improve this would be greatly appreciated.
- I'm currently having to repeat a block to allow for the default `the <resource> <name> does not exist` for each warn|pass|fail outcome when we aren't specifically looking for the absence of a deployment. I'd like to reduce this repitition if possible.

### DIFF
--- a/pkg/analyze/common_status.go
+++ b/pkg/analyze/common_status.go
@@ -59,7 +59,7 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, name string, iconKey 
 
 			if exists == false  && outcome.Warn.When != "absent" {
 				result.IsFail = true
-				result.Message = fmt.Sprintf("The %s %s was not found", resourceType, name)
+				result.Message = fmt.Sprintf("The %s %q was not found", resourceType, name)
 				result.URI = outcome.Fail.URI
 				return result, nil
 			}
@@ -95,7 +95,7 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, name string, iconKey 
 
 			if exists == false  && outcome.Pass.When != "absent" {
 				result.IsFail = true
-				result.Message = fmt.Sprintf("The %s %s was not found", resourceType, name)
+				result.Message = fmt.Sprintf("The %s %q was not found", resourceType, name)
 				result.URI = outcome.Fail.URI
 				return result, nil
 			}
--- a/pkg/analyze/common_status.go
+++ b/pkg/analyze/common_status.go
@@ -21,7 +21,7 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, name string, iconKey 
 		if outcome.Fail != nil {
 
 			// if we're not checking that something is absent but it is, we should throw a default but meaningful error.
-			if exists == false  && outcome.Fail.When != "absent" {
+			if exists == false && outcome.Fail.When != "absent" {
 				result.IsFail = true
 				result.Message = fmt.Sprintf("The %s %q was not found", resourceType, name)
 				result.URI = outcome.Fail.URI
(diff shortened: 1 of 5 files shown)


edgarlanting (commented):
@danj-replicated Based on my understanding of the original shortcut and when we discussed this, it's what we're looking for. I would say that keeping it like this will make sure its better to adjust where needed. Personally I think its a good approach and I'm curious as to what @diamonwiggins thinks of the change.

pkg/analyze/common_status.go
```diff
@@ -39,6 +56,14 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, title, iconKey string
 				return result, nil
 			}
 		} else if outcome.Warn != nil {
+
+			if exists == false  && outcome.Warn.When != "absent" {
+				result.IsFail = true
+				result.Message = fmt.Sprintf("The %s %s was not found", resourceType, name)
```
divolgin: I would keep %q for name here

divolgin:
Overall approach looks ok to me.  Can we have this for all other related resources: jobs, replicate sets, and stateful sets?

Also this needs tests.

danj-replicated:
I don't think it would be too hard to add this to the other resources

edgarlanting (approved):
LGTM @danj-replicated!

divolgin (approved):

PR #647, opened 2022-08-05
Update cluster_resources.go - Shortcut 44743

edgarlanting:
When troubleshooting an issue we want to see and check the `PodDisruptionBudget` for any application. As we currently do not have that 'standard' output in our support bundle, this PR serves as an implementation of providing one with an overview of `PodDisruptionBudget` on a per-namespace basis.

After creating a support bundle, the output can be reviewed by checking out `../cluster-resources/pdb/*.json`

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -99,12 +99,14 @@ func ClusterResources(c *Collector, clusterResourcesCollector *troubleshootv1bet
 		}
 	}
 
-	// pod disruption budgets
-	pdb, pdbError := getPodDisruptionBudgets(ctx, client, namespaceNames)
-	for k, v := range pdb {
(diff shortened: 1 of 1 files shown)


danj-replicated (commented):
This may just be me being picky, but I think the directory structure would look a bit neater using the long resource names instead of the abbreviated version. 

e.g: `PodDisruptionBudgets` or `pod-disruption-budgets` over `pdb`

edgarlanting:
> 

You've got a point there @danj-replicated as this will improve readability instead of having a nice abbreviated one.

edgarlanting (commented):
Made small readability changes by moving `pdb` to `PodDisruptionBudgets`

danj-replicated (approved):

PR #648, opened 2022-08-05
Add a check for longhorn volume replicas

### LINKED ISSUE #655: Warn when Longhorn or Ceph isn't configured with multiple replicas on HA clusters
### Problem to solve
We should warn when Longhorn or Ceph isn't configured properly in HA environments. We can likely safely make an assumption that a customer wouldn't want a single point of failure with the persistent storage provider if they have more than one node

### Definition of done:

* Includes a test to verify the collector and analyzer work
* The analyzer provides an explanation on how to resolve the issue for the user, linking to any additional information that might be needed to fix the issue.

danj-replicated:
This is to address part of internal shortcut 48952, aka #655 


edgarlanting (approved):
LGTM!

PR #649, opened 2022-08-09
Collect networkPolicies

danj-replicated:
This addresses internal shortcut 53924, and gives us some visibility to networkPolicies deployed in clusters.

The usefullness of this may be improved upon further with the addition of some kind of ip address mapping rather than opaque obfuscation.


diamonwiggins (approved):
lgtm

xavpaice (approved):

PR #670, opened 2022-08-17
Refactor in-clusters collectors to use struct per collector

### LINKED ISSUE #680: Switch Collector from struct to interface
## Outline 

With #665 we have Troubleshoot appending multiple specs provided on the CLI (or call) into one long spec.  Some collectors which may be specified more than once might have specific de-duplication rules which need to be taken into account, to avoid collecting the same thing multiple times.

This task is to enable a means to run a collector specific merge which could de-duplicate, or aggregate, based on the collector type.

In scope:
* one generic Collector type

Out of scope:
* Analyzer and Redactor de-duplication.

## Proposal

* change the `Collector` type to an interface with methods that include `Collect()`
* write a new generic type for collectors to use, that implements the interface
* modify each collector to use the generic type, so that it can be called with `collector.Collect()` to actually get the data
* Implement at least 1 merge to demonstrate how it works

## assumptions

* #665 provides a spec with all the specs provided to Troubleshoot into one `SupportBundle`, no other changes are needed to receive multiple inputs.

## Definition of done

- [ ] Interface implemented for Collector that includes a merge function with 1 collector implementing it
- [ ] Interface implemented for Collector that doesn't include merge functional with remaining collectors implementing it
- [ ] Documentation for how to make use of the interface for a new collector
- [ ] Test both interfaces to confirm merge and non-merge works as expected

## Followup tasks

* Similar implementation for Analyzers and Redactors

### LINKED ISSUE #728: Support Bundle progress UI gets mashed together with bundle name
**Bug Description**

Support Bundle progress UI gets mashed together with bundle name

Screen Shot 2022-09-20 at 4 15 09 PM

**Expected Behavior**

Clear progress UI or drop a newline before printing the bundle path. Or even "bundle generated at $PATH".

**Steps To Reproduce**

```
cat <<EOF > support-bundle.yaml
kind: SupportBundle
metadata:
  name: workshop
spec:
  collectors:
    - clusterInfo: {}
    - clusterResources:
        namespaces:
        - default
EOF

kubectl support-bundle support-bundle.yaml
```
**Additional Context**

0.42.0

- Operating system

MacOs

- Operating system version
 12.2.1

single node GKE cluster

diamonwiggins:


pkg/collect/collector.go
```diff
@@ -2,29 +2,31 @@ package collect
...
+type Collector interface {
+	Title() string
+	IsExcluded() (bool, error)
+	GetRBACErrors() []error
+	HasRBACErrors() bool
+	CheckRBAC(ctx context.Context, collector *troubleshootv1beta2.Collect) error
+	Collect(progressChan chan<- interface{}) (CollectorResult, error)
 }
 
-type Collectors []*Collector
```
xavpaice, line 27: not sure why we're taking this out?

diamonwiggins: It can be added back. I think i scrapped it not thinking about backwards compatibility as well as the fact that it wasn't being used much in the code.

pkg/collect/ceph.go
```diff
@@ -105,23 +106,65 @@ var CephCommands = []CephCommand{
...
+
+func (c *CollectCeph) HasRBACErrors() bool {
+	return len(c.RBACErrors) > 0
+}
+
+func (c *CollectCeph) CheckRBAC(ctx context.Context, collector *troubleshootv1beta2.Collect) error {
+	exclude, err := c.IsExcluded()
+	if err != nil || exclude != true {
+		return nil
+	}
```
emosbaugh: should this be exclude == true?

```suggestion
	exclude, err := c.IsExcluded()
	if err != nil || exclude {
		return nil
	}
```

pkg/collect/ceph.go
```diff
@@ -105,23 +106,65 @@ var CephCommands = []CephCommand{
...
+
+func (c *CollectCeph) Title() string {
+	return collectorTitleOrDefault(c.Collector.CollectorMeta, "Cluster Info")
+}
+
+func (c *CollectCeph) IsExcluded() (bool, error) {
+	return isExcluded(c.Collector.Exclude)
+}
+
+func (c *CollectCeph) GetRBACErrors() []error {
```
emosbaugh: im not sure im in love with my code here but one way to reduce a lot of duplicate code

https://go.dev/play/p/bNGSuEj4IAT

pkg/collect/collector.go
```diff
@@ -47,305 +49,59 @@ func isExcluded(excludeVal *multitype.BoolOrString) (bool, error) {
...
-			c.Collect.Sysctl.Namespace = namespace
-		}
-		result, err = Sysctl(ctx, c, client, c.Collect.Sysctl)
-	} else {
-		err = errors.New("no spec found to run")
-		return
-	}
-	if err != nil {
-		return
+	switch {
```
xavpaice, line 56: would we be wise at this stage to use this 'type detection based on populated data' type of pattern to build a slice of collectors, which we can then loop over using `collector.Collect(chan)`?  I'm thinking of a future where we put them in goroutines rather than collect syncronously.  Maybe a second PR though after this one?

emosbaugh (approved):

PR #671, opened 2022-08-19
Create initial README.md

edgarlanting:
This README will need to provide those working with the collector/analyzer CRD with all the necessary information to operate & maintain the CRD specs. This README should always reside in this `config/crds` location.


xavpaice:
I agree we need a README there, but I'd like it to have some detailed information in it rather than just merging an empty file for future use.

edgarlanting (commented):
@xavpaice I've made a short start documenting the Troubleshoot CRDs which we can also work together with the current project in #682.

config/crds/README.md
```diff
@@ -1 +1,16 @@
+**Managing Troubleshoot CRDs**
 
+The CRDs in this folder are used by Troubleshoot in combination with the spec that can be passed on to the Troubleshoot CLI in various ways (the location of the spec can be passed on as a web address, yaml file or a secret residing on a cluster).
+
+Should the CRDs (and corresponding schemas) need to be (re-) generated or if an additional field or type needs to be added, the following steps need to be taken;
+
+**Adding new field(s)/type(s):**
```
chris-sanders:
```suggestion
## Adding new field(s)/type(s)
```

config/crds/README.md
```diff
@@ -1 +1,16 @@
...
+
+New fields (types) need to be defined in the `./pkg/apis/troubleshoot/v1beta2/supportbundle_types.go` file.
+
+
+**(Re)generate new schema's/CRDs:**
```
chris-sanders:
```suggestion
## Regenerate schemas/CRDs
```

config/crds/README.md
```diff
@@ -1 +1,16 @@
...
+
+Generating new schemas in `./schemas` can be done by running `make schemas`, which will then implement the new type into the CRDs
+
+*TBD: More detailed documentation on the CRDs and corresponding components.*
```
chris-sanders: I tend to avoid TBD values like this just because the nature of being open source there's always opportunity for people to add more. If you're leaving this as a note before you finish the PR that's fine, otherwise I'd drop it.

config/crds/README.md
```diff
@@ -1 +1,16 @@
+**Managing Troubleshoot CRDs**
```
chris-sanders:
```suggestion
## Managing Troubleshoot CRDs
```

config/crds/README.md
```diff
@@ -1 +1,16 @@
...
 
+The CRDs in this folder are used by Troubleshoot in combination with the spec that can be passed on to the Troubleshoot CLI in various ways (the location of the spec can be passed on as a web address, yaml file or a secret residing on a cluster).
```
chris-sanders, line 3: You mention the spec here but never again. Is the spec relevant here, if not I'd not bother mentioning it, if so some clarification on why it's important would be good. I don't think how a spec is passed to the cli is particularly relevant to the CRDs.

chris-sanders (changes requested):
Mostly formatting to use Markdown syntax for section dividers rather than just bold text. A few comments to consider on conent.

config/crds/README.md
xavpaice: I think this statement is describing why there are CRDs and what they're used for  - in this case parsing specs for various Troubleshoot components.  If that's the case, this could be reworded to make the intent of the CRDs clearer and where they're used in the code (or, at least, what for).  I'd avoid talking about the CLI specifically unless the CRDs are _only_ relevant when Troubleshoot is called from the CLI, but I don't think that's the case?

config/crds/README.md
edgarlanting: Thanks @chris-sanders for the review! I'm dropping the TBD part.

Issue #650, opened 2022-08-09
[EPIC] Mergeable Spec


chris-sanders:
### Problem
Troubleshoot doesn't have a modular way for different components to specify specifications. If a software project wanted to include project specific items the end user would have a hard time knowing where to find those and how to collect them. Furthermore if multiple projects were to do this the user would have to run troubleshoot multiple times.

Ideally troubleshoot would allow merging of specs to allow building a spec either automatically or influenced by user input to target specific needs.

### Proposal
Add the ability for troubleshoot to merge specs and deal with combining overlapping configurations to build one master spec. This should allow any of the input specifications to build up a list of specs that are then merged for the final specification. The final specification should be included in the final bundle.

### Definition of done
* Ability for troubleshoot to merge specs
* Tests added to verify the merge works including edge cases where combining is not straight forward
* Documentation added explaining how merging works

### Started

- [x] https://github.com/replicatedhq/troubleshoot/issues/663
- [x] https://github.com/replicatedhq/troubleshoot/issues/665
- [x] https://github.com/replicatedhq/troubleshoot/issues/680
- [x] #699
- [x] Support multiple Redactor specs from the CLI

### Planned

- [x] #713
- [ ] de-deuplicate Redactor specs (Redactor is already an Interface in package `redact`)
- [ ] Switch `troubleshootv1beta2.Analyze`to Interface

chris-sanders:
Documenting some discussion from today for broader review.

It's unclear if "merging" really needs to mean anything more than deduplicating. Even deduplicating might be an optimization for reducing runtime that has little affect on the final outcome. As an example, if logs are collected twice for the same pod the logs will still be present but collection may incrementally increase because the API call is made twice to get the same information.

While ideally these duplicate issues would be reduced it may not be necessary in a first implementation to address explicitly for all collectors. Having said that, the current proposal is to consider adding the merge logic at the collector level. Ideally there would be a default "Merge" function that does nothing more than append all of the values into a single file with no merging or deduplication. Collectors can then each implement a collector specific merge to override the default if 'merging' has value for that particular collector. This also lets merge logic be scoped to specific collectors where the decision to append, merge, or deduplicate can be made knowing the context of what is being collected.

Some examples to help clarify this.
* Two runPod collectors with the same name and different commands are unmergable. Only one should be run and an error logged.
* Two configMap collectors can be deduplicated and merged depending on the configurations
* Two logs collectors likely only need deduplication to prevent collecting the same thing twice

[...]

PR #681, opened 2022-08-26
add proposal for updateable specs

xavpaice:


design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,90 @@
...
+* Decouple the provision of default specs from the kots application manager, so that kots upgrades are not required in order to get access to updated specs
+
+## Non Goals
+
+* Maintain the ability for Troubleshoot to run in an airgapped environment
+* maintain compatilbility with existing Troubleshoot specs
+
+## High-Level Design
+
+Add a new field to the Troubleshoot spec definition, which includes a URI used to locate the current spec online.
```
xavpaice: change to "includes a URI used to locate additional specs" rather than specify online, and rather than hinting that we replace the existing spec.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,90 @@
...
+
+If the field is populated, Troubleshoot is to attempt to download the additional spec(s) from the source online, and add to the spec already provided.
```
xavpaice: remove online, could be a secret or a file

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,90 @@
...
+
+Add a CLI flag to prevent any attmept to download, for use in airgap environments and/or when we simply do not want to download.
```
xavpaice: Discussion point - do we have a CLI flag here that prevents us needing to go online, or do we disable this specURI feature entirely?  If it's only there for airgap, the former, if it's there to avoid grabbing extra sources, the latter.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,90 @@
...
+
+We could add a new type, `specURIs`, which contains a list of URIs from which to obtain specs:
+
+```yaml
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: default
+spec:
+  specURIs:
```
xavpaice: specURIs, additionalURIs, all sorts of names could be used.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,90 @@
...
+    - clusterInfo: {}
+    - clusterResources: {}
+  analyzers:
+    - cephStatus: {}
+    - longhorn: {}
+```
+
+## Impact on kURL and kots
+
+For kURL, each component/addon installed could include adding a URI pointing at a specific spec that
```
xavpaice: oops.

For kURL, each component/addon installed could include adding a URI pointing at a specific spec for that component, which could be updated independently of other components.

design/proposal-updateable-specs.md
chris-sanders: It'sa small thing, I think singular is more appropriate here? The spcURI (or specUri) is good for the whole spec right? If you need multiple make multiple specs and use the merging?

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+## Background
+
+Troubleshoot is limited in that when folks write an update or a new spec for data collection and analysis, that upgrade is not available when Troubleshoot is invoked from within the kots application manager, or from specs stored in a file.  Updating the local file is simple enough, but updating the specs supplied by kots requires an upgrade to kots in order to collect updates to the default spec supplied by kots, and/or an update to the application which is deployed if the spec supplied by the application is changed.  This update requirement delays folks from accessing updated Troubleshoot specs and therefore discourages people from writing new/updated specs.
+
+It is also difficult to justify the tight coupling of Troubleshoot configuration, which is intended to be an entirely client-side concept, and the application manager which is installed to a Kubernetes environment.
+
+## Goals
+
+* Provide the ability to automatically obtain a spec from a source that receives updates, whenever Troubleshoot is run
+* Decouple the provision of default specs from the kots application manager, so that kots upgrades are not required in order to get access to updated specs
```
chris-sanders: I don't think decoupling KOTS is a goal unless you actually intend to modify the KOTS code. Instead I would suggest the goal is to allow Specs inputs which are not URL based to still receive an update from a source URL. In the case of KOTS it uses secrets so then a KOTS spec could receive updates by simply including the new URI field.

While I would think decoupling the default spec would be good, simply adding a URI to the existing spec would (I believe) solve our primary issue. Actually decoupling is more of a KOTS decision that I don't think is linked to the scope of this change.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Non Goals
+
+* Maintain the ability for Troubleshoot to run in an airgapped environment
```
chris-sanders: I disagree with this one, Troubleshoot should still be able to run in an airgapped environment. That doesn't mean it has to have the same features as it would with internet access. Some acceptable solutions I would see are.

* Use of a --no-updates flag to simply disable all checks for updates
* Failing forward if no update can be found continue with the spec at hand

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
+# Provide a facility to automatically obtain updated specs
+
+## Background
+
+Troubleshoot is limited in that when folks write an update or a new spec for data collection and analysis, that upgrade is not available when Troubleshoot is invoked from within the kots application manager, or from specs stored in a file.  Updating the local file is simple enough, but updating the specs supplied by kots requires an upgrade to kots in order to collect updates to the default spec supplied by kots, and/or an update to the application which is deployed if the spec supplied by the application is changed.  This update requirement delays folks from accessing updated Troubleshoot specs and therefore discourages people from writing new/updated specs.
+
+It is also difficult to justify the tight coupling of Troubleshoot configuration, which is intended to be an entirely client-side concept, and the application manager which is installed to a Kubernetes environment.
```
chris-sanders: Expanding on this below I feel like this is out of scope.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
+# Provide a facility to automatically obtain updated specs
+
+## Background
+
+Troubleshoot is limited in that when folks write an update or a new spec for data collection and analysis, that upgrade is not available when Troubleshoot is invoked from within the kots application manager, or from specs stored in a file.  Updating the local file is simple enough, but updating the specs supplied by kots requires an upgrade to kots in order to collect updates to the default spec supplied by kots, and/or an update to the application which is deployed if the spec supplied by the application is changed.  This update requirement delays folks from accessing updated Troubleshoot specs and therefore discourages people from writing new/updated specs.
```
chris-sanders: This explanation requires fairly deep understanding of KOTS and how it uses Troubleshoot. Could we capture the same sentiment w/o making the assumption users have that background?

Simply referencing secrets as specs that are installed as part of a release (KOTS, Helm, whatever) would make the same point and require less KOTS specific background wouldn't it?

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Goals
+
+* Provide the ability to automatically obtain a spec from a source that receives updates, whenever Troubleshoot is run
+* Decouple the provision of default specs from the kots application manager, so that kots upgrades are not required in order to get access to updated specs
+
+## Non Goals
+
+* Maintain the ability for Troubleshoot to run in an airgapped environment
+* maintain compatilbility with existing Troubleshoot specs
```
chris-sanders: I'm curious about this one, it would seem to me that ideally we would maintain compatibility. I'm no expert here, but I was under the impression as long as the default value (empty string?) was defined we could add this and not break old clients?

If at all possible I wouldn't want Troubleshoot to refuse to process a new Spec. I would like  to be able to start using the new format and updates immediately w/o having to maintain a version for people that haven't yet upgraded.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+## High-Level Design
+
+Add a new field to the Troubleshoot spec definition, which includes a URI used to locate additional specs.
+
+If the field is populated, Troubleshoot is to attempt to collect the additional spec(s) from the location provided, and add to the spec already provided.
+
+If the additional spec is not found at the location (or, if there is no network access to that location), Troubleshoot is to continue processing the remainder of the spec provided, with a log message describing the failure.
```
chris-sanders: I was expecting this to not be 'additional' specs but a replacement for the current spec. IE if you include a URI and Troubleshoot finds a spec at that URI it supersedes the current spec.

We should consider which we actually want, by superseding the new spec has full control over what's run which is what I think I prefer. An example is, if you ship a spec and then later realize the recommendation it provides is inaccurate, you want to be able to stop that spec from saying that not just add a new analyzer which provides a different message.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+Add a CLI flag to disable the URI specified locations from being accessed.  This would be useful in airgap environments and/or when we simply do not want to have additional specs used, without having to update the yaml, secret or remote source.
```
chris-sanders:
```suggestion
Add a CLI flag to disable the URI specified locations from being accessed.  This would be useful in airgap environments and/or when we simply do not want to have additional specs used, without having to update the yaml, secret, or remote source.
```

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+Update the [default spec included with kots](https://github.com/replicatedhq/kots/blob/main/pkg/supportbundle/defaultspec/spec.yaml) to include a URI pointing at [the troubleshoot-specs repo](https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/in-cluster/default.yaml).
```
chris-sanders: I see KOTS changes have been intentional. I think we should scope the change for Troubleshoot only, we could still propose a change to KOTS but I'd prefer we do a document and concerted effort specifically on that separately.

There might be reasons why KOTS itself doesn't use this feature but it can still be available for use until KOTS implements it. If this change unintentionally impacts KOTS usage of Troubleshoot we should consider how to avoid that or require a major version of Troubleshoot (which I would hope isn't necessary).

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+URI's provided may be any of the URI types already accepted by Troubleshoot, including web addresses (https://), secrets, or files.
```
chris-sanders, line 26: That's cool, didn't think about it but sure why not.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Impact on kURL and kots
+
+For kURL, each component/addon installed could include adding a URI pointing at a specific spec for that component, which could be updated independently of other components.
+
+## Limitations
+
+Redactors are currently excluded from this proposal, though it would be desirable to add a similar feature in the future.
+
+If a URI provided includes a spec which includes another URI, it is possible to get into some kind of recursion.  We should count the number of URI calls and limit that.
```
chris-sanders: I think this is a non-issue if you treat this as an update/replacement only and not additional inputs. If for example, the phases of collecting are to retrieve each input, check and replace the input if a URI is found, and then process the spec there is no recursion because you don't re-check what is considered the canonical version found in the URI.

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Assumptions
+
+Troubleshooot is able to accept multiple specs on invocation.
```
chris-sanders: I don't think this is necessary for this feature. It might be true based on current development, but is there any reason you can't update a single provided input?

design/proposal-updateable-specs.md
```diff
@@ -0,0 +1,92 @@
...
+## Testing
+
+## Alternatives Considered
+
+* Some mechanism to trigger a process to update secrets stored in Kubernetes.  This was rejected due to the need to run a manual update spec for something installed in the cluster.
+* Altering kots to provide a URL for the spec rather than storing it in code.  This was rejected to allow airgap compatibility.
+
+## Security Considerations
+
+Troubleshoot currently allows downloading from URL, and that pattern is frequently used from the CLI when customers are directed to do so.  If we made this change, customers would need to ensure that their firewall allows outbound connections to the URL target from the kots container running Troubleshoot.
```
chris-sanders: You are speaking to KOTS changes here. I think I would approach this differently, to include all calls of Troubleshoot including the CLI. Something like the following address not just KOTS but the CLI and does so without implying a KOTS change:

"This change will make additional network calls with default usage of Troubleshoot. To ensure a user has control of limiting these calls an argument to disable this feature entirely will be provided. This allows users with different security postures to select what's appropriate for their environment."

chris-sanders (commented):
Only after leaving comments am I realizing there's a larger fundamental difference to how we were seeing this feature that we should discuss which would likely cover most of the in line comments.

You have written this to include the ability to obtain an arbitrary number of _additional_ specs as a new field in a spec. Then you've scoped this to implement the change in how KOTS uses Troubleshoot to split out the defaults into a location that can be updated.

I was approaching this as providing a single canonical URI that is the _truth_ for the given spec. There's only one and it's always assumed to be the best version of that spec if it's available. Furthermore, I would not include changes to KOTS or any client of Troubleshoot but implement the feature as a new option for CLI and embedded to have access to. This would mean ensuring the change doesn't brake clients (like KOTS) that already have it embedded if they see a new spec.

chris-sanders (changes requested):
Just marking this as "Request Changes" to indicate we need to talk through this more to close some of the gap in our approaches/expectations.

design/proposal-updateable-specs.md
xavpaice: If Troubleshoot doesn't recognize the new field, it's possible that could cause issues.  Just tested and the current release of Troubleshoot survives having the new field included in a bundle without issues (it just does nothing with the info).

design/proposal-updateable-specs.md
xavpaice: I was thinking we'd have a list, and merge them all together along with the spec provided originally - that way we can have a master spec with a list of remotes, and/or an update URI.  If we agree to replace the spec entirely then we need to determine if the download of the URI is successful, and what to do if not.  If we have multiple that's a problem (what if we get 2/3 of them?), so I think we go to just one URI per spec, and have that entirely replace if downloaded.

design/proposal-updateable-specs.md
xavpaice: problem is if the spec in the URI includes another URI, and is processed in the same way - so a URI points to a spec with a URI, which points to another spec with another URI, and we go round in circles.  If we ensure that only the URI provided in the spec parsed at Troubleshoot init is processed then this is not an issue.

design/proposal-updateable-specs.md
xavpaice: agreed

design/proposal-updateable-specs.md
xavpaice: I have left (and rewritten) a section for KOTS and kURL changes that might be useful, but been specific that they're excluded from this proposal.  I think it's important that we explain how/why the KOTS project would adopt these features, and what to consider when they do.

xavpaice:
Thanks for reviewing this so clearly.  I've made some updates based on the feedback and with some further thought, which I think resolves a bunch of what we were talking about, would be good to discuss further to ensure we're on the same page.

In particular, I think that until we have mergeable specs released, the KOTS project isn't going to be able to use the update-able spec feature - however it is still of great value alone.

design/proposal-updateable-specs.md
chris-sanders: I'm not clear what distinction you're making between not going online and not using the URI for updates. Those feel like the same outcome can you clarify a bit what the nuance is between those two options?

design/proposal-updateable-specs.md
chris-sanders: I think for updating specifically singular has less edge cases and is more predictable what's going to happen. However, once you're dealing with mergable specs I can how it would be fairly clean to have a spec that just references other specs. That sounds like a good feature, although maybe more related to #650 as an additional input field? Something like "additionalSpecs" that takes a list, then people could give that one input to Troubleshoot and it would be the same as giving multiple arguments.

Back to the name, if we're sticking with a single canonical spec for this field. I think the name should drop to just `Uri` or `URI` (I'm unclear on fully capital vs camelcase). The reason I propose that is because the full path to the field as written would be `spec.specURI` which seems like unnecessary repetition.

design/proposal-updateable-specs.md
xavpaice: URI is a path - could be an https:// address, or a file, or a secret.  Some are online, some not.

design/proposal-updateable-specs.md
xavpaice: agreed

chris-sanders (approved):
This looks good we can work out any more details in implementation.

PR #683, opened 2022-08-26
Danj multiple inputs

### LINKED ISSUE #665: Make Troubleshoot accept multiple specs
Per the design in https://github.com/replicatedhq/troubleshoot/pull/662, make the code changes to Troubleshoot to support multiple specs during a single run.

Things to consider when working on this task:

* define a new interface that provides the merge functionality for collectors, analyzers, and redactors
* there should be a generic implementation of that interface which is used by default for all objects.  This can simply use `append()`.
* specific collectors may have alternative implementations of the interface where overrides are required and should have a way to have the collector specific function called instead.

Particular overrides known at this point in time (for example only not to be implemented now):

* Longhorn and Ceph have a namespace config, if multiple collectors are specified with different namespaces then each should be collected
* clusterResources has an option namespaces config, these should be merged and deduplicated
* Two runPod collectors with the same name and different commands are unmergable. Only one should be run and an error logged.
* Two configMap collectors can be deduplicated and merged depending on the configurations
* Two logs collectors likely only need deduplication to prevent collecting the same thing twice

This particular task:

* Make the support-bundle CLI accept multiple specs for Collectors and Analyzers

Definition of done:

- [x] CLI to accept multiple spec inputs for Collectors and Analyzers (excluded: Redactors)
- [x] Troubleshoot to append each spec provided to make one, aggregated spec
- [x] changes include tests
- [x] Compatibility with kots maintained without modification to kots
- [x] Documentation for how to add deduplication for other collector types

danj-replicated:
This PR attempts to address issue https://github.com/replicatedhq/troubleshoot/issues/665 

This is by no means a complete implimentation, and relies on https://github.com/replicatedhq/troubleshoot/issues/680 to enable collectors to be able to perform "intelligent merging".

but for now it allows multiple specs to be passed to `support-bundle` and will aggregate all defined items in the support-bundle spec, with some limited deduplication for cluster-info and cluster-resources collectors.


pkg/apis/troubleshoot/v1beta2/supportbundle_interfaces.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (s *SupportBundleSpec) MergeCollectors(spec *SupportBundleSpec) {
+    for _,c := range spec.Collectors {
+        // we want to move away from checking for specific collectors in favor of allowing collectors to expose their own merge method
+        // ideally we'd just want to be able to call something like c.Merge(spec) here and have the collector's own merge method work out what to do'
+
+        if c.ClusterInfo != nil {
+            // we only actually want one of these so skip if there's already one
+            y := 0
```
divolgin: this should be a boolean with a meaningful name, like `hasClusterInfo := false`

divolgin (changes requested):
This needs formatting (go fmt) and dead code removed.

cmd/troubleshoot/cli/run.go
```diff
@@ -68,53 +68,92 @@ func runTroubleshoot(v *viper.Viper, arg string) error {
...
+		collectorContent, err := supportbundle.LoadSupportBundleSpec(v)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get redactor spec %s, #%d", redactor, idx)
+			return errors.Wrap(err, "failed to load collector spec")
 		}
-
-		if redactorObj != nil {
-			additionalRedactors.Spec.Redactors = append(additionalRedactors.Spec.Redactors, redactorObj.Spec.Redactors...)
+		multidocs := strings.Split(string(collectorContent), "\n---\n")
+		supportBundle, err := supportbundle.ParseSupportBundleFromDoc([]byte(multidocs[0]))
```
divolgin, line 96: This shouldn't assume that the first doc is support bundle

pkg/apis/troubleshoot/v1beta2/supportbundle_interfaces.go
```diff
@@ -0,0 +1,58 @@
...
+            for _,v := range s.Collectors{
+                if v.ClusterInfo != nil {
+                    y = 1
```
divolgin: this can `break`

pkg/apis/troubleshoot/v1beta2/supportbundle_interfaces.go
```diff
@@ -0,0 +1,58 @@
...
+                s.Collectors = append(s.Collectors, c)
+            }
+            continue
+        }
+
+        if c.ClusterResources != nil {
+            y := 0
+            for _,v := range s.Collectors {
+                if v.ClusterResources != nil {
+                    y = 1
```
divolgin: this can `break`

danj-replicated:
After talking to @diamonwiggins I'm going to move the new methods from the api package to the support-bundle package

pkg/apis/troubleshoot/v1beta2/supportbundle_interfaces.go
danj-replicated: I removed these in favor of addressing them properly in another task

pkg/supportbundle/supportbundle.go
```diff
@@ -226,3 +226,24 @@ func AnalyzeSupportBundle(spec *troubleshootv1beta2.SupportBundleSpec, tmpDir st
 	}
 	return analyzeResults, nil
 }
+
+// the intention with these appends is to swap them out at a later date with more specific handlers for merging the spec fields
+func ConcatSpec(target *troubleshootv1beta2.SupportBundle, source *troubleshootv1beta2.SupportBundle) *troubleshootv1beta2.SupportBundle {
+	newBundle := target.DeepCopy()
+	for _, v := range source.Spec.Collectors {
```
divolgin: No need for any of these loops
```
newBundle.Spec.Collectors = append(target.Spec.Collectors, source.Spec.Collectors...)
```

danj-replicated: Thanks! I didn't know about this magic variadic argument syntax

xavpaice:
I've tested this manually by adding several bundles at the CLI, some with multiple docs, and a redactor spec.  All the bundles were merged as expected, and the redactors ran fine.

The changes requested have been taken care of except the go fmt, which I'd like to put into a separate commit.

The task that relates to this shouldn't be closed without adding tests, but given the test coverage in the current implementation, I think that can be a separate PR.

divolgin (dismissed):

xavpaice (approved):
LGTM.  Future options for testing is to add something for runTroubleshoot, worth considering for a future PR.

Issue #694, opened 2022-08-31
Add logic to update a spec if the URI field is defined


xavpaice:
## Request

Design: #681 

When presented with a spec document that contains a URI, if that spec is one of the specs provided on the command line or initial call to Troubleshoot, attempt to collect a replacement spec from that URI.  If successful, replace the entire spec with the one downloaded.  If unsuccessful, log the error and use the spec provided.

The URI is provided based on changes in #682 :

```yaml
apiVersion: troubleshoot.sh/v1beta2
kind: SupportBundle
metadata:
  name: default
spec:
  uri: https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/in-cluster/default.yaml
  collectors:
```

Likely starting point for this is to alter `ParseSupportBundleFromDoc()`in `pkg/supportbundle/load.go`.

We will need to test if the resulting supportBundle has a uri, and if it does, use the `loadSpec()` function to load a replacement spec.

## In scope

* Replacing the first spec document in a yaml file/secret/URL provided on the CLI with a URL download result (oci://, http://, https://), if successful
* Optionally, this URI could direct to a file, since `loadSpec()` tests for that already

## Out of scope

* Loading specs where the URI points to a secret (that's a different code path).
* yaml files with multiple documents will only read the URI from the first document in the file
* Redactors are out of scope, just the content of `collectors`and `analyzers`will be replaced
* preflights are out of scope
* KOTs calls `troubleshootv1beta2.CollectSupportBundleFromURI()`and `troubleshootv1beta2.CollectSupportBundleFromSpec()`to trigger support bundle collection.  These calls are not expected to be affected by these changes.

## Definition of done

- [x] CLI can accept a yaml file that has a `uri`specified (as above), and will download a replacement spec from that URI.
- [x] If the download fails, Troubleshoot continues to run without error, logging the failure only.  The originally provided spec is used.
- [ ] Code changes have tests

Docs to be added in #711

adamancini:
@xavpaice have you considered how we might handle circular references in the URI specified in the spec?  Something David and I discussed today is perhaps we can keep track of the hash of the spec referenced in the URI and break a circular reference:

imagine

## https://foo.bar.com/spec1.yaml
uri: https://foo.bar.com/spec2.yaml


## https://foo.bar.com/spec2.yaml
uri: https://foo.bar.com/spec3.yaml


## https://foo.bar.com/spec3.yaml
[...]

Issue #682, opened 2022-08-26
Add URI type to spec definition


xavpaice:
Add a new field to the Troubleshoot spec definition, will enable a URI to be used to locate another spec.

In PR https://github.com/replicatedhq/troubleshoot/pull/681 we describe using a new field in the Troubleshoot spec definition to allow a URI to be specified for a replacement (i.e. updated) spec for the spec provided.

The spec URI listed should be any valid location that Troubleshoot can accept on the CLI - this could be a secret, file, or an address.

This particular task is to update the CRDs so that when a spec is provided to Troubleshoot, it can include a new field per the example below.

Current spec format is like this example:

```yaml
apiVersion: troubleshoot.sh/v1beta2
kind: SupportBundle
metadata:
  name: default
spec:
  collectors:
    - clusterInfo: {}
    - clusterResources: {}
  analyzers:
    - cephStatus: {}
    - longhorn: {}
```

We could add a new type, `uri` under the top level spec, which contains a list of URIs from which to obtain specs:

```yaml
apiVersion: troubleshoot.sh/v1beta2
kind: SupportBundle
metadata:
  name: default
spec:
  uri: https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/in-cluster/default.yaml
  collectors:
    - clusterInfo: {}
    - clusterResources: {}
  analyzers:
    - cephStatus: {}
    - longhorn: {}
```

In scope:

* adding the ability to accept the new data without error

Out of scope:

* doing anything with that new data

Definition of done:

- [x] Troubleshoot can accept a spec yaml with the second format provided above, without error
- [x] Any tests which validate spec definition are updated (or written)
- [x] A draft (but not available online) document detailing the required format

chris-sanders:
I mentioned this on the design doc, but since we're going to run a bit ahead and start on this I'll say it here as well.

I think we should just use `uri` instead of `specURI` because the full path as shown above is `spec.specURI` which is repetitive.

edgarlanting:
On todays CRE team call we pretty much agreed to use `uri` to minimize stuttering amongst others. So for example; instead of `spec`.`specURI` we'd want to use `spec`.`uri`

edgarlanting:
`./cmd/schemagen/cli/root.go` generates a new schema in `./schemas`.

These are:

`analyzer-troubleshoot-v1beta2.json`
`preflight-troubleshoot-v1beta2.json`
`supportbundle-v1beta2.json`

[...]

PR #696, opened 2022-09-01
EL 20220901 - Implement new type  for #682

edgarlanting:
This PR implements a new type for Troubleshoot as per issue #682 

Added are:

- A new type: `uri` to specify a spec located on web, a secret in a cluster or a spec yaml file
- An updated sample spec containing the new uri field

Actual functionality will be built into Troubleshoot with #694 

! to implement in a new build, run `make schemas` before running `make support-bundle`.

While automated test options are not yet built into to check spec validity et al, I have manually tested this new version/PR with the following current and older versions of Troubleshoot:

v0.34.0 (May 2022 - arm64 & amd64) 
v0.28.1 (January 2022 - arm64 & amd64)
v0.13.14 (September 2021 - amd64) 

And a 'new' spec to test with:

```
kind: SupportBundle
metadata:
  name: default
spec:
  uri: https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/in-cluster/default.yaml
  collectors:
    - clusterInfo: {}
    - clusterResources: {}
  analyzers:
    - cephStatus: {}
    - longhorn: {}
```

### DIFF
--- a/pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
+++ b/pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
@@ -27,6 +27,13 @@ type SupportBundleSpec struct {
 	HostCollectors  []*HostCollect     `json:"hostCollectors,omitempty" yaml:"hostCollectors,omitempty"`
 	Analyzers       []*Analyze         `json:"analyzers,omitempty" yaml:"analyzers,omitempty"`
 	HostAnalyzers   []*HostAnalyze     `json:"hostAnalyzers,omitempty" yaml:"hostAnalyzers,omitempty"`
+	// Uri defines the location of the spec file that needs to be used. This is optional and can be
+	// either a secret, url or spec yaml file.
+	// When presented with a spec document that contains a URI, if that spec is one of the specs provided
+	// on the command line or initial call to Troubleshoot, attempt to collect a replacement spec from that URI.
+	// If successful, it will replace the entire spec with the one downloaded.
+	// If unsuccessful, log the error and use the spec provided.
+	Uri string `json:"uri,omitempty" yaml:"uri,omitempty"`
 }
 
 // SupportBundleStatus defines the observed state of SupportBundle
--- a/pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
+++ b/pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
@@ -27,12 +27,7 @@ type SupportBundleSpec struct {
 	HostCollectors  []*HostCollect     `json:"hostCollectors,omitempty" yaml:"hostCollectors,omitempty"`
 	Analyzers       []*Analyze         `json:"analyzers,omitempty" yaml:"analyzers,omitempty"`
 	HostAnalyzers   []*HostAnalyze     `json:"hostAnalyzers,omitempty" yaml:"hostAnalyzers,omitempty"`
-	// Uri defines the location of the spec file that needs to be used. This is optional and can be
-	// either a secret, url or spec yaml file.
-	// When presented with a spec document that contains a URI, if that spec is one of the specs provided
-	// on the command line or initial call to Troubleshoot, attempt to collect a replacement spec from that URI.
-	// If successful, it will replace the entire spec with the one downloaded.
-	// If unsuccessful, log the error and use the spec provided.
+	// URI optionally defines a location which is the source of this spec to allow updating of the spec at runtime
 	Uri string `json:"uri,omitempty" yaml:"uri,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
```diff
@@ -27,6 +27,13 @@ type SupportBundleSpec struct {
 	HostCollectors  []*HostCollect     `json:"hostCollectors,omitempty" yaml:"hostCollectors,omitempty"`
 	Analyzers       []*Analyze         `json:"analyzers,omitempty" yaml:"analyzers,omitempty"`
 	HostAnalyzers   []*HostAnalyze     `json:"hostAnalyzers,omitempty" yaml:"hostAnalyzers,omitempty"`
+	// Uri defines the location of the spec file that needs to be used. This is optional and can be
+	// either a secret, url or spec yaml file.
+	// When presented with a spec document that contains a URI, if that spec is one of the specs provided
+	// on the command line or initial call to Troubleshoot, attempt to collect a replacement spec from that URI.
+	// If successful, it will replace the entire spec with the one downloaded.
+	// If unsuccessful, log the error and use the spec provided.
+	Uri string `json:"uri,omitempty" yaml:"uri,omitempty"`
```
chris-sanders, line 31: I'm curious the use case for anything other than a URL for this field. Is there a a reason to use a secret or yaml file in this reference?

This nuance may not matter for this actual change, which if true this comment is overly verbose I think. I like documenting in code, but if the comment documents things that aren't implemented here it runs the risk of being incorrect when other code changes and that change wouldn't know to come update this comment.

To me, I would simply include "URI optionally defines a location which is the source of this spec to allow updating of the spec at runtime".

chris-sanders (commented):
one small comment

pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
diamonwiggins: Agree with Chris' comments above. Also for:

> If successful, it will replace the entire spec with the one downloaded.

I didn't know this was our intention?

xavpaice: Regards why we might use a secret, yaml file, or URL with an oci:// type, it's possible that these could be handy but far from essential.  The reality is that the simplest way to use the code to ingest the content of the URI could be to use the same logic that's already there, which includes these types.

I do agree that it would be good to be less specific here with the type of location - the code change to implement use of that will dictate the end result though it looks currently like the simplest change is to accept any type of URI.

> > If successful, it will replace the entire spec with the one downloaded.
> 
> I didn't know this was our intention?

yup, that's the intent - see https://github.com/replicatedhq/troubleshoot/blob/main/design/proposal-updateable-specs.md

xavpaice (commented):
one nit with inline comments, otherwise LGTM.

pkg/apis/troubleshoot/v1beta2/supportbundle_types.go
edgarlanting: > I'm curious the use case for anything other than a URL for this field. Is there a a reason to use a secret or yaml file in this reference?
> 
> This nuance may not matter for this actual change, which if true this comment is overly verbose I think. I like documenting in code, but if the comment documents things that aren't implemented here it runs the risk of being incorrect when other code changes and that change wouldn't know to come update this comment.
> 
> To me, I would simply include "URI optionally defines a location which is the source of this spec to allow updating of the spec at runtime".

Thanks @chris-sanders, it makes perfect sense indeed to minimize the explanation and clarify a comment line into just one sentence for this new field.

edgarlanting: > Regards why we might use a secret, yaml file, or URL with an oci:// type, it's possible that these could be handy but far from essential. The reality is that the simplest way to use the code to ingest the content of the URI could be to use the same logic that's already there, which includes these types.
> 
> I do agree that it would be good to be less specific here with the type of location - the code change to implement use of that will dictate the end result though it looks currently like the simplest change is to accept any type of URI.
> 
> > > If successful, it will replace the entire spec with the one downloaded.
> > 
> > 
> > I didn't know this was our intention?
> 
> yup, that's the intent - see https://github.com/replicatedhq/troubleshoot/blob/main/design/proposal-updateable-specs.md

Thanks @xavpaice for your feedback!

edgarlanting: > Agree with Chris' comments above. Also for:
> 
> > If successful, it will replace the entire spec with the one downloaded.
> 
> I didn't know this was our intention?

Thanks @diamonwiggins for your feedback!

edgarlanting (commented):
Thanks all for the feedback, requesting a re-review as the comment line has been changed.

chris-sanders (approved):
lgtm

PR #702, opened 2022-09-07
`uri:` implementation

### LINKED ISSUE #694: Add logic to update a spec if the URI field is defined

adamancini:
if a spec is provided that specifies a `uri:` field, get the referenced spec from upstream and use it.  if getting the upstream spec fails, fall back to what is provided.

avoid a recursion loop by adding a `recurse` argument to `GetSupportBundleFromURI()` and `ParseSupportBundleFromDoc()` so that it only performs this operation once.

### DIFF
--- a/pkg/supportbundle/load.go
+++ b/pkg/supportbundle/load.go
@@ -15,6 +15,7 @@ import (
 	troubleshootclientsetscheme "github.com/replicatedhq/troubleshoot/pkg/client/troubleshootclientset/scheme"
 	"github.com/replicatedhq/troubleshoot/pkg/docrewrite"
 	"github.com/replicatedhq/troubleshoot/pkg/httputil"
+	"github.com/replicatedhq/troubleshoot/pkg/logger"
 	"github.com/replicatedhq/troubleshoot/pkg/oci"
 	"github.com/replicatedhq/troubleshoot/pkg/specs"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,7 +29,7 @@ func GetSupportBundleFromURI(bundleURI string) (*troubleshootv1beta2.SupportBund
 
 	multidocs := strings.Split(string(collectorContent), "\n---\n")
 
-	supportbundle, err := ParseSupportBundleFromDoc([]byte(multidocs[0]))
+	supportbundle, err := ParseSupportBundle([]byte(multidocs[0]), true)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse collector")
 	}
@@ -36,7 +37,7 @@ func GetSupportBundleFromURI(bundleURI string) (*troubleshootv1beta2.SupportBund
 	return supportbundle, nil
 }
 
-func ParseSupportBundleFromDoc(doc []byte) (*troubleshootv1beta2.SupportBundle, error) {
+func ParseSupportBundle(doc []byte, followURI bool) (*troubleshootv1beta2.SupportBundle, error) {
 	doc, err := docrewrite.ConvertToV1Beta2(doc)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to convert to v1beta2")
@@ -49,6 +50,9 @@ func ParseSupportBundleFromDoc(doc []byte) (*troubleshootv1beta2.SupportBundle, 
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse document")
 	}
+
+	// parse doc and detect if it's a SupportBundle type,
+	// if it's a Collector type, convert it to a SupportBundle
 
 	collector, ok := obj.(*troubleshootv1beta2.Collector)
 	if ok {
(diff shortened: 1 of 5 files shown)


adamancini:
resolves #694

adamancini:
still need to add tests for the new code path

xavpaice:
The change to add an arg to `ParseSupportBundleFromDoc` might cause some issues with kots, as it's used in a couple of places there:

```
./pkg/supportbundle/helm.go:	supportBundle, err := troubleshootsb.ParseSupportBundleFromDoc([]byte(multidocs[0]))
./pkg/supportbundle/defaultspec/embed.go:	spec, err = supportbundle.ParseSupportBundleFromDoc(raw)
```

https://bojanz.github.io/optional-parameters-go/ has some options which might help, maybe using a wrapper function to keep things clear.  Alternately, we might need to coordinate a change in kots at the same time, but I think it'd be best if we could keep kots changes out of scope at this stage.

pkg/supportbundle/load.go
```diff
@@ -15,28 +15,29 @@ import (
...
+	supportbundle, err := ParseSupportBundleFromDoc([]byte(multidocs[0]), recurse)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to parse collector")
 	}
 
 	return supportbundle, nil
 }
 
-func ParseSupportBundleFromDoc(doc []byte) (*troubleshootv1beta2.SupportBundle, error) {
+func ParseSupportBundleFromDoc(doc []byte, recurse bool) (*troubleshootv1beta2.SupportBundle, error) {
```
xavpaice: what we could do rather than add the arg here, is to create a new function (I am totally not attached to the name):

```
func ParseSupportBundleFromDocRecursed(doc []byte, recurse bool) (*troubleshootv1beta2.SupportBundle, error) {
   <the logic pulled from ParseSupportBundleFromDoc()>
}

```

Then have a new `ParseSupportBundleFromDoc`:

```
func ParseSupportBundleFromDoc(doc []byte) (*troubleshootv1beta2.SupportBundle, error) {
    return ParseSupportBundleFromDocRecursed(doc, false)
}
```

That way, the original function name is still there without modification (for kots import), but the new functionality is available.

adamancini:
@xavpaice have a look at the latest and let me know if this is closer on-track

~I'm still a little unsure on the error-handling parts around falling back, though - if we return an error and also a working spec, `if err != nil` blocks might still trigger and we might fail when there actually is a usable spec~

pkg/supportbundle/load.go
```diff
@@ -70,12 +74,36 @@ func ParseSupportBundleFromDoc(doc []byte) (*troubleshootv1beta2.SupportBundle,
...
+				logger.Printf("failed to load upstream supportbundle, falling back")
+				return supportBundle, nil
+			}
+
+			multidocs := strings.Split(string(upstreamSupportBundleContent), "\n---\n")
+
+			upstreamSupportBundle, err := ParseSupportBundle([]byte(multidocs[0]), false)
+			if err != nil {
+				logger.Printf("failed to parse upstream supportbundle, falling back")
+				return supportBundle, nil
```
diamonwiggins, line 93: should we be returning an error here?

adamancini: I want to say "yes" but in code that calls the containing function, if we return `supportBundle, err` and the containing codes checks for `if err != nil` then we may still exit with an error even though a valid supportBundle gets returned, and then any code that calls this might have to change their error handling

pkg/supportbundle/supportbundle_test.go
```diff
@@ -7,40 +7,71 @@ import (
...
+
+	if reflect.DeepEqual(providedSpec, badURISpec) == false {
+		t.Errorf("expected bad URI to fall back to provided bundle.  expected %v but got %v", providedSpec, badURISpec)
+	}
+
+	//if we can't parse the upstream doc, we should fall back to the provided
+	badYAMLBundle, _ := LoadSupportBundleSpec("test/bad-upstream-uri-spec.yaml")
+	badYAMLSpec, _ := ParseSupportBundleFromDoc(badYAMLBundle)
+	providedBundle, _ = LoadSupportBundleSpec("test/bad-upstream-uri-spec.yaml")
+	providedSpec, _ = ParseSupportBundle(providedBundle, false)
```
diamonwiggins: Wonder if this should be in a separate test since we aren't testing concatenating here?

adamancini: yeah makes sense, I need to move the upstream specs out of my `troubleshoot-specs` fork and into the test itself

diamonwiggins:
+1 on your offline comment about the tests manifests you're using:

> “upstream” manifests should be moved into troubleshoot itself or just expressed as code in the test itself

pkg/supportbundle/load.go
diamonwiggins: understood. i guess i'm curious what the behavior should be in the event an upstream bundle cannot be parsed. Does it fallback and return the support-bundle spec with the `URI` field to the list to be run. Do you need to always specify `collectors` and `analyzers` in your spec even when using the URI field so that it doesn't panic?

adamancini: if the upstream bundle has invalid YAML, it should fall back to simply the contents in the current bundle - but that's a good point, I didn't add a test case with _only_ a uri: field so - adding that to the to-do list

xavpaice:
One question, for the docs probably, is what happens when a spec yaml file contains multi docs - e.g. a SupportBundle type and a Redactor.  Does the entire file get replaced, or the single doc?  We should document the behavior in the docs change, regardless of the answer, just so it's clear for folks using the software.

adamancini:
@xavpaice with this change the entire spec document gets replaced

adamancini:
#759 for followup

adamancini:
https://github.com/replicatedhq/troubleshoot.sh/pull/418 for docs

pkg/supportbundle/test/bad-uri-spec.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: default
+spec:
+  uri: https://githubusercontent.com/adamancini/troubleshoot-specs/main/host/cluster-down
```
diamonwiggins, line 6: should these bad uris be more generic?

diamonwiggins (approved):

diamonwiggins:
Only other question I had is what will behavior be if the upstream spec has another URI in it? Is it simply ignored?

PR #709, opened 2022-09-11
Load support bundle specs from secrets matching label selector

### LINKED ISSUE #699: Add support to CLI to search for all secrets that contain Troubleshoot specs
## Request

Ref design #663.

Currently we can specify a secret with the URI for a Troubleshoot spec.  With #683 we add the ability to list multiple specs on the CLI, but not a wildcard, directory, or search pattern.

This request is to implement a search function whereby any secrets that can be found with a matching label (default to  `troubleshoot.io/kind=supportbundle-spec`), will be collected and used as if they had been specified on the CLI.  This search should be made in the namespace defined in the same way as the kubectl client.

## In scope

* add a switch to the CLI, which triggers Troubleshoot to search for secrets
  * in the namespace specified with `-n` or `--namespace`, or the current context for the Kubernetes client
  * all namespaces if `-A` or `--all-namespaces` is provided
  * for any secret with the label `troubleshoot.io/kind=supportbundle-spec` unless the `-l` switch is provided to override that
* Include each secret found as if it had been specified on the CLI (i.e. use them all)
* Test coverage
* Documentation

## Out of scope

* KOTS changes to make use of this addition

## NOTE

The package `"k8s.io/cli-runtime/pkg/genericclioptions"` may be helpful with the kubectl options, see [this example](https://github.com/kubernetes/sample-cli-plugin).

This task may be affected by work on #694, which uses the URI specified in a spec only on the specs provided on the CLI.  To be clear, the specs found in secrets with this task need to be considered as specs provided on the CLI, therefore the URI will need to be read and used.

Also: Note that KOTS uses the `kots.io/kind=supportbundle`label [here](https://github.com/replicatedhq/kots/blob/main/pkg/store/kotsstore/supportbundle_store.go#L239) to describe collected support bundles.

Kots does store a Troubleshoot spec, with the name `kotsadm-$APPSLUG-supportbundle`, however it does not currently have a label which can be used to identify it.

## Definition of done

- [x] CLI switch enabling Troubleshoot to search for and use all secrets with a label `troubleshoot.io/kind=supportbundle-spec` (default), or a CLI specified label
- [x] CLI supports kubectl's `-l, --selector=''` switch to override the label searched for
- [x] CLI supports kubectl's `-A`, `--all-namespaces`, `-n`, `--namespace` switches to override namespace
- [x] Unit tests for the new implementation
- [ ] Documentation updated to show usage of this new feature in https://troubleshoot.sh/docs

diamonwiggins:


xavpaice:
Converted to draft - need to add tests before we fully review and merge.

cmd/troubleshoot/cli/run.go
```diff
@@ -91,23 +95,47 @@ func runTroubleshoot(v *viper.Viper, arg []string) error {
...
-			if !ok {
-				continue
+		parsedRedactors, err := supportbundle.ParseRedactorsFromSpec(multidocs)
+		if err != nil {
+			return errors.Wrap(err, "failed to parse redactors from doc")
+		}
+		additionalRedactors.Spec.Redactors = append(additionalRedactors.Spec.Redactors, parsedRedactors...)
+	}
+
+	if v.GetBool("load-specs-from-secrets") {
```
chris-sanders: While clear, this is a fairly long flag for a user to type. Does it/can it also have a short version? If it can't have a long and short versions can we reduce the flag to be shorter?

chris-sanders (commented):
Just one comment in-line thus far.

diamonwiggins:
I've added tests. I've also dropped the minimum number of arguments down to 0 and added some additional handling for when no args are provided and `load-cluster-specs` isn't set. If you don't use the flag or specify any args you will get:

> flag load-cluster-specs must be set if no specs are provided on the command line

and if you don't specify any spec files on the cli, but you use `load-cluster-specs` and it doesn't find anything in the cluster you get:

> no specs provided to run

chris-sanders:
Maybe the message if no specs are found should be `no specs found in cluster` or something similar? If someone provides the flag and gets the message that no spec was provided they might assume it didn't even look in cluster. I might even suggest if the namespace flag was provided to add that as well `no specs found in cluster namespace X`.

Have you tested what happens if there is no configured cluster or you fail to access the cluster? Does that provide an appropriate alternate message?

adamancini (approved):

cmd/troubleshoot/cli/run.go
```diff
@@ -91,24 +101,73 @@ func runTroubleshoot(v *viper.Viper, arg []string) error {
...
+				parsedRedactors, err := supportbundle.ParseRedactorsFromSpec(multidocs)
+				if err != nil {
+					logger.Printf("failed to parse redactors from doc:  %s", err)
+					continue
+				}
+				additionalRedactors.Spec.Redactors = append(additionalRedactors.Spec.Redactors, parsedRedactors...)
 			}
-			additionalRedactors.Spec.Redactors = append(additionalRedactors.Spec.Redactors, multidocRedactors.Spec.Redactors...)
 		}
+		if mainBundle == nil {
```
divolgin, line 162: This check is not needed.  It's duplicated on line 167

Issue #695, opened 2022-08-31
Add argument to suppress runtime updates


xavpaice:
In #694, we add the ability to specify a `uri`in the support bundle definition provided to Troubleshoot, to download an updated spec.

In design #681, we discuss having the option to disable that, based on a CLI switch.

## In scope

* CLI usage of Troubleshoot via the `support-bundle`binary, providing Viper arguments

## Out of scope

* Any calls to Troubleshoot other than that in package `cli`, `runTroubleshoot()`

## Definition of done

- [x] the Troubleshoot CLI accepts a `--follow-uri` boolean switch which disables any action taken for the `uri`field and sets `followURI` to `false`.  Default (i.e. switch not provided) is to download from the uri.
- [x] changes have tests
- [x] the new CLI option is documented in code (via `--help`)

edgarlanting:
> `--ignore-uri` sounds fine to me, it's more specific than `--no-updates`.

Thanks @xavpaice I was initially going for `noupdates` but `ignore-uri` is more clear 👍🏻

xavpaice:
`--ignore-uri` sounds fine to me, it's more specific than `--no-updates`.

edgarlanting:
Current steps for implementation of the `--ignoreuri` flag are;

When passed as `support-bundle --ignoreuri <spec.yaml>` the flag will make sure that `followURI` is set to `false` as defined in ./pkg/supportbundle.load.go will be ignored:

```
		// check if there is a uri field and if so,
		// use the upstream spec, otherwise fall back to
		// what's defined in the current spec
		if supportBundle.Spec.Uri != "" && followURI {
			logger.Printf("using upstream reference: %+v\n", supportBundle.Spec.Uri)
			upstreamSupportBundleContent, err := LoadSupportBundleSpec(supportBundle.Spec.Uri)
			if err != nil {
				logger.Printf("failed to load upstream supportbundle, falling back")
				return supportBundle, nil
			}
....
```
I am using @adamancini's branch at: https://github.com/replicatedhq/troubleshoot/pull/702/files for reference.

~I expect to have a draft PR ready this Thursday for review.~

edgarlanting:
Current status of this task:
 - Add functionality to `./pkg/supportbundle/load.go` in order to set `followURI` to `false`
 - Add functionality to `./cmd/troubleshoot/cli/root.go / run.go` in order to make the flag functional and be able to use it.

[...]

PR #717, opened 2022-09-16
Issue #695 - Add `--no-uri` flag to `support-bundle` to ignore the uri field and revert back to the default spec…

edgarlanting:
For Troubleshoot issue #695 we described the need for an additional flag that will allow a user to ignore the uri field in a spec if that field contains a value.

This flag has been implemented by introducing some changes to our CLI and does not interfere with KOTS functionality as KOTS does not use the CLI for Troubleshoot functionality. (It directly deals with `load.go`)

The uri field implementation that precedes this change can be found [here](https://github.com/replicatedhq/troubleshoot/pull/702) and is needed to use this newly introduced `support-bundle` flag.

To implement this, a small change was also needed to ensure continuity of Viper functionality in `run.go`. This change consists of the following:

Line 89 of `run.go` implements a for loop of `for i, v := range arg {`

Unfortunately, `v` then renders its predecessor, `v` when used for Viper useless. Therefore I implemented a workaround of:
`for i, val := range arg {` to rewrite this for loop and still be able to use `v` for Viper references.

To successfully refer to `followURI` as implemented in `./pkg/supportbundle/load.go`, I changed the use of `ParseSupportBundleFromDoc` on line `100` to use `ParseSupportBundle` and added a secondary argument of `v.GetBool("follow-uri"))` to specify the newly introduced flag.

To use this flag, a user can run:

`support-bundle --no-uri <spec-with-uri-entry.yaml>`

And the support-bundle generation will revert back to the default spec in that file and will ignore the url specified in the `uri` field.

Troubleshoot help (`support-bundle --help`) now also includes a reference to this flag:

```
      --no-uri               Forces the support bundle to revert back to the default and not use a URI for a spec when passed on the commandline (default true)
```


edgarlanting:
FYI that this step fails to complete : https://github.com/replicatedhq/troubleshoot/actions/runs/3067800529/jobs/4954490433 

This is related to the preceding change ([add uri](https://github.com/replicatedhq/troubleshoot/pull/702)) is currently not implemented.

xavpaice:
perhaps you could merge this one to Ada's branch with the PR #702, then it can all land together?

Also, this could do with test coverage.

edgarlanting:
To test the functionaltity & behaviour, I am using the following spec file that contains all the necessary test materials:

```
kind: SupportBundle
metadata:
  name: default
spec:
  uri: https://raw.githubusercontent.com/adamancini/troubleshoot-specs/main/host/cluster-down.yaml
  collectors:
    - clusterInfo: {}
    - clusterResources: {}
  analyzers:
    - cephStatus: {}
    - longhorn: {}
```

This results in the way that when `--no-uri` is not passed as an option/flag, the `uri` fields contents of `https://raw.githubusercontent.com/adamancini/troubleshoot-specs/main/host/cluster-down.yaml` is being used.

When `--no-uri` is passed as an option/flag, we revert back to the default spec values in the file.

banjoh (dismissed):

cmd/troubleshoot/cli/root.go
```diff
@@ -58,6 +58,10 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 	cmd.Flags().Bool("allow-insecure-connections", false, "when set, do not verify TLS certs when retrieving spec and reporting results")
 	cmd.Flags().MarkHidden("allow-insecure-connections")
 
+	// `no-uri` references the `followURI` functionality where we can use an upstream spec when creating a support bundle
+	// This flag makes sure we can also disable this and fall back to the default spec.
+	cmd.Flags().Bool("no-uri", false, "Forces the support bundle to revert back to the default and not use a URI for a spec when used.`")
```
adamancini: I think the only nit I have here is in the usage docs, I might emphasize the outcome instead of language like "revert to the default" since someone might not know what the default behaviour was, and I think the "new" default behaviour is that we *will* follow a URI.

Maybe something like "When set to `true`, do not attempt to retrieve the bundle referenced by the `uri:` field"

xavpaice: agreed

xavpaice (changes requested):

edgarlanting (commented):
@xavpaice just rewrote the docs/help text. Please review

cmd/troubleshoot/cli/root.go
edgarlanting: Thanks Ada & Xav!

edgarlanting: @adamancini changed this, thanks for reviewing!

edgarlanting (commented):
@xavpaice I think your requested change has been done by rewriting the `--help` text. Agree?

cmd/troubleshoot/cli/run.go
```diff
@@ -83,14 +83,21 @@ func runTroubleshoot(v *viper.Viper, arg []string) error {
...
+
+		collectorContent, err := supportbundle.LoadSupportBundleSpec(val)
 		if err != nil {
 			return errors.Wrap(err, "failed to load support bundle spec")
 		}
 		multidocs := strings.Split(string(collectorContent), "\n---\n")
-		supportBundle, err := supportbundle.ParseSupportBundleFromDoc([]byte(multidocs[0]))
+		// Referencing `ParseSupportBundle with a secondary arg of `no-uri`
+		// Will make sure we can enable or disable the use of the `Spec.uri` field for an upstream spec.
+		// This change will not have an impact on KOTS' usage of `ParseSupportBundle`
```
adamancini, line 98: Only other comment I might have is maybe we don't need to reference KOTS in documentation?  Otherwise I think this lgtm.

adamancini (approved):

cmd/troubleshoot/cli/run.go
edgarlanting: That was for internal references. We can indeed just as easy take that part out

xavpaice (approved):
lgtm

PR #718, opened 2022-09-16
Update CONTRIBUTING.md

adamancini:
add notes about squashing history


chris-sanders (approved):
Wow I'm really curious if you did this because you noticed it was a good practice or if you noticed I enforced squash merging today in the repo while updating some settings. Either way, nice update.

PR #720, opened 2022-09-19
Revert part of PR714 FOSSA scan

xavpaice:
PR #714 added a FOSSA scan which needs further work before it passes for any PRs from forks.  This reverts that change till it's working.


xavpaice:
abandon in favor of #721

PR #729, opened 2022-09-21
Collect resourcequotas

### LINKED ISSUE #719: Resource Quota collection
### Problem to solve
Sometimes the problem is caused by ResourceQuota being too low. Cannot schedule pods when the limit is reached. Troubleshoot tool does not collect ResourceQuota config and usage.

### Impact
We wasted lots of time troubleshooting before realizing the problem with ResourceQuota.

### Proposed Solution
Collect ResourceQuota info. Similar to `kubectl get ResourceQuota`.  This could be added to clusterResources.

danj-replicated:
Closes #719


xavpaice:
Just one note - is it possible to add tests to the change?

danj-replicated:
I didn't get chance to finish the tests today so I'll leave an update with the plan:

there's currently no testing for the functions that get the data for the cluster-resources collector, so I'm starting with just the one I added in this PR. I'm going to create some mock resourceQuota resources for the test and run them through the same table-testing we do for other parts of the code.

danj-replicated:
an update upon reading deeper into these cluster resource collectors. I think the current state of testing can be explained by the functionality being essentially a thin wrapper around go's kubernetes client and json.Marshal(), as such we'd be testing things outside of our control.

edgarlanting (approved):
lgtm!
-Edgar

xavpaice (approved):
lgtm.

PR #734, opened 2022-09-22
remove IP address redaction

### LINKED ISSUE #735: Make redaction of IP addresses opt-in

Many times during troubleshooting issues with networking, particularly Weave issues, we ask folks to retrieve logs and other info that include IP addresses.  While this information is redacted from all Troubleshoot bundles, we do not see the IP addresses and so have to ask outside of Troubleshoot for the information if it is necessary to have to understand the problem.

**Describe the feature**

* Remove the IP address redaction from the Troubleshoot default redactors in the Troubleshoot code itself
* Update the docs to remove the statement that it's redacted by default
* Folks that want IP address redaction need to go and add that into their custom redactors (as documented)

* Indexing redactors - so that e.g. address `192.168.1.1` gets replaced with something specific to that address, e.g. `a.b.c.d`.  This would relieve many of the issues we have with redaction right now, however if an end user needs to go check a firewall rule they are going to need the addresses, not the index.
* CLI switch to disable _some_ redactors including IP addresses.  This would mean anyone using the CLI has the option to not redact IP addresses, which keeps the default behavior, but does require folks to potentially collect a bundle twice (once after learning that IP info is required).

**Additional context**

Environments with particularly restrictive security requirements do require IP addresses to be redacted.  This request changes the default to not redact, rather than redact.

xavpaice:


xavpaice:
Further discussion needed: "Many government agencies and a few other orgs treat ip addresses as sensitive". If we remove this redaction from the defaults, everyone who needs it needs to add that to their redactor spec (via KOTS or manually), and running from the CLI risks missing the redactor if it's not entirely clear.

xavpaice:
Depends on agreement for #823

chris-sanders (changes requested):
Yup this looks good to me, let's be sure to coordinate documentation before we merge this.

xavpaice:
Let's hold merging this until docs PR is ready, plus release notes and announcement are all ready

xavpaice:
Docs changes: https://github.com/replicatedhq/troubleshoot.sh/pull/435 and https://github.com/replicatedhq/replicated-docs/pull/780

chris-sanders (approved):
Yup still looking good

PR #742, opened 2022-09-27
Add logging design proposal

### LINKED ISSUE #672: Create proposal for streamlined log collection

Log collection has grown organically and is need of a review for UX. The collection and analysis of logs while possible is confusing. Here are some examples I'm aware of directly.

1. The `logs` collector, as documented, only collects logs from running pods. If a pod fails to start, even if some containers start and others don't the collector produces nothing. This means an analyzer for the logs collector can only identify issues for healthy containers which is fairly limiting.
2. The `logs` collector stores logs, as all collectors, in a collector specific folder based on the name of the collector. Because of item 1 above this folder may or may not appear in a support bundle depending if it was able to run. This adds to confusion as logs seem to 'sometimes' be present.
3. The `cluster-resources` collector only collects logs from pods that fail to start. This collector therefore has information on failed logs but doesn't have information pods that are running, that requires the above `logs` collector.
4. The `sbctl` tool only knows about `cluster-resources` collector so the only logs you can ever find with it are logs for pods that are unhealthy.
5. Other collectors like the `ceph` collector (which is actually a Rook collector) also store logs which are unknown to sbctl and it is unclear from documentation how it behaves with regards to healthy or unhealthy pods.

**Describe the feature**

Fixing this requires a big picture review of the UX and interactions between various collectors and the `sbctl` tool. I won't attempt to design the solution in this issue but instead request that a design document be developed and discussion had to determine the solution. Goals for this design then should include:

1. Ensure all logs that are collected are accessible to `sbctl`
2. Improve usability for users who desire to review logs for pods in all stages of health. This need not necessarily require sweeping changes to collectors if it's determined there is a well thought out process to do this if documenting that process solves the issue. Although I have my doubts that is the case.
3. Ensure the final solution keeps Analyzers as DRY (don't repeat yourself) as possible. A user shouldn't have to write 2 or 3 analyzer specs to check for different lifecycles if the same check can reasonably apply to all lifecycles.

There are architecture changes such as [this design proposal](design/proposal-standard-interface.md) which in implementing make enough changes to address this issue entirely as well. It may be that the solution is to prioritize other improvements which will by their nature fix this and several other issues as well.

xavpaice:


xavpaice:
TODO: consider how this affects preflights

The support bundle collection is driven in pkg/supportbundle/collect.go, preflight in pkg/preflight/collect.go.  Both run a loop over `collector.Collect(opts.ProgressChan)`, however we're only going to want to actually get pod logs from the support bundle, not the preflights (maybe?).

design/proposal-logging.md
```diff
@@ -0,0 +1,188 @@
...
+Add a channel to each collector (including `logs`), of a type where collectors can, towards the end of their run, send a slice to the channel specifying the pods that particular collector wants to collect logs from.
+
+The new pod log collection function can read that channel, and trigger collecting logs from any pods that have not already been collected.
+
+Alter the clusterResources collector:
+
+* add an option (default none) to collect pod logs from all pods/containers according to a selector rather than just error pods
+* add an option (boolean, default false) to collect logs from all pods/containers within the namespace(s) supplied, or all namespaces if no namespace is listed
+* add an option to override the default `MaxLines` (500) currently used in the logs collector
+* Add the current `unhealthyPods` slice, plus a slice of the pods found with the new selector option, plus a slice from other collectors (e.g. Ceph, logs collectors) to a new slice, and sort/dedupe it.
```
dexhorthy: This all sounds great -- any chance of seeing a draft of what a yaml snippet of this would look like in a SupportBundle yaml?

design/proposal-logging.md
```diff
@@ -0,0 +1,188 @@
...
+	close(podChan)
+	logsWg.Wait()
+}
+```
+
+## Limitations
+ 
+## Assumptions
+
+* All collectors can be run concurrently, none depend on other collectors
```
diamonwiggins: Everything makes sense in this proposal to me, but we may have some edge cases here. I know i've seen users in the past rely on the synchronous nature of how collectors run, and require one to run before the other. An example:

```yaml
    - exec:
        name: generate-output
        collectorName: generate-output
        namespace: '{{repl Namespace}}'
        selector:
          - app=some-selector
        command: ["sh"]
        args: ["-c", "cd /some/data/dir && mv `ls -t | head -n 1` latest"]
        timeout: 5s
    - copy:
        name: collect-output
        collectorName: collect-output
        selector:
          - app=some-selector
        namespace: '{{repl Namespace}}'
        containerPath: /some/data/dir/latest
```

I wouldn't doubt there are other use cases out there as well. May need to think about adding `depends-on` similar to what you see in CI pipelines.

design/proposal-logging.md
chris-sanders: I love this suggestion having a sample input can really help drive understanding of the change.

xavpaice:
I'm going to revisit the design proposal and split it into two:

* simple logging alterations, so that the goal here is met
* take the more complex refactor of log collection and concurrent collections to it's own, separate, proposal which likely has a different priority and different discussion/decision points.

design/proposal-logging.md
xavpaice: thanks - yeah that needs to go in here

xavpaice: I've pulled back from adding this to clusterResources, simply because it is already implemented in the logs collector.  We can, and should, make sure that this stores the logs that are relevant and that might mean adding 'all' at some point.  However, the logs collector currently stores the logs in a place where sbctl can't find them, which is addressed in #744

design/proposal-logging.md
xavpaice: re-written to address this (among other things).  Thanks for the feedback!

xavpaice:
This proposal has been updated to be more accurate to our needs now.

design/proposal-concurrent-collectors.md
```diff
@@ -0,0 +1,207 @@
...
+This concept requires
+
+* error handling (what if that collector isn't in the spec, or fails)
+* logic to order all the collectors that depend on others as there might be several in the line
+
+Introduce a new function(s) which does the work of pod log collection, storing those logs where `sbctl` can find them in `/cluster-resources/pods/logs/[namespace]/[pod]/[container].log`.  This should start in a goroutine, prior to any collectors being run.
+
+Switch all collectors to use goroutines to collect (concurrently).  This should be written such that collectors are split into separate lists, in order that any collectors with `dependsOn` are in a phase later than the collector it depends on.  Run each phase in turn, at this stage it may be acceptable to wait for the entire list of collectors in that phase to complete prior to starting the next phase.
+
+Add a channel to each collector (including `logs`), of a type where collectors can, towards the end of their run, send a slice to the channel specifying the pods that particular collector wants to collect logs from.  For the `logs` and `clusterResources` collectors, this replaces the actual pod log collection routine.
```
danj-replicated: I think this might be overcomplicating things, it might be simpler to initialise the collector goroutine with everything it needs up front and leave the channel for communicating back to the parent process.

banjoh: +1

I think having a single place where any collector can say "collect logs from this pod" is probably much simpler compared to sharing a channel to all collectors.

banjoh:
> I'm going to revisit the design proposal and split it into two:
> 
> * simple logging alterations, so that the goal here is met
> * take the more complex refactor of log collection and concurrent collections to it's own, separate, proposal which likely has a different priority and different discussion/decision points.

Also the `dependsOn` (including run phases) feature is non-trivial and will most likely need a separate proposal.

design/proposal-logging.md
xavpaice: From the docs: "The exec collector can be used to run a command in an existing pod and include the output in the collected data."  Surely if this is the case, we shouldn't need to run the copy after the exec?  I mean, if it needs to create a backup or something, rather than copy the file it shouldn't be writing to the pod at all, and sending the result to stdout.

xavpaice:
> Also the `dependsOn` (including run phases) feature is non-trivial and will most likely need a separate proposal.

Indeed it isn't simple, my concern is that we have identified cases where folks are relying on the ordering in yaml to provide sequencing, and are going to need some kind of replacement even though we do not guarantee the sequencing at this stage.

I note that we don't seem to promise the sequence of collectors anywhere in the docs, and the example that was mentioned involved an `exec` followed by `copy` which isn't a pattern we should support in any case - we should never write to a pod's storage when running a collector, ideally not even logs.

In which case, are we agreed that we can leave the `dependsOn` for a separate proposal, to be evaluated based on whether it's needed, after the collectors are run async?

We could, at this stage, split this into 3 now:
* some proposal around having a generic log collector that all others can use (obtaining the list of pods from the parent process via some means)
* a proposal to run the collectors in goroutines
* a proposal to add a `dependsOn` flag for collectors to depend on other collectors

xavpaice:
Closing this, have added #809, #810 and #811 which replace this PR

PR #743, opened 2022-09-27
test commit

drohnow:
testing.


drohnow:
Hi @xavpaice!  Please check out progress thus far.  I added context withTimeout to Logs() and configured to cancel when timeout expires.  Included mock-context code for POC; will not be merged into main.  

Want to discuss if we want to add additional context timeout for the "sub-processes" within the logs().

xavpaice:
Re how to structure the context/timeout.  I recall we talked about whether each log collection gets an individual timeout, i.e. 100 pods would add up to 100x the timeout, or if it's an overall timeout for the entire collection.  I think this becomes less of an issue if/when we change the logs collection to run concurrently, however sequentially it's a problem.  For this particular change, I think it would be preferable to have a 'per pod' timeout, which we could set to a time that is a fairly safe "if it is still going after that long we should call it failed and move on".  That time is more predictable.

The intent of the change here is to prevent Troubleshoot from hanging indefinitely because one pod log isn't returning.

Future changes should improve the timeout for large clusters with lots of pods if we trigger log collection concurrently, as the overall time for multiple pods would be similar to 1 pod.  I think given that this is in the roadmap, we can live with the tradeoff above till then.

PR #748, opened 2022-09-28
Update logs collector storage location

### LINKED ISSUE #57: sbctl not displaying pod logs
Support bundles from our app cannot display the pod's logs. Is this something that can be enabled? Are we also able to capture the logs of system pods as well?

```
ryan.knauer@RyanKnauer-Lenovo:~$ kubectl logs openebs-localpv-provisioner-5945d7879f-g44np -n openebs
2022/10/17 21:21:52 http: superfluous response.WriteHeader call from github.com/replicatedhq/sbctl/pkg/api.(*logRecord).WriteHeader (middleware.go:20)
                                                                                                                                                      log files not found in support-bundle.
/tmp/sbctl-1447890770/support-bundle/cluster-resources/pods/logs/openebs/openebs-localpv-provisioner-5945d7879f-g44np/openebs-provisioner-hostpath.log
/tmp/sbctl-1447890770/support-bundle/cluster-resources/pods/logs/openebs/openebs-localpv-provisioner-5945d7879f-g44np/openebs-provisioner-hostpath-logs-errors.log
```

### LINKED ISSUE #744: Logs collector stores logs where sbctl can't find them
**Bug Description**

**Expected Behavior**

**Steps To Reproduce**

**Additional Context**

- Troubleshoot version: 0.42

xavpaice:
Changes the storage location for the logs collector to be the same as the clusterResources collector, so that sbctl can find them.

This will alter the filepath for pod logs, and require analyzers that use pod logs to direct to the new path.


pkg/collect/logs.go
```diff
@@ -107,10 +101,7 @@ func savePodLogs(ctx context.Context, bundlePath string, client *kubernetes.Clie
 
 	setLogLimits(&podLogOpts, limits, convertMaxAgeToTime)
 
-	fileKey := fmt.Sprintf("%s/%s", name, pod.Name)
-	if container != "" {
-		fileKey = fmt.Sprintf("%s/%s/%s", name, pod.Name, container)
-	}
+	fileKey := fmt.Sprintf("%s/%s/%s/%s", "cluster-resources/pods/logs", pod.Namespace, pod.Name, container)
```
xavpaice, line 107: need to duplicate here, have one that is the 'old' path and one that's the new.  Then symlink the new one to the old one when it's saved.

edgarlanting: If I test this non-refactored change with `sbctl` we also miss the `*json` files located in `cluster-resources/`

I'm currently looking into adding the functionality for a symlink of the new location to the old location of the logs as described earlier and see if the other issue with missing `json` files is also resolved.

xavpaice:
Likely to be replaced by #807

PR #752, opened 2022-10-03
Add golangci-lint linter make target

### LINKED ISSUE #754: Add capability to run linters and commit hooks to the projects development flow
### Problem to solve

Perform quality gating checks on the developers machine before committing code into version control.

### Customer impact

Potentially reduce bugs in the software by catching lint errors earlier

### Developer experience impact
- Git pre-commit hooks will help detect and (optionally) autofix various issues (lint, formatting etc) before committing code
- Initially, it will take time to get used to the workflow for developers who have not used pre-commit hooks in their workflow. This can be a barrier to entry for people willing to make contributions. We can settle with making `pre-commit` optional to mitigate this.
- Initial runs of the linters and pre-commit hooks will expose many changes which will require time commitment to resolve.

### Proposed solution

Add the following two tools that will be run on developer machine.

- [golangci-lint ](https://golangci-lint.run/)- a tool that bundles [many well known linters](https://golangci-lint.run/usage/linters/) into a single binary. It also provides an [easy configuration mechanism](https://golangci-lint.run/usage/configuration/) where a project can select a few linters and pass in options that best work for the project. The defaults given should be sufficient for this project.
- [pre-commit](https://pre-commit.com/) - a tool that injects a [git pre-commit hook](https://git-scm.com/docs/githooks#_pre_commit) that gets run whenever someone commits changes. There are a many [pre-commit plugins](https://pre-commit.com/hooks.html) to pick from.

banjoh:
Add [golangci-lint](https://golangci-lint.run/) to the project - a tool that bundles [many well known linters](https://golangci-lint.run/usage/linters/) into a single binary. It also provides an [easy configuration mechanism](https://golangci-lint.run/usage/configuration/) where a project can select a few linters and pass in options that best work for the project. The defaults given should be sufficient for this project.


xavpaice:
There's some issues to discuss before merging a change like this, perhaps the next steps could be:

* Start an issue to discuss the change in terms of intent, and the minimal requirements to achieve that
* Discuss the approaches in the issue
* Once agreed, write up an ADR and store it in the Troubleshoot repo to record the outcome
* Then we would want to resolve the lint issues _prior_ to the pre-commit change, otherwise the pre-commit change would affect all changes in flight and between fixing the precommit and the lint fixes.

xavpaice (changes requested):
I would be happier with this change if we just do the addition to Makefile for now, but not include the pre-commit right away.  The current list of failures is a blocker.

banjoh:
> I would be happier with this change if we just do the addition to Makefile for now, but not include the pre-commit right away. The current list of failures is a blocker.

I removed pre-commit until we get agree to use it in the project

banjoh:
> There's some issues to discuss before merging a change like this, perhaps the next steps could be:
> 
> * Start an issue to discuss the change in terms of intent, and the minimal requirements to achieve that
> * Discuss the approaches in the issue
> * Once agreed, write up an ADR and store it in the Troubleshoot repo to record the outcome
> * Then we would want to resolve the lint issues _prior_ to the pre-commit change, otherwise the pre-commit change would affect all changes in flight and between fixing the precommit and the lint fixes.

I added https://github.com/replicatedhq/troubleshoot/issues/754 issue to the backlog for discussion

xavpaice (approved):
LGTM

PR #760, opened 2022-10-04
Discover Support Bundle and Redactor specs in cluster configmaps and secrets

### LINKED ISSUE #713: Add support to CLI to search for all configMaps that contain Troubleshoot redactors
## Request

Ref design #663, and task #699

Currently Troubleshoot can use configMaps for Redactor specs, and when called within KOTS we can see multiple redactors added to the defaults built in.  With #683 we add the ability to list multiple specs on the CLI, but not a wildcard, directory, or search pattern.

This request is to implement a search function whereby any configMaps that can be found with a matching label (default to  `troubleshoot.io/kind=redactor-spec`), will be collected and used as if they had been specified on the CLI.  This search should be made in the namespace defined in the same way as the kubectl client.

## In scope

* add a switch to the CLI, which triggers Troubleshoot to search for configMaps
  * in the namespace specified with `-n` or `--namespace`
  * all namespaces if  no namespace is specified
  * for any configMap with the label `troubleshoot.io/kind=redactor-spec` unless the `-l` switch is provided to override that
* Include each configMap redactor found as if it had been specified on the CLI (i.e. use them all)
* Test coverage
* Documentation

NOTE: the `-l` switch is already used in #699 and we should discuss if we want to share the switch or make a new one.  Same for `-n` though it's more likely this should be shared.

## Out of scope

* KOTS changes to make use of this addition

## NOTE

The package `"k8s.io/cli-runtime/pkg/genericclioptions"` may be helpful with the kubectl options, see [this example](https://github.com/kubernetes/sample-cli-plugin).

This task may be affected by work on #694, which uses the URI specified in a spec only on the specs provided on the CLI.  To be clear, the specs found in secrets with this task need to be considered as specs provided on the CLI, therefore the URI will need to be read and used.

Also: Note that KOTS uses the `kots.io/kind=supportbundle`label [here](https://github.com/replicatedhq/kots/blob/main/pkg/store/kotsstore/supportbundle_store.go#L239) to describe collected support bundles.

Kots does store a Troubleshoot spec, with the name `kotsadm-$APPSLUG-supportbundle`, however it does not currently have a label which can be used to identify it.  Also redactors are stored in configMaps kotsadm-redact, kotsadm-redact-spec and kotsadm-appslug-redact-spec amonst possibly other places.  These are not labeled other than with default kots labels.

## Definition of done

- [x] CLI switch enabling Troubleshoot to search for and use all configMaps with a label `troubleshoot.io/kind=redactor-spec` (default), or a CLI specified label
- [x] CLI supports kubectl's `-l, --selector=''` switch to override the label searched for
- [x] CLI supports kubectl's `-A`, `--all-namespaces`, `-n`, `--namespace` switches to override namespace
- [x] Unit tests for the new implementation
- [ ] Documentation updated to show usage of this new feature in https://troubleshoot.sh/docs

diamonwiggins:
Expanding the initial feature added in https://github.com/replicatedhq/troubleshoot/pull/709 to discover both support bundle and redactors specs across configmaps and secrets

* Removes the `selector` flag which complimented the `load-cluster-specs` flag. It's my opinion that it's okay for this to be a default as I can't really think of a use case where this needs to be changed by an end user. If we did want to add selectors back, then you would need flags for each type eg. SupportBundle, Redactor, etc.

* Based on the code today it made sense to maintain mutlidoc support for standalone redactors found in configmaps or secrets

* However, the multidoc support we have today is confusing. It expects the first document in the file to be a support bundle spec and anything after it can be a redactor. I'm pretty sure if you were to have a mutlidoc with two docs, the first being a redactor and the second being a support bundle, that it wouldn't work. Should we continue supporting this as is today? Should multidocs be able to recognize any order? Do we still even want multidoc support? Should we move to only expect a single type of object in each multidoc? Worth noting that much of this feature isn't documented today and it only impacts the CLIs in this project.


xavpaice:
I'd prefer that we make sure that existing yaml files with a supportBundle and Redactor type doc, in that order, still work after the change.  If we add support for different orders, that's great but not required.

I don't believe that any cluster stored specs will contain multidocs today - Redactors and SupportBundle specs have been stored in ConfigMaps and Secrets, separately, so there's been no need for multidoc there.

Regards removing `-l`, https://github.com/replicatedhq/troubleshoot/issues/699#issuecomment-1235761525 has some background as to why it was added.  I think @emosbaugh might have some thoughts on this and I'd appreciate the advice.

diamonwiggins:
The current plan after a bit of discussion in today's standup is to continue with the flexibility of being able to use the `-l` selector, but reuse the same label across both SupportBundle and Redactor so that we don't need to have multiple selectors per type. So the selector will be shared between them.

PR #763, opened 2022-10-07
Support for making TLS connection to redis in the redis collector

### LINKED ISSUE #746: Preflight checks for Redis fail after TLS is enabled
### Problem to solve
Preflight checks for Redis fail after TLS is enabled on the respective service. Currently, Preflight checks cannot be configured to utilize a CA bundle, certificate, or a private key to validate authentication. 

### Customer Impact
- Increased support burden with support tickets related to the failed preflight checks.
- Some customers refuse to upgrade to a later version due to preflight check failure 

### Describe possible workarounds
- Disable TLS for Redis (not feasible)
- Disable certificate checks entirely

### Potential solution
Add collector options to the Redis collectors:

- require server authentication,
- passing a custom CA bundle to deal with self signed ones,
- passing a certificate and a private key for client authentication purpose.

https://pkg.go.dev/github.com/go-redis/redis/v8#Options shows TLS config coming from the https://pkg.go.dev/crypto/tls#Config struct which is used in the Redis collector.

banjoh:


banjoh:
Work done in a different branch

PR #766, opened 2022-10-08
Supports the Kubernetes distribution analyzer identifying VMware Tanzu

crdant:
Adds a check to the Kubernetes distribution analyzer to identify VMware Tanzu using the same approach as identifying OpenShift. I started down the road of trying to identify Tanzu via the node after a discussion with @jdewinne, but there was nothing unique to a Tanzu cluster vs. any other cluster created by Cluster API. 

After seeing how OpenShift was identified, I was able to use a similar approach to identify Tanzu. There is an API service `run.tanzu.vmware.com` and a capabilities resource that will be found in a Tanzu cluster. I looked for the group `run.tanzu.vmware.com` in the same way the OpenShift check looks for `apps.openshift.io`.

Originally I implemented it as a separate check (so we had `CheckOpenShift` and `CheckTanzu`) but I generated it into one function `CheckApiResourcesForProviders` for readability.

### DIFF
--- a/pkg/analyze/distribution.go
+++ b/pkg/analyze/distribution.go
@@ -18,6 +18,7 @@ type providers struct {
 	gke           bool
 	digitalOcean  bool
 	openShift     bool
+	tanzu         bool
 	kurl          bool
 	aks           bool
 	ibm           bool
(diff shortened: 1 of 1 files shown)


crdant:
I see the CLA Assistant comment. Do I need to do that as a Replicated employee?

jdewinne:
> I see the CLA Assistant comment. Do I need to do that as a Replicated employee?

No. You also have to sign ✍️

crdant:
Thanks @jdewinne. In past companies my employment agreement covered contributing to their open source repos, so I wasn't sure.

xavpaice (approved):
LGTM, thanks for the contribution

PR #768, opened 2022-10-10
Ensure clusterResources is added prior to other collectors

### LINKED ISSUE #767: Analyzer result shows pods created by Troubleshoot as terminated
## Problem to solve

The support bundle screen in KOTS looks very busy since the metadata/errors about pods failing was added.

Furthermore, there are analyzers shown for pods that were run *by the support-bundle process*!

## Proposal

Any or all of:

- hide failed pods that are descendants of K8s Jobs
- hide failed pods that were created by troubleshoot (e.g. `run` collector)
- hide automatic failed pod analyzers entirely until we get something that looks good
- if a deployment or job is in a bad state, just show that and hide analyzers about its descendant replicasets and pods (or, hide analyzers about parent objects and just show the pod analyzers)

## Further details

A big contributor here is the Sentry-Provision-User Job usually has 3-5 pods fail while the database comes up, but the job eventually succeeds. This is a very common pattern in many apps, and is kind of what Jobs were designed for.

There have been months of iteration on this feature since SE raised concerns about the original change, and I feel like there's still not a lot of empathy for how this ends up looking when CE runs demos for prospects. This still looks really rough.  I am totally in favor of iteration here but it's becoming frustrating the number of SCs that have been written and how long this has been generally degraded.

An alternative would be to change/rearchitect the application all SEs demo with, but that doesn't help the second impact point -- if a prospect uses a Job that has similar behavior (and they are quite common in my experience) then they will see the same busy UI.

## Customer Impact (importance/urgency)

This impacts our ability to demo well, and our ability to show newer features. This affects our demo->PoV conversion rate.

This impacts the customers experience with the product -- I fear it will hurt our PoV->close rate as well.

## What does business success look like, and how can we measure that?

SEs are demoing with the latest version of kots instead of intentionally downgrading to avoid this busy view.

Higher PoV close rates

Less frequent feedback on lost deals that troubleshoot is "not that good" or that "we can build this ourselves easily".

## Links / references

Relates to: https://app.shortcut.com/replicated/story/38792/some-insights-have-null-as-the-level https://app.shortcut.com/replicated/story/38540/support-bundle-analyzer-errors https://app.shortcut.com/replicated/story/39739/analysis-overview-shows-errors-about-pods-in-other-namespaces

Shortcut story for this issue: https://app.shortcut.com/replicated/story/40888/support-bundle-screen-is-too-noisy-it-includes-pods-run-by-the-support-bundle-itself-starting-in-kots-1-59-1

xavpaice:
If Troubleshoot adds the clusterResources or ClusterInfo collector, ensure that this is added at the first element of the collectors array.

Fixes: #767

### DIFF
--- a/cmd/troubleshoot/cli/root.go
+++ b/cmd/troubleshoot/cli/root.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"github.com/go-logr/logr"
-	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"github.com/replicatedhq/troubleshoot/pkg/k8sutil"
 	"github.com/replicatedhq/troubleshoot/pkg/logger"
 	"github.com/spf13/cobra"
@@ -80,19 +79,6 @@ func InitAndExecute() {
 func initConfig() {
 	viper.SetEnvPrefix("TROUBLESHOOT")
 	viper.AutomaticEnv()
-}
-
-func ensureCollectorInList(list []*troubleshootv1beta2.Collect, collector troubleshootv1beta2.Collect) []*troubleshootv1beta2.Collect {
-	for _, inList := range list {
-		if collector.ClusterResources != nil && inList.ClusterResources != nil {
-			return list
-		}
-		if collector.ClusterInfo != nil && inList.ClusterInfo != nil {
-			return list
-		}
-	}
-
-	return append(list, &collector)
 }
 
 func writeFile(filename string, contents []byte) error {
--- a/pkg/preflight/collect.go
+++ b/pkg/preflight/collect.go
@@ -123,8 +123,9 @@ func CollectHost(opts CollectOpts, p *troubleshootv1beta2.HostPreflight) (Collec
 func Collect(opts CollectOpts, p *troubleshootv1beta2.Preflight) (CollectResult, error) {
 	collectSpecs := make([]*troubleshootv1beta2.Collect, 0, 0)
 	collectSpecs = append(collectSpecs, p.Spec.Collectors...)
-	collectSpecs = ensureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterInfo: &troubleshootv1beta2.ClusterInfo{}})
-	collectSpecs = ensureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterResources: &troubleshootv1beta2.ClusterResources{}})
+	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterInfo: &troubleshootv1beta2.ClusterInfo{}})
+	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterResources: &troubleshootv1beta2.ClusterResources{}})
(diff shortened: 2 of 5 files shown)


diamonwiggins:
@xavpaice just curious. did something happen where the ordering mattered here? Or is the concern that troubleshoot could completely fail on other collectors so we want to capture these as soon as possible?

diamonwiggins (dismissed):

xavpaice:
switched to draft:

* there's 3 different ensureCollectorInList functions
* suggestion was to make it so that we have a function that takes a list of collectors, finds the one we want, and moves it to the beginning, then use that resulting list to execute the collectors themselves.

cmd/troubleshoot/cli/root.go
```diff
@@ -82,19 +81,6 @@ func initConfig() {
 	viper.AutomaticEnv()
 }
 
-func ensureCollectorInList(list []*troubleshootv1beta2.Collect, collector troubleshootv1beta2.Collect) []*troubleshootv1beta2.Collect {
```
xavpaice, line 85: This function was unused

pkg/supportbundle/test/minimal.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: troubleshoot.sh/v1beta2
```
xavpaice: file is handy for testing, not essential in this change

pkg/preflight/collect.go
```diff
@@ -374,15 +375,16 @@ func ensureCollectorInList(list []*troubleshootv1beta2.Collect, collector troubl
...
 }
 
-func ensureHostCollectorInList(list []*troubleshootv1beta2.HostCollect, collector troubleshootv1beta2.HostCollect) []*troubleshootv1beta2.HostCollect {
-	for _, inList := range list {
-		if collector.CPU != nil && inList.CPU != nil {
-			return list
-		}
-		if collector.Memory != nil && inList.Memory != nil {
-			return list
+// collect ClusterResources earliest in the list so the pod list does not include pods started by collectors
```
xavpaice: would have preferred to add this function only once, in pkg/supportbundle, but till we combine the collect for both supportbundle and preflight, that refactor is beyond the scope of this change, so it's duplicated here (several other parts already are)

diamonwiggins: Any issue with putting it in `pkg/collect` and then making it public?

diamonwiggins:
The newly added function makes sense to me. I was wondering if it could benefit from being a more general _sort_ function that could take any collector as input, but i don't think I envision a use case for this outside of `clusterResources` 👍

pkg/preflight/collect.go
xavpaice: thanks, needed a push there :)

pkg/preflight/collect.go
```diff
@@ -360,29 +361,3 @@ func CollectRemote(opts CollectOpts, p *troubleshootv1beta2.HostPreflight) (Coll
...
-		}
-		if collector.ClusterInfo != nil && inList.ClusterInfo != nil {
-			return list
-		}
-	}
-
-	return append(list, &collector)
-}
-
-func ensureHostCollectorInList(list []*troubleshootv1beta2.HostCollect, collector troubleshootv1beta2.HostCollect) []*troubleshootv1beta2.HostCollect {
```
xavpaice, line 377: couldn't see any usage of this function, removed

diamonwiggins (approved):
lgtm

PR #769, opened 2022-10-10
make runPreflight and preflight cli flags public

stefanrepl:
This PR makes the `runPreflights` function publicly accessible and also includes an `AddFlags` implementation for `preflight` so we can easily utilize the same CLI flags in our helm plugin.

### DIFF
--- a/pkg/preflight/flags.go
+++ b/pkg/preflight/flags.go
@@ -1,0 +1,96 @@
+package preflight
+
+import (
+	flag "github.com/spf13/pflag"
+	utilpointer "k8s.io/utils/pointer"
+)
+
+const (
+	flagInteractive               = "interactive"
+	flagFormat                    = "format"
+	flagCollectorImage            = "collector-image"
+	flagCollectorPullPolicy       = "collector-pullpolicy"
+	flagCollectWithoutPermissions = "collect-without-permissions"
+	flagSelector                  = "selector"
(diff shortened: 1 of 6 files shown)


xavpaice:
I need to understand this a bit better - it looks from a short glance like we're setting up a new 'preflight' package which is able to be imported and consumed in a standard manner, by the CLI and by other applications importing it?

One issue we have is where other applications are importing and then we're more likely to make breaking changes in the future, does this change reduce the chances of that?

divolgin:
@xavpaice this is in preparation for making preflight checks from Helm plugins.  Command will look like this: `helm admin-console preflight ...`, and the rest of the command line will use standard preflight flags and will produce the same output that the standalone preflight binary would produce.

This is similar to how kots plugin for kubectl uses standard kubectl flags.

pkg/preflight/flags.go
```diff
@@ -0,0 +1,96 @@
...
+	CollectorPullPolicy       *string
+	CollectWithoutPermissions *bool
+	Selector                  *string
+	SinceTime                 *string
+	Since                     *string
+	Output                    *string
+	Debug                     *bool
+}
+
+var preflightFlags *PreflightFlags
```
divolgin, line 34: Would be great to not have globals, but this can be revisited.

divolgin (approved):

PR #770, opened 2022-10-11
update CODEOWNERS file

xavpaice:
Adds a new group to CODEOWNERS for global owner, and CREs for .md files.


chris-sanders (approved):
I didn't even think about using groups this is super clean, thanks!

PR #771, opened 2022-10-11
Make all Collector Fields Public

diamonwiggins:


adamancini (commented):
@diamonwiggins why did we need to make context public?  otherwise the change looks good,

diamonwiggins:
> @diamonwiggins why did we need to make context public? otherwise the change looks good,

it was an oversight initially. Anywhere you need to instantiate a collector to run, you'll need to provide the context at that time as well.

adamancini (approved):

PR #780, opened 2022-10-13
Add generic kubernetes resource analyzer

### LINKED ISSUE #764: make Kubernetes object field comparisons easier

While trying to write a sample spec that looks for a particular image version of a component I expect to have installed, I realized it was tricky to get the API description of a specific deployment.

I would like to be able to easily identify a specific object and get its equivalent `kubectl describe` or `kubectl get -o json|yaml` API response and write an analyzer that looks at specific fields.

Today we have `jsonCompare` which can take a path to a JSON field for comparison, but feeding the data into that analyzer is tricky.

Right now I'm considering using the data from `cluster-resources` but API descriptions of objects come from `kubectl describe <noun> -n <namespace>` and all of the deployments for a given namespace are listed in an `items[]` array.  It's not very elegant to have to figure out what integer index belongs to the deployment you want, and it's not very reproducible if the # of deployments changes.

**Describe the feature**

Add a collector to inspect a specific Kubernetes object, in a specific namespace, and return the complete description of that object in a file so any analyzer can use it.

- `runPod` with a kubectl client and `jq` to handle pulling out the data I need and `regex` analyzer
- add additional functionality to `deploymentStatus`, `replicasetStatus` etc. analyzers that can do field comparisons on something other than `replicas`

**Additional context**

https://github.com/replicatedhq/troubleshoot/issues/712

danj-replicated:
This is an attempt at solving a problem where we can't currently directly address resources collected by the cluster-resources collector.

example bundle spec:
```
kind: SupportBundle
metadata:
  name: collectors
spec:
  collectors:
    - cluster-resources: {}
  analyzers:
    - clusterResource:
        checkName: ekcko image version
        kind: Deployment
        namespace: kurl
        name: ekc-operator
        yamlPath: "spec.template.spec.containers.[0].image"
        value: replicated/ekco:v0.22.0
        outcomes:
          - fail:
              when: "false"
              message: The ekco deployment is not the latest version
          - pass:
              when: "true"
              message: the ekco deployment is up to date

```


danj-replicated:
TODO:

- [x] handle non-namespaced resources
- [x] add tests

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/pkg/errors"
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	iutils "github.com/replicatedhq/troubleshoot/pkg/interfaceutils"
+	"gopkg.in/yaml.v2"
+)
+
+func analyzeResource(analyzer *troubleshootv1beta2.ClusterResource, getFileContents func(string) (map[string][]byte, error)) (*AnalyzeResult, error) {
+
+	filemap := make(map[string]string)
```
xavpaice: could possibly make this more readable:
```
filemap := make(map[string]string){
    "Deployment": "deployments",
    etc
}
```

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	var selected interface{}
+	items, _ := iutils.GetAtPath(resource, "items")
+	itemslice := items.([]interface{})
+	for _, item := range itemslice {
+		name, _ := iutils.GetAtPath(item, "metadata.name")
+		if name == analyzer.Name {
+			selected = item
```
xavpaice: should we add a `break` here rather than continue looping?

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,140 @@
...
+		title = analyzer.CollectorName
+	}
+
+	result := &AnalyzeResult{
+		Title:   title,
+		IconKey: "kubernetes_text_analyze",
+		IconURI: "https://troubleshoot.sh/images/analyzer-icons/text-analyze.svg",
+	}
+
+	equal := reflect.DeepEqual(actual, expected)
```
xavpaice, line 113: presumably that means we can compare a yaml to an expected yaml, rather than just string compare.  Is there an option later to consider extracting parts to do things like semver comparison?  E.g. the example given in the PR is `value: replicated/ekco:v0.22.0` but really what we want is to split that out further and use semver to compare `v0.22.0`.  Later on, we could compare images running against kURL spec collected.

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,140 @@
...
+	var selected interface{}
+	items, _ := iutils.GetAtPath(resource, "items")
+	itemslice := items.([]interface{})
+	for _, item := range itemslice {
+		name, _ := iutils.GetAtPath(item, "metadata.name")
+		if name == analyzer.Name {
+			selected = item
+		}
+	}
+
```
xavpaice, line 90: I think it would be handy to split the "get data from this resource" part of the analyzer to a separate function that other analyzers could make use of, and have this particular analyzer just the first example of using it for a simple object comparison.  That "get data from the resource" part that abstracts the file location away from the analyzer would be a good first step to making an API that can be called for use with sbctl, analyzers, etc.

xavpaice (commented):
This is a great concept.  I've a few stylistic comments, take them or leave them, and some queries so I understand it better.

xavpaice:
https://github.com/replicatedhq/troubleshoot/pull/280/files has some useful things about semver comparisons, though we could have a separate analyzer that uses the same data extraction function to do semver comparison if you wanted.

xavpaice:
One limitation of this approach that we should be sure to document is that the `name` is the actual name of the resource itself, and for things like pods this is more dynamic than we might want.  Future additions might include a means to select by label (coping with multiple resources that might match) and/or using wildcards/regex.

pkg/analyze/kube_resource.go
danj-replicated: When splitting this out, would it be ok to leave it as a floating function (not associated with a specific analyzer) in t he analyze package? 

there's also considerations around parameters for this function. at the moment I'm having to pass through the GetFileContents function but I'd like to streamline the use of findResource() even if it means a slight duplication of code.

banjoh:
You might want to run `make schemas` to align the new analyser with troubleshoot CRDs. I believe [kots-lint](https://github.com/replicatedhq/kots-lint) depends in these schemas being correct.

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+	resourceLocation, ok := filemap[kind]
+
+	if ok != true {
+		return nil, errors.New("failed to find resource")
+	}
+
+	if namespace != "" {
+		datapath = filepath.Join("cluster-resources", resourceLocation, fmt.Sprintf("%s.json", namespace))
+	} else {
+		datapath = filepath.Join("cluster-resources", resourceLocation)
```
banjoh: I believe this ought to be `...resourceLocation, "*.json")` cause the file finding operation is a [glob pattern matcher](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/download.go#L272)

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+		"pv":                   "pvs.json",
+		"Node":                 "nodes.json",
+		"StorageClass":         "storage-classes.json",
+	}
+
+	var datapath string
+
+	resourceLocation, ok := filemap[kind]
+
+	if ok != true {
```
banjoh: The go way

```go
if !ok {
   ....
}
```

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/pkg/errors"
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	iutils "github.com/replicatedhq/troubleshoot/pkg/interfaceutils"
+	"gopkg.in/yaml.v2"
+)
+
+func FindResource(kind string, namespace string, name string, getFileContents func(string) (map[string][]byte, error)) (interface{}, error) {
+
+	filemap := map[string]string{
```
banjoh: I would have this map initialised once since it has constants

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+	"reflect"
+	"strconv"
+
+	"github.com/pkg/errors"
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	iutils "github.com/replicatedhq/troubleshoot/pkg/interfaceutils"
+	"gopkg.in/yaml.v2"
+)
+
+func FindResource(kind string, namespace string, name string, getFileContents func(string) (map[string][]byte, error)) (interface{}, error) {
```
banjoh: Please add some documentation of what this function is doing. Some questions worth addressing are

- Where are the k8s resources coming from (bundle or querying k8s apiserver?)
- What does the function give me back? The resource itself (which parts of it?), the name of the resource? Looking at the code I see that we return the full k8s object
- What if I have multiple objects with the same name in different namespaces? From the code it seems we get back the first one we find.

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+
+	itemslice := items.([]interface{})
+	for _, item := range itemslice {
+		name, err := iutils.GetAtPath(item, "metadata.name")
+		if err != nil {
+			return nil, errors.Wrapf(err, "Failed to find resource with name: %s", name)
+		}
+		if name == name {
+			selected = item
+			break
```
banjoh: Should we collect all resources found instead of picking the first one we find? The caller ought to decide what to do with the list of resources if they needed to find only one e.g ensure that a namespace is provided.

pkg/analyze/kube_resource_test.go
```diff
@@ -0,0 +1,68 @@
...
+			req := require.New(t)
+
+			getCollectedFileContents := func(dirName string) (map[string][]byte, error) {
+				files, err := filepath.Glob(filepath.Join("files/support-bundle", dirName))
+				if err != nil {
+					return nil, errors.Wrapf(err, "invalid glob %q", dirName)
+				}
+				fileArr := map[string][]byte{}
+				for _, filePath := range files {
+					bytes, err := ioutil.ReadFile(filePath)
```
banjoh: Use `os.ReadFile`. `ioutil` package functions are deprecated

pkg/analyze/kube_resource_test.go
```diff
@@ -0,0 +1,68 @@
...
+				CollectorName: "Check namespaced resource",
+				Kind:          "Node",
+				Name:          "repldev-marc",
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			req := require.New(t)
```
banjoh: Use `assert` package. `require` package [fails immediately an assertion is invalid](https://github.com/stretchr/testify#require-package) as opposed to checking the reset of the assertions and showing the rest of the failures. The package is also intended to check preconditions before starting a test e.g ensuring we have a DB connection etc.

pkg/analyze/kube_resource.go
danj-replicated: > What if I have multiple objects with the same name in different namespaces?

it shouldn't matter since to select a namespaced object you have to be explicit about the namespace

banjoh: > it shouldn't matter since to select a namespaced object you have to be explicit about the namespace

I could run an analyser without selecting a namespace.

```yaml
kind: SupportBundle
metadata:
  name: example
spec:
  collectors:
    - cluster-resources: {}
  analyzers:
    - clusterResource:
        checkName: Nginx image version
        kind: Pod
        name: nginx
        yamlPath: "spec.containers.[0].image"
        value: nginx:1.23
        outcomes:
            - fail:
                when: "false"
                message: The nginx version is old
            - pass:
                when: "true"
                message: Happy that you have the latest nginx version
```

I have two pods, one with an old version and the other with a new version

```sh
[evans] $ k get pods -A | grep nginx
default       nginx                                     1/1     Running     0                 4m23s       
kube-system   nginx                                     1/1     Running     0                 4m23s  
```

When I analyse a support bundle generated using the spec above I get a pass. The analyser picks the pod in the kube-system namespace (`nginx:1.23`). The one in the default is `nginx:1.23`

pkg/analyze/kube_resource.go
banjoh: If the intention is to always have a namespace specified, we'd need to remove `datapath = filepath.Join("cluster-resources", resourceLocation)` condition and return with an error if `namespace == ""`

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	files, err := getFileContents(datapath)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to read collected resources")
+	}
+
+	var resource interface{}
+	for _, file := range files {
+		err = yaml.Unmarshal(file, &resource)
```
banjoh: `resource` will always have the last successfully parsed objects. The rest will be ignored.

If there will always be one namespace, then the for-loop can be removed cause there should always be one file

pkg/analyze/kube_resource.go
banjoh: To run the analyser spec above I had to apply [this](https://github.com/replicatedhq/troubleshoot/pull/780#discussion_r1026310814) patch first, else an error was just being raised.

pkg/analyze/kube_resource.go
danj-replicated: The intention is that if a namespace isn't specified, we try and match a cluster-scoped resource.  the structure of a support bundle makes this a bit awkward as it dumps cluster scoped resources in the directory at the same level as the namespace folders.

diamonwiggins: nitpicking here, but you could skip the `else` and do:

```
        datapath = filepath.Join("cluster-resources", resourceLocation)
	if namespace != "" {
		datapath = filepath.Join("cluster-resources", resourceLocation, fmt.Sprintf("%s.json", namespace))
	}
```

diamonwiggins:
Took a look here, and I can come back and give a final review after Evan's comments has been addressed as I agree with what he's suggested also.

pkg/analyze/kube_resource.go
banjoh: > The intention is that if a namespace isn't specified, we try and match a cluster-scoped resource. the structure of a support bundle makes this a bit awkward as it dumps cluster scoped resources in the directory at the same level as the namespace folders.

I see. In that case I think we need to document it in the analyser documentation. It'll catch people off guard a lot.

pkg/analyze/kube_resource.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return selected, nil
+
+}
+
+func analyzeResource(analyzer *troubleshootv1beta2.ClusterResource, getFileContents func(string) (map[string][]byte, error)) (*AnalyzeResult, error) {
+
+	selected, err := FindResource(analyzer.Kind, analyzer.Namespace, analyzer.Name, getFileContents)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to find resource")
```
banjoh, line 88: Shouldn't we treat not finding a resource business as usual and have analysis fail and reported in `AnalyzeResult.Message`? Resources can come and go. Errors bubbled up don't get reported very well at the moment.

My cli gave me this
```sh
[evans] $ ./bin/support-bundle examples/support-bundle/sample-supportbundle.yaml

 Collecting support bundle ⠴ cluster-info2022/11/24 17:25:15 An analyzer failed to run: failed to find resource: failed to read collected resources: read "/var/folders/19/bp6c9chj0sgcpcxmxxl69s040000gn/T/supportbundle1967949124/support-bundle-2022-11-24T17_25_10/cluster-resources/pods": read /var/folders/19/bp6c9chj0sgcpcxmxxl69s040000gn/T/supportbundle1967949124/support-bundle-2022-11-24T17_25_10/cluster-resources/pods: is a directory
support-bundle-2022-11-24T17_25_10.tar.gz
```

Perhaps differentiate not finding a resource by checking  `selected == nil` and handle resource-not-found in `FindResource` where you return runtime errors as `error`s and if a resource is not found, just return `nil, nil`

pkg/analyze/kube_resource.go
danj-replicated: The problem with removing the loop here is that `getFileContents` as used elsewhere in the code returns a `map[string][]byte` and I don't know what the `string` will be

banjoh: I think what you want here is to store all unmarshalled resources into a list and search for your selected item from all resources
```go
	resources := []interface{}{}
	for _, file := range files {
		... unmarshal resource and store
		resources = append(resources, resource)
	}

	for _, resource := range resources {
		items, err := iutils.GetAtPath(resource, "items")
		...

		itemslice := items.([]interface{})
		for _, item := range itemslice {
			...
			if name == name {
				return item, nil
			}
		}
	}

    # We did not find any item using the provided selector
    return nil, nil
```
Something like above

banjoh (dismissed):
LGTM

pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
```diff
@@ -58,6 +58,18 @@ type DeploymentStatus struct {
...
+type ClusterResource struct {
+	AnalyzeMeta   `json:",inline" yaml:",inline"`
+	CollectorName string     `json:"collectorName,omitempty" yaml:"collectorName,omitempty"`
+	Outcomes      []*Outcome `json:"outcomes" yaml:"outcomes"`
+	Kind          string     `json:"kind" yaml:"kind"`
+	ClusterScoped bool       `json:"clusterScoped" yaml:"clusterScoped"`
+	Namespace     string     `json:"namespace,omitempty" yaml:"namespace,omitempty"`
+	Name          string     `json:"name" yaml:"name"`
+	YamlPath      string     `json:"yamlPath" yaml:"yamlPath"`
+	Value         string     `json:"value,omitempty" yaml:"value,omitempty"`
```
diamonwiggins: I wonder if this would be clearer if it was `expectedValue` instead of `value`. Makes it more intentional in the analyzer code what you're testing/looking for.

diamonwiggins (dismissed):

pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
danj-replicated: Yeh that's a point.

banjoh (approved):

diamonwiggins (approved):

PR #782, opened 2022-10-14
Remove unused Preflight spec

diamonwiggins:


xavpaice:
do we know what this file was doing there in the first place?  There's some others that might also need removal - preflight_local.yaml, preflight_remote.yaml, sample-troubleshoot.yaml, collect_local.yaml, collect_remote.yaml.

diamonwiggins:
@xavpaice removed a couple more. `sample-troubleshoot.yaml` i believe is used or referenced in docs somewhere, but the other that were contributed with the "Remote Host Collector" work can be scrapped.

xavpaice (approved):
Thanks!  LGTM

PR #797, opened 2022-10-21
Support the use of Helm templates and Helm's values file

divolgin:
This reverts a change made in order to support kots as helm plugin

The new change allows the use of Helm templates and Helm's values files in troubleshoot specs.  For example

```
kubectl preflight preflight.yaml --values values.yaml
```


chris-sanders (changes requested):
Wow this is a big change. Is there a design proposal or something for this? I feel like we need to review what we're doing and why. It's not immediately obvious to me why we would do this and what alternatives we considered.

xavpaice:
This introduces a useful, but significant feature.  As such, it's going to need some documentation added to Troubleshoot.sh in order to able to be considered 'supported', but I have no objection to adding the feature right now.

How about we introduce a feature flag, e.g. environment variable `TROUBLESHOOT_ENABLE_TEMPLATE` that, if set to true, enables the feature right away.  Once the docs are merged, we will simply add a PR that removes the feature flag logic.  This could be added with just an `if` wrapping `cmd.Flags().StringSlice...` in the cli package.

What we're trying to avoid here is support requests coming in for the feature when it's not clear how to use it, what the limitations are, and what its intended purpose is.  As such, the docs change doesn't need to be very large, but does want to describe:

* what can be used in templates and what can not - e.g. if we don't support logic such as `if`, but do support using the `Values` from Helm, we should state that so it's clear
* the expected locations of values files
* how the values are merged if there are duplicates
* how failure is handled should a value be missing from values.yaml
* a simple example (the one included with tests in this change might be adequate)

divolgin:
@xavpaice 

> How about we introduce a feature flag, e.g. environment variable `TROUBLESHOOT_ENABLE_TEMPLATE`...

Cobra supports hidden flags and commands.  [This](https://github.com/replicatedhq/kots/blob/0f23d8d1d97c9340140067068d047b992a4e8362/cmd/kots/cli/install.go#L506-L509) is the typical way to hide Alpha and undocumented features.  Will that work?

xavpaice:
> 



> Cobra supports hidden flags and commands. [This](https://github.com/replicatedhq/kots/blob/0f23d8d1d97c9340140067068d047b992a4e8362/cmd/kots/cli/install.go#L506-L509) is the typical way to hide Alpha and undocumented features. Will that work?

Perfect!

chris-sanders (changes requested):
Can you add a link to the PR for troubleshoot.sh so we can track the docs that were agreed would be done with this before we land the change?

xavpaice:
Do we _have_ to have the revert commit in this PR?  I think it was a valuable change even though it's no longer needed by one particular project.  It would be ideal to have each PR for a separate purpose in any case.

pkg/helm/templates_test.go
```diff
@@ -0,0 +1,57 @@
...
+//go:embed testdata/spec-before.yaml
+var specBeforeTemplating string
+
+//go:embed testdata/spec-after.yaml
+var specAfterTemplating string
+
+//go:embed testdata/values.yaml
+var valuesData string
+
+func Test_ApplyTemplates(t *testing.T) {
```
xavpaice, line 30: let's expand this test to cover malformed templates, and missing values in values.yaml (because those are the ones we'll see in production the most)

xavpaice (changes requested):
I'd like to extend the test coverage to handle failures gracefully, plus split out the PR to a single change rather than the rollback and values.yaml file combined.

pkg/helm/templates_test.go
divolgin: updated

xavpaice:
@divolgin it looks like this change is blocked on a couple of things:

- documentation (preferably over hiding the feature behind a flag, though we've agreed that's an option temporarily)
- not rolling back the change that moves preflight to a package (because we want that, and would like to extend it)

How can we progress this one?

xavpaice:
This concept is being reconsidered, but will be a different PR.  I'll close this one now, since it's not going to be merged, but will be used as reference.

PR #804, opened 2022-10-26
Update cluster_resources.go

edgarlanting:
Due to deprecation of the `v1beta1` API at `policy/v1beta1` for `PodDisruptionBudgets` and `batch/v1beta1` for `CronJobs` I've added this PR to change the `client.Batch<API-version>` and `client.Policy<API-version` to future proof these two functions.

This has been tested against multiple Kubernetes versions starting from 1.25.x and multiple Troubleshoot versions (that contain PodDisruptionBudgets and CronJobs)

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -414,7 +414,7 @@ func getPodDisruptionBudgets(ctx context.Context, client *kubernetes.Clientset, 
 	errorsByNamespace := make(map[string]string)
 
 	for _, namespace := range namespaces {
-		PodDisruptionBudgets, err := client.PolicyV1beta1().PodDisruptionBudgets(namespace).List(ctx, metav1.ListOptions{})
+		PodDisruptionBudgets, err := client.PolicyV1().PodDisruptionBudgets(namespace).List(ctx, metav1.ListOptions{})
 		if err != nil {
 			errorsByNamespace[namespace] = err.Error()
 			continue
@@ -617,7 +617,7 @@ func cronJobs(ctx context.Context, client *kubernetes.Clientset, namespaces []st
 	errorsByNamespace := make(map[string]string)
 
 	for _, namespace := range namespaces {
-		nsCronJobs, err := client.BatchV1beta1().CronJobs(namespace).List(ctx, metav1.ListOptions{})
+		nsCronJobs, err := client.BatchV1().CronJobs(namespace).List(ctx, metav1.ListOptions{})
 		if err != nil {
 			errorsByNamespace[namespace] = err.Error()
 			continue
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -142,7 +142,7 @@ func (c *CollectClusterResources) Collect(progressChan chan<- interface{}) (Coll
 		}
 	}
 
-	// pod disruption budgets (for all existing namespaces)
+	// pod disruption budgets
 
 	PodDisruptionBudgets, pdbError := getPodDisruptionBudgets(ctx, client, namespaceNames)
 	for k, v := range PodDisruptionBudgets {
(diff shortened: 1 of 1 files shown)


banjoh (dismissed):

diamonwiggins:
@edgarlanting for backwards compatibility should we have a function like https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/cluster_resources.go#L650 here as well so that we can do some discovery on which apiVersion should be used?

edgarlanting:
> @edgarlanting for backwards compatibility should we have a function like https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/cluster_resources.go#L650 here as well so that we can do some discovery on which apiVersion should be used?

Hmm That would be a very good thing @diamonwiggins thanks for mentioning this

pkg/collect/cluster_resources.go
```diff
@@ -617,7 +626,16 @@ func cronJobs(ctx context.Context, client *kubernetes.Clientset, namespaces []st
 	errorsByNamespace := make(map[string]string)
 
 	for _, namespace := range namespaces {
-		nsCronJobs, err := client.BatchV1beta1().CronJobs(namespace).List(ctx, metav1.ListOptions{})
+		ok, err := discovery.HasResource(client, "batch/v1", "CronJob")
+		if err != nil {
+			errorsByNamespace[namespace] = err.Error()
+			continue
+		}
+		if !ok {
```
banjoh: This also needs to handle the old behaviour

```go
if err != nil {
			
if ok {
	nsCronJobs, err := client.BatchV1().CronJobs(namespace).List(ctx, metav1.ListOptions{})
} else {
	nsCronJobs, err := client.BatchV1beta1().CronJobs(namespace).List(ctx, metav1.ListOptions{})
}
```

pkg/collect/cluster_resources.go
```diff
@@ -414,7 +414,16 @@ func getPodDisruptionBudgets(ctx context.Context, client *kubernetes.Clientset,
 	errorsByNamespace := make(map[string]string)
 
 	for _, namespace := range namespaces {
-		PodDisruptionBudgets, err := client.PolicyV1beta1().PodDisruptionBudgets(namespace).List(ctx, metav1.ListOptions{})
+		PodDisruptionBudgets, err := client.PolicyV1().PodDisruptionBudgets(namespace).List(ctx, metav1.ListOptions{})
```
banjoh, line 430: I believe this should be like so

```go
		ok, err := discovery.HasResource(client, "policy/v1", "PodDisruptionBudgets")
		if err != nil {
			return nil, map[string]string{"": err.Error()}
		}
		if ok {
			// This is a v1 API call
			PodDisruptionBudgets, err = client.PolicyV1().PodDisruptionBudgets(namespace).List(ctx, metav1.ListOptions{})
		} else {
			PodDisruptionBudgets, err = client.PolicyV1beta1().PodDisruptionBudgets(namespace).List(ctx, metav1.ListOptions{})
		}
```

edgarlanting: I see I pushed an erroneous/insufficient version of cluster_resources.go @banjoh fixing this atm.

pkg/collect/cluster_resources.go
```diff
@@ -613,29 +667,76 @@ func jobs(ctx context.Context, client *kubernetes.Clientset, namespaces []string
...
+	}
+
+	return cronJobsV1beta(ctx, client, namespaces)
+}
+
+func cronJobsV1(ctx context.Context, client *kubernetes.Clientset, namespaces []string) (map[string][]byte, map[string]string) {
+	cronJobsByNamespace := make(map[string][]byte)
+	errorsByNamespace := make(map[string]string)
+
+	for _, namespace := range namespaces {
```
banjoh, line 687: I wonder if we can DRY this code a bit. If its non-trivial then just add a TODO for to have it looked at as tech-debt

edgarlanting: I've added a TODO comment for function `pdbV1`.

diamonwiggins (approved):
lgtm

PR #807, opened 2022-10-28
Update logs.go

### LINKED ISSUE #57: sbctl not displaying pod logs

### LINKED ISSUE #744: Logs collector stores logs where sbctl can't find them

edgarlanting:
Patch logs.go to use a symbolic link in order to let SBCTL access logs via `kubectl logs <pod-name> ...`

While this is not yet as simple as it may seem, this draft PR will allow us to further discuss/look at how to implement this.

### DIFF
--- a/pkg/collect/logs.go
+++ b/pkg/collect/logs.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"strings"
 	"time"
+	"os"
 
 	"github.com/pkg/errors"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
--- a/pkg/collect/logs.go
+++ b/pkg/collect/logs.go
@@ -184,10 +184,10 @@ func savePodLogs(ctx context.Context, bundlePath string, client *kubernetes.Clie
 	}
 	
 	// EL - Temporarily moved defer to this location for further testing
-	defer podLogs.Close()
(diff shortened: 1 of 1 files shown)


xavpaice:
in order to have the tests check that the symlink works, you could swap things around and have the new, sbctl compatible, path be the location of the actual file, and the 'old' path be a symlink to it.  That way, the functional test which confirms the logs are collected will check the symlink is there.

PR #809, opened 2022-10-31
Add proposal for concurrent collection

xavpaice:
Design proposal for running collectors concurrently instead of sequentially.

### DIFF
--- a/design/proposal-concurrent-collectors.md
+++ b/design/proposal-concurrent-collectors.md
@@ -4,7 +4,6 @@
 
 Increase the speed at which collectors complete.
 
-Ensure that the cluster-resources collector runs prior to any others, so that things like pod status and pod logs only include the pods from the application rather than the ones that Troubleshoot started as part of collectors.
 ## Non Goals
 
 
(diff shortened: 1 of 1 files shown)


design/proposal-concurrent-collectors.md
```diff
@@ -0,0 +1,55 @@
...
+
+Ensure that the cluster-resources collector runs prior to any others, so that things like pod status and pod logs only include the pods from the application rather than the ones that Troubleshoot started as part of collectors.
+## Non Goals
+
+
+## Background
+
+Currently all collectors run sequentially, in the order in which they are specified in the spec.  This is implied, and has not been guaranteed.  With the addition of the ability to supply multiple specs, the order is no longer predicatble.
+
+Although sequence is not guaranteed for any collectors, #768 ensures that the `clusterResources` collector runs first, so that the list of resources does not include those that are started during the Troubleshoot collection sequence.
```
diamonwiggins, line 14: There has traditionally been usage of collectors in the past that relied on them being sequential. Do we intend to scrap those use cases with this change? An example:

```
    - exec:
        name: get-latest-prometheus-snapshot
        collectorName: get-latest-prometheus-snapshot
        namespace: some-namespace
        selector:
          - app=some-app
        command: ["sh"]
        args: ["-c", "cd /prometheus/data/snapshots && mv `ls -t | head -n 1` latest"]
        timeout: 5s
    - copy:
        name: collect-prometheus-snapshot
        collectorName: collect-prometheus-snapshot
        selector:
          - app=some-app
        namespace: some-namespace
        containerPath: /prometheus/data/snapshots/latest
```

The copy collector relies on the exec collector preceding it. To continue supporting this after collectors run concurrently, we'd likely need some sort of `depends_on` functionality

diamonwiggins (commented):
the comment i left is my only feedback

diamonwiggins:
This would also imply that certain types of collectors would need to be written for uniqueness. For example:

```
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - runPod:
        name: "run-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
    - runPod:
        name: "run-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
            args: ["-w", "5", "www.reddit.com"]
```

This works today, but would cause a collision after this change.

design/proposal-concurrent-collectors.md
```diff
@@ -0,0 +1,55 @@
...
+
+TODO: review host and remote collectors
+
+## Limitations
+
+Relies on splitting up the list of collectors into `clusterResources` and others.
+
+## Assumptions
+
+* The Kubernetes API can handle the load of collectors running concurrently, including when the pod logs collector runs
```
diamonwiggins: Wonder if for this reason this should be optional in some way? You could have a property in the spec to enable it, but that wouldn't work well with how we combine specs today.

chris-sanders: I think this is the type of thing we need make a reasonable decision on and not try to pass through to the user to figure out. Appropriate throttling should be something we can make a reasonable decision on and it's unlikely to be something and end user has sufficient information to make any informed decision on.

design/proposal-concurrent-collectors.md
chris-sanders: This is an interesting use case, I'm curious if it's the correct way to encourage people to do this sort of operation. I think a depends_on might be useful, but I'm also concerned that the example given above isn't idempotent. Each run creates and replaces the 'latest' file in the pod and doesn't clean it up when it's done!

For this example, would it be better to have a collector that gathers that latest snapshot w/o creating artifacts in the pod and returns that? I'm wondering if the use case for dependent collectors is the best way to address issues or if it's an indication we could use some more/better collectors?

chris-sanders (commented):
Comments in-line

design/proposal-concurrent-collectors.md
xavpaice: There's no guarantee of sequencing, and even less so when we have multiple specs merged.  The more we merge specs, the less we can predict the sequence.

I'm keener, for this use case and others like it, to write a new collector that would, e.g. run a command and process the result somehow.  For this example, would we not pipe the result to stdout which is collected in any case?

I did write up a proposal for a depends-on type of field but that is a complex subject which would need some extra business justification for the effort involved in writing and maintaining it.

design/proposal-concurrent-collectors.md
xavpaice: does the k8s API not throttle incoming requests properly?  If that's an issue we would probably want to consider how we resolve it for other instances of high query load.

Maybe we have a concurrency limit that is a reasonable default, optionally some config for it?

xavpaice:
> This would also imply that certain types of collectors would need to be written for uniqueness. For example:
> 
> ```
> kind: SupportBundle
> metadata:
>   name: sample
> spec:
>   collectors:
>     - runPod:
>         name: "run-ping"
>         namespace: default
>         podSpec: 
>           containers:
>           - name: run-ping
>             image: busybox:1
>             command: ["ping"]
>             args: ["-w", "5", "www.google.com"]
>     - runPod:
>         name: "run-ping"
>         namespace: default
>         podSpec: 
>           containers:
>           - name: run-ping
>             image: busybox:1
>             command: ["ping"]
>             args: ["-w", "5", "www.reddit.com"]
> ```
> 
> This works today, but would cause a collision after this change.

Where would the collision sit - in writing the result output to the bundle or somewhere else?  Either sequential or concurrent, they would overwrite the same data.

diamonwiggins:
> > This would also imply that certain types of collectors would need to be written for uniqueness. For example:
> > ```
> > apiVersion: troubleshoot.sh/v1beta2
> > kind: SupportBundle
> > metadata:
> >   name: sample
> > spec:
> >   collectors:
> >     - runPod:
> >         name: "run-ping"
> >         namespace: default
> >         podSpec: 
> >           containers:
> >           - name: run-ping
> >             image: busybox:1
> >             command: ["ping"]
> >             args: ["-w", "5", "www.google.com"]
> >     - runPod:
> >         name: "run-ping"
> >         namespace: default
> >         podSpec: 
> >           containers:
> >           - name: run-ping
> >             image: busybox:1
> >             command: ["ping"]
> >             args: ["-w", "5", "www.reddit.com"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This works today, but would cause a collision after this change.
> 
> Where would the collision sit - in writing the result output to the bundle or somewhere else? Either sequential or concurrent, they would overwrite the same data.

The collision here would actually be in creating the Pod. Because we don't use a controller like `Deployment` to create the Pod, the `name` would be the same since both collectors have the same pod name. We could address that individually by using `generateName` in the collector though. I'm not sure of what other similiar scenarios there could be. 

Beyond that, i'm sure we'd also want to control concurrent access to files. The actual collection portion could be concurrent, but saving to the file would likely need to remain sequential.

Merging/deduplication may eliminate many scenarios, but not sure its something we'd want to rely on.

xavpaice:
> The collision here would actually be in creating the Pod.

We could probably overcome that with unique pod names, but the concurrent file access would still be an issue.

However, if the collectors run sequentially, we still only get a result from the last one to run - so it's no _worse_ at least.

I think we could add a section to this doc to consider how we fix that but I also think we can consider moving that to an independent issue.

design/proposal-concurrent-collectors.md
xavpaice: I think the depends-on feature would need to be a separate design doc, it's not something we have right now and only works currently by accident.  The fact that we now merge specs makes it less predictable and I'd discourage anyone from relying on the order of collectors.

diamonwiggins (approved):

PR #810, opened 2022-10-31
Add proposal to consolidate collection code

xavpaice:
Proposal to reduce duplication of code, and implement a more stable, public, collection API.


diamonwiggins (approved):
lgtm

banjoh:
Since this proposal will lead to quite a bit of refactoring, I wonder whether we should adopt using [internal packages](https://go.dev/doc/go1.4#internalpackages) so as to have a clear separation of truly public APIs (ones that can be imported externally) and internal public APIs (ones that are used within troubleshoot project). This would make it easier to know and manage which public APIs can accept breaking changes and which ones can't.

banjoh:
The proposal LGTM as is though

banjoh (approved):

PR #811, opened 2022-10-31
Add proposal for common pod log collector

xavpaice:
Adds a proposal to centralize the pod log collection function so that multiple collectors may use that, concurrently.

### DIFF
--- a/design/proposal-generic-log.md
+++ b/design/proposal-generic-log.md
@@ -143,7 +143,7 @@ No changes required.
 ## Alternatives Considered
 
 * Having each collector append the list of pods to a global, then on completion of all collectors run the collector for pod logs (starting with deduplicating the list of pods).  This would be slower since we need to wait for all collectors to complete prior to pod log collection starting.
-* Making a generic logs collector function that can be imported by any collector that wishes to collect logs.  This would not allow the same pod's logs to be collected several times.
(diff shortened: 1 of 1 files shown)


design/proposal-generic-log.md
```diff
@@ -0,0 +1,150 @@
...
+The existing Collect functions will need unit tests altered (or added).
+
+## Documentation
+
+No changes required.
+
+## Alternatives Considered
+
+* Having each collector append the list of pods to a global, then on completion of all collectors run the collector for pod logs (starting with deduplicating the list of pods).  This would be slower since we need to wait for all collectors to complete prior to pod log collection starting.
+* Making a generic logs collector function that can be imported by any collector that wishes to collect logs.  This would not allow the same pod's logs to be collected several times.
```
diamonwiggins: Not sure I understand `This would not allow the same pod's logs to be collected several times.`.

xavpaice: need to express it better in the text - the idea is that the logs collection function keeps a list of pods it has already collected, and ignores requests for pods it's already grabbed.

xavpaice: I've updated the statement, but I'm not sure when looking at these alternatives - maybe a common log collector function that others import is a simpler option rather than messing with channels etc, and a better option therefore.

diamonwiggins: I think there could be value in the channels approach if it makes the barrier to entry for contributing/maintaining collectors easier. Not sure how big of a difference it would be to import a function that saves pod logs to a bundle vs. sending a pod name into a channel.

xavpaice:
parking this as we have found other priorities that would be better use of time.  Recent changes to logging have improved performance and availability.

PR #814, opened 2022-10-31
Add omitempty to StorageClassName in schema

### LINKED ISSUE #813: Lint errors with storageClassName
**Bug Description**

When we lint an analyzer/preflight spec that uses the following, we get an error stating that we need to specify storageClassName:

```yaml
spec:
  analyzers:
    - storageClass: {}
```

**Expected Behavior**

No errors, since the docs state that not specifying anything in the analyzer has suitable defaults.

**Steps To Reproduce**

Include an analyzer in a preflight spec that includes `- storageClass: {}`, or at least does not include `storageClassName`.

**Additional Context**

xavpaice:
Allow a StorageClass spec to not require specify storageClassName.

Includes the associated schema update, which pulls in a number of other updates.

Fixes: #813

### DIFF
--- a/pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
+++ b/pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
@@ -12,7 +12,7 @@ type ClusterVersion struct {
 type StorageClass struct {
 	AnalyzeMeta      `json:",inline" yaml:",inline"`
 	Outcomes         []*Outcome `json:"outcomes" yaml:"outcomes"`
-	StorageClassName string     `json:"storageClassName" yaml:"storageClassName"`
+	StorageClassName string     `json:"storageClassName,omitempty" yaml:"storageClassName,omitempty"`
 }
 
 type CustomResourceDefinition struct {
--- a/config/crds/troubleshoot.replicated.com_analyzers.yaml
+++ b/config/crds/troubleshoot.replicated.com_analyzers.yaml
@@ -687,7 +687,6 @@ spec:
                           type: string
                       required:
                       - outcomes
-                      - storageClassName
                       type: object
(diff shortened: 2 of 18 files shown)


pkg/apis/troubleshoot/v1beta2/analyzer_shared.go
```diff
@@ -12,7 +12,7 @@ type ClusterVersion struct {
 type StorageClass struct {
 	AnalyzeMeta      `json:",inline" yaml:",inline"`
 	Outcomes         []*Outcome `json:"outcomes" yaml:"outcomes"`
-	StorageClassName string     `json:"storageClassName" yaml:"storageClassName"`
+	StorageClassName string     `json:"storageClassName,omitempty" yaml:"storageClassName,omitempty"`
```
erquhart, line 15: @xavpaice I'm guessing by the code path that this only applies for api version `troubleshoot.replicated.com/v1beta2`, but https://github.com/replicated-collab/cased-kots/issues/25 is using `v1beta1`. Does this still apply in that case?

xavpaice: thanks, I'm not sure we should be making any changes to v1beta1, isn't it there because of history and needing to be (eventually) deprecated?

xavpaice: I don't see a big problem with such a minor update - ran it anyway.

erquhart: Really good point, I honestly don't know.

adamancini (approved):

diamonwiggins (approved):
lgtm

PR #818, opened 2022-11-01
chore: Add vscode launch configurations to allow debugging

### LINKED ISSUE #817: Add vscode launch configurations to allow debugging troubleshoot applications

To increase ease of debugging troubleshoot applications for vscode users, add launch configurations to the repository so contributors do not have to keep recreating them.

Reference on debugging using vscode: https://code.visualstudio.com/docs/editor/debugging

banjoh:
With these launch configurations, you should be able to run or debug applications (supportbundle, analyze or preflight) from within vscode. The configurations assume vsode's workspace is the root of your repo so you need to launch vscode from your repo.

More on vscode debugging: https://code.visualstudio.com/docs/editor/debugging

Fixes: #817

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,8 +30,9 @@ vendor
 dist
 try.sh
 
-.vscode/
+__debug_bin
 workspace.*
+*.tar.gz
 
 cosign.key
 sbom/
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,43 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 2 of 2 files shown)


xavpaice:
My concern here is that previously we excluded .vscode in order to allow folks to have their own setup for the IDE - I'm not entirely confident we actually want this checked into the repo, however I suspect it would be a good addition and lower the entry barrier to people being able to get productive with this codebase.

banjoh:
> My concern here is that previously we excluded .vscode in order to allow folks to have their own setup for the IDE - I'm not entirely confident we actually want this checked into the repo, however I suspect it would be a good addition and lower the entry barrier to people being able to get productive with this codebase.

Overwriting contributor's own configuration might be annoying, perhaps outweighing the value of the configuration. I think we should avoid it and share such knowledge/configs around.

PR #821, opened 2022-11-03
feat(collectors): Store all pod logs in cluster-resources directory

### LINKED ISSUE #744: Logs collector stores logs where sbctl can't find them

banjoh:

All pod logs collected by the logs collector will now be stored in `/cluster-resources/pods/logs/[namespace]/[pod]/[container].log`. This will provide consistency and allow `sbctl` to find the logs when we run `kubectl logs <pod>`. To allow backwards compatibility, symlinks of the log files will be created in the current expected locations.

Closes: https://github.com/replicatedhq/troubleshoot/issues/744

## Checklist

- [ ] Yes
- [x] No


cmd/troubleshoot/cli/root.go
```diff
@@ -37,6 +39,14 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 		},
 	}
 
+	// Consume -v flag to set log level instead of using "debug"
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use
+	// TODO: Needs some further testing to ensure it works as expected
+	// TODO: Add more log-level aware logs in the project
+	fs := goflags.NewFlagSet("", goflags.PanicOnError)
```
banjoh: I added this here to test it out. I needed my logs that I can turn off by adding a log level :)

pkg/collect/logs.go
```diff
@@ -132,15 +137,63 @@ func savePodLogs(ctx context.Context, bundlePath string, client *kubernetes.Clie
...
-	setLogLimits(&podLogOpts, limits, convertMaxAgeToTime)
-
 	fileKey := fmt.Sprintf("%s/%s", name, pod.Name)
 	if container != "" {
 		fileKey = fmt.Sprintf("%s/%s/%s", name, pod.Name, container)
 	}
 
 	result := NewResult()
 
+	// Create symlinks for the pod name and container name
```
banjoh: This bit needs some TLC. DRY me please

pkg/collect/redact.go
```diff
@@ -33,6 +33,16 @@ func RedactResult(bundlePath string, input CollectorResult, additionalRedactors
 			reader = bytes.NewBuffer(v)
 		}
 
+		info, err := os.Lstat(filepath.Join(bundlePath, k))
+		if err != nil {
+			return errors.Wrap(err, "failed to stat file")
+		}
+
+		// Do not redact symbolic links. The target file of the link will be redacted instead.
```
banjoh: When redacting, the file gets opened which converts the symlink to a regular file. This is why you saw empty files in your implementation @xavpaice

pkg/collect/logs_test.go
```diff
@@ -77,3 +79,62 @@ func Test_setLogLimits(t *testing.T) {
...
 }
+
+func Test_savePodLogs(t *testing.T) {
+	tests := []struct {
+		name              string
+		withContainerName bool
+		want              CollectorResult
+	}{
+		{
+			name:              "with container name",
```
camilamacedo86, line 91: Could we improve the description such as should fail ... or should pass when ... should obtained the logs when ...

pkg/supportbundle/supportbundle.go
```diff
@@ -53,11 +53,12 @@ func CollectSupportBundleFromSpec(spec *troubleshootv1beta2.SupportBundleSpec, a
 		return nil, errors.New("did not receive collector progress chan")
 	}
 
-	tmpDir, err := ioutil.TempDir("", "supportbundle")
+	tmpDir, err := os.MkdirTemp("", "supportbundle")
 	if err != nil {
 		return nil, errors.Wrap(err, "create temp dir")
 	}
 	defer os.RemoveAll(tmpDir)
+	klog.V(2).Infof("Support bundle temp dir: %s", tmpDir)
```
camilamacedo86:
```suggestion
	klog.V(2).Infof("Support bundle created in temporary directory: %s", tmpDir)
```

wdyt?

Makefile
```diff
@@ -138,7 +138,7 @@ run-preflight: preflight
 
 .PHONY: run-troubleshoot
 run-troubleshoot: support-bundle
-	./bin/support-bundle ./examples/support-bundle/sample-supportbundle.yaml
+	./bin/support-bundle ./examples/support-bundle/sample-supportbundle.yaml --debug --interactive=false
```
camilamacedo86: Have we the same options for preflight?
If so, should the changes no be applied in both?

banjoh: > Have we the same options for preflight? If so, should the changes no be applied in both?

Actually this change should not have ended up here. I'll remove it. Thanks for noticing it

WRT to CLI flags and options, we'll have https://github.com/replicatedhq/troubleshoot/issues/824 address changes to be introduced.

pkg/collect/logs_test.go
banjoh: The BDD styly naming of the tests does not feel quite right here IMHO

```
=== RUN   Test_savePodLogs
=== RUN   Test_savePodLogs/should_have_a_container_name_in_path_when_container_name_is_defined
=== RUN   Test_savePodLogs/without_container_name
--- PASS: Test_savePodLogs (0.00s)
```

pkg/collect/collector.go
```diff
@@ -67,7 +67,7 @@ func GetCollector(collector *troubleshootv1beta2.Collect, bundlePath string, nam
 	case collector.ConfigMap != nil:
 		return &CollectConfigMap{collector.ConfigMap, bundlePath, namespace, clientConfig, client, ctx, RBACErrors}, true
 	case collector.Logs != nil:
-		return &CollectLogs{collector.Logs, bundlePath, namespace, clientConfig, client, ctx, sinceTime, RBACErrors}, true
+		return &CollectLogs{collector.Logs, bundlePath, namespace, client, ctx, sinceTime, RBACErrors}, true
```
camilamacedo86: All collectors are passing both.
Which I agree that is strange why I would pass the client and the config?

Besides I go along with the change I am favor to have standards always.
So, what about it be done in a follow up and for all case(s) instead of change only this one?
On top of that, would have some reason for we are getting a new client for each case? 

Anyway, I think would be very beneficial a follow up here where we would chose by default one of the both. I mean, we pass the config and get a new client always or we use the same client for all case scenarios. So that we can discuss if has a reason to create a new client.  

See https://pkg.go.dev/k8s.io/client-go/kubernetes#NewForConfig so we might need to create a new Client for each case because of rate-limiter .

pkg/collect/logs.go
```diff
@@ -36,11 +35,6 @@ func (c *CollectLogs) IsExcluded() (bool, error) {
 }
 
 func (c *CollectLogs) Collect(progressChan chan<- interface{}) (CollectorResult, error) {
-	client, err := kubernetes.NewForConfig(c.ClientConfig)
```
camilamacedo86: See: https://github.com/replicatedhq/troubleshoot/pull/821/files#r1026258904

pkg/collect/logs_test.go
camilamacedo86: Hun, that is a nit and not a blocker at all.

However, 

`=== RUN   Test_savePodLogs/without_container_name` Should pass or fail? What is the expected behaviour?

About BDD style, see: https://dev.to/smyrman/test-with-expect-a-bdd-style-go-naming-pattern-5eh5#:~:text=GWT%2C%20or%20%22Given%2DWhen,you%20start%20the%20detailed%20implementation.

Example
```
TestSum/With_non-empty_int_slice/Expect_no_error
TestSum/With_non-empty_int_slice/Expect_correct_sum
TestSum/With_non-empty_int_slice/Expect_input_is_unchanged
TestSum/With_empty_int_slice/Expect_no_error
TestSum/With_empty_int_slice/Expect_zero
TestSum/With_empty_int_slice/Expect_input_is_unchanged
```

So, should we not expected something like:

```
=== RUN   Test_savePodLogs/When_container_name_is_defined_expected_name_in_the_path
=== RUN   Test_savePodLogs/When_container_name_is_undefined_expected_no_error 
--- PASS: Test_savePodLogs (0.00s)
```

Makefile
camilamacedo86: But do we not need to ensure that when we change support bundle we also go along with the changes into the preflight and the other way around?  All share all logs is not also valid for preflights?

pkg/collect/logs.go
```diff
@@ -13,17 +14,15 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 )
 
 type CollectLogs struct {
-	Collector    *troubleshootv1beta2.Logs
-	BundlePath   string
-	Namespace    string
-	ClientConfig *rest.Config
```
camilamacedo86: See:  https://github.com/replicatedhq/troubleshoot/pull/821/files#r1026258904

ihmo we need to revert this one and create a follow up to discuss if we should pass the client or the config to create a new client for each collector. My 2 cents here is that we need create a new client for each collector for not bitten for the rate-limiter.

Anyway, the change need to be made horizontally ( same for all collectors ).

banjoh: At the moment the the logs collector is one of the the few creating its own client. The reason I removed it is to ensure the code is testable. This way we can pass in a client fake to create mocks.

pkg/collect/collector.go
banjoh: I'm for creating fewer clients (one at best) that have the required rate limiting configurations (QPS, Burst, HTTPClient etc) rather than creating many clients. Trying to circumvent rate limiting by having many clients usually ends up failing in the long run cause if the API server is struggling, it will respond with errors.

In general (no just in k8s) I think, in the sprit of having well behaving clients to a server, an application ought to create one client session (can have connection pools) to a server at a time. This way operational strategies such as rate limiting, session revocation/renewal, session tracking, logging, metrics etc work well and as expected.

banjoh: > Anyway, I think would be very beneficial a follow up here where we would chose by default one of the both. I mean, we pass the config and get a new client always or we use the same client for all case scenarios. So that we can discuss if has a reason to create a new client.

I agree. This needs to be looked at holistically in a separate issue/proposal

banjoh: We also need to remember that creating support bundles usually takes place in production systems. We'd rather get rate limited and handle it accordingly (retry, error out) rather than put more pressure on the systems.

pkg/collect/logs.go
camilamacedo86: Hi @banjoh,

So, I think you just discover why all collector func receive both; :-)  the client and the config:
-> Client : to help on the tests 
-> config : for the real code use the config to get a new client

pkg/collect/logs.go
```diff
@@ -70,10 +64,7 @@ func (c *CollectLogs) Collect(progressChan chan<- interface{}) (CollectorResult,
 				}
 
 				for _, containerName := range containerNames {
-					if len(containerNames) == 1 {
-						containerName = "" // if there was only one container, use the old behavior of not including the container name in the path
-					}
```
camilamacedo86, line 75: This part I would try to ping out its author or someone that has more history context to ensure that I would not be introducing an undesired behaviour.

examples/support-bundle/e2e.yaml
```diff
@@ -11,10 +11,10 @@ spec:
         data: "5"
     - runPod:
         collectorName: "static-hi"
-        podSpec: 
+        podSpec:
           containers:
           - name: static-hi
-            image: alpine:3.5
+            image: alpine:latest
```
camilamacedo86: I do not think that is a good idea use latest
That can result in the tests starts to fail and we do not know why.

To ensure that the test are idempotent the best is to pinned the versions

pkg/collect/logs.go
banjoh: > This part I would try to ping out its author or someone that has more history context to ensure that I would not be introducing an undesired behaviour.

I did so already and they said removing the code should be OK, but thanks for noting it.

examples/support-bundle/e2e.yaml
banjoh: > I do not think that is a good idea use latest That can result in the tests starts to fail and we do not know why.
> 
> To ensure that the test are idempotent the best is to pinned the versions

Changed to `alpine:3`

pkg/collect/collector.go
camilamacedo86: Hi @banjoh,

> I agree. This needs to be looked at holistically in a separate issue/proposal

I think we both are in the same page here that remove  `clientConfig` as arg and stop to create a newClient from the config would better fit in a new PR. Right? So, I'd say let's to discuss the motivations there. I am not sure If I can follow up your argumentations (sorry). I am afraid that too much comments make harder for others follow up and help on the review. 

So, could you revert the `clientConfig` changes and push it in another pr?  WDYT?

Makefile
banjoh: > 

Yes. All the CLIs will all be made consistent

pkg/collect/collector.go
banjoh: > So, could you revert the clientConfig changes and push it in another pr? WDYT?

It makes sense. Change done

pkg/collect/logs.go
```diff
@@ -168,11 +202,13 @@ func savePodLogs(ctx context.Context, bundlePath string, client *kubernetes.Clie
 	}
 	defer podLogs.Close()
 
-	prevLogWriter, err := result.GetWriter(bundlePath, fileKey+"-previous.log")
+	prevLogWriter, err := result.GetWriter(bundlePath, filePathPrefix+"-previous.log")
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get previous log writer")
 	}
-	defer result.CloseWriter(bundlePath, fileKey+"-previous.log", logWriter)
+	// NOTE: deffered calls are executed in LIFO order i.e called in reverse order
```
xavpaice: nit: spelling

pkg/collect/logs.go
```diff
@@ -148,11 +180,13 @@ func savePodLogs(ctx context.Context, bundlePath string, client *kubernetes.Clie
 	}
 	defer podLogs.Close()
 
-	logWriter, err := result.GetWriter(bundlePath, fileKey+".log")
+	logWriter, err := result.GetWriter(bundlePath, filePathPrefix+".log")
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get log writer")
 	}
-	defer result.CloseWriter(bundlePath, fileKey+".log", logWriter)
+	// NOTE: deffered calls are executed in LIFO order i.e called in reverse order
```
xavpaice: nit: spelling

pkg/collect/result.go
```diff
@@ -18,13 +18,72 @@ func NewResult() CollectorResult {
...
+	}
+
+	// Do nothing if the symlink already exists
+	_, err = os.Lstat(linkPath)
+	if err == nil {
+		return nil
+	}
+
+	// Create the symlink
+	err = os.Symlink(filePath, linkPath)
```
xavpaice, line 70: Question: does this create a link that has the entire directory path when it's tar'd or does it have the appropriate `../../foo/bar` type of relative path?

xavpaice (commented):
This looks great - couple of nits and a question which may be answered later on in the code.

I'd like a CRE to review this as well before we merge.

pkg/collect/logs.go
banjoh: I knew there was a double something somewhere! :)

pkg/collect/result.go
banjoh: > 

It does not. That is handled when creating the tar archive [here](https://github.com/banjoh/troubleshoot/blob/em/sbctl-pod-logs-744/pkg/collect/result.go#L256). We also do not want to do that here cause the view of the structure needs to remain intact for use cases such as running preflights where the collected data is only used in memory only and never archived.

banjoh: I added a code comment

diamonwiggins (approved):

diamonwiggins:
lgtm. needs a squash on the commits and not clear if you still have outstanding input from Camila to address.

xavpaice (approved):

PR #827, opened 2022-11-07
Discover Redactors in Cluster

diamonwiggins:
Discovers redactors in both secrets and configmaps with the `load-cluster-specs` flag


diamonwiggins:
This is ready for review. Before, there was some discussion on if we should change the new function I added to return a support bundle or redactor object as opposed to the `string` or `byte` contents, but I think that's a potential larger change that would be scope creep here. Behavior with this PR is that when `load-cluster-specs` is used it will search across both Configmaps and Secrets for SupportBundles or Redactors. It distinguishes between the actual content in a configmap/secret based on how the `key` is named. `support-bundle-spec` or `redactor-spec`.

danj-replicated (dismissed):
lgtm

cmd/troubleshoot/cli/run.go
```diff
@@ -138,34 +134,71 @@ func runTroubleshoot(v *viper.Viper, arg []string) error {
...
 
-		if bundlesFromSecrets != nil {
-			for _, bundle := range bundlesFromSecrets {
+		bundlesFromConfigMaps, err := specs.LoadFromConfigMapMatchingLabel(client, parsedSelector.String(), namespace, specs.SupportBundleKey)
+		if err != nil {
+			logger.Printf("failed to load support bundle spec from secrets: %s", err)
+		}
+		bundlesFromCluster = append(bundlesFromCluster, bundlesFromConfigMaps...)
+
+		if bundlesFromCluster != nil {
```
divolgin: There's no need for this check.  Loops handle nil arrays gracefully.

divolgin (dismissed):

xavpaice (approved):
LGTM, comments from others have been addressed.

PR #832, opened 2022-11-09
Deduplication for Cluster Resources Collector

diamonwiggins:

Add deduplication for Cluster Resources collector to prevent the same collector configuration from running twice and also to ensure if no namespace is provided on any one cluster resource definition, that all namespaces are collected by default

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/supportbundle/collect.go
+++ b/pkg/supportbundle/collect.go
@@ -18,6 +18,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type ProcessedCollector struct {
+	Name        string
+	IsMergeable bool
+}
+
 func runHostCollectors(hostCollectors []*troubleshootv1beta2.HostCollect, additionalRedactors *troubleshootv1beta2.Redactor, bundlePath string, opts SupportBundleCreateOpts) (collect.CollectorResult, error) {
 	collectSpecs := make([]*troubleshootv1beta2.HostCollect, 0, 0)
 	collectSpecs = append(collectSpecs, hostCollectors...)
@@ -73,8 +78,6 @@ func runCollectors(collectors []*troubleshootv1beta2.Collect, additionalRedactor
 	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterResources: &troubleshootv1beta2.ClusterResources{}})
 	collectSpecs = collect.EnsureClusterResourcesFirst(collectSpecs)
 
-	var allCollectors []collect.Collector
-
 	allCollectedData := make(map[string][]byte)
 
 	k8sClient, err := kubernetes.NewForConfig(opts.KubernetesRestConfig)
@@ -82,41 +85,39 @@ func runCollectors(collectors []*troubleshootv1beta2.Collect, additionalRedactor
 		return nil, errors.Wrap(err, "failed to instantiate Kubernetes client")
 	}
 
+	newDesiredCollectors := make(map[interface{}][]collect.Collector)
+
 	for _, desiredCollector := range collectSpecs {
-		if collectorInterface, ok := collect.GetCollector(desiredCollector, bundlePath, opts.Namespace, opts.KubernetesRestConfig, k8sClient, opts.SinceTime); ok {
+		if collectorInterface, cType, ok := collect.GetCollector(desiredCollector, bundlePath, opts.Namespace, opts.KubernetesRestConfig, k8sClient, opts.SinceTime); ok {
 			if collector, ok := collectorInterface.(collect.Collector); ok {
 				err := collector.CheckRBAC(context.Background(), collector, desiredCollector, opts.KubernetesRestConfig, opts.Namespace)
 				if err != nil {
 					return nil, errors.Wrap(err, "failed to check RBAC for collectors")
 				}
-				allCollectors = append(allCollectors, collector)
-				/*if mergeCollector, ok := collectorInterface.(collect.MergeableCollector); ok {
-					allCollectors, err = mergeCollector.Merge(allCollectors)
-					if err != nil {
-						msg := fmt.Sprintf("failed to merge collector: %s: %s", collector.Title(), err)
-						opts.CollectorProgressCallback(opts.ProgressChan, msg)
-					}
-				} else {
-					allCollectors = append(allCollectors, collector)
-				}*/
-			}
-		}
-	}
-	for _, collec := range allCollectors {
-		if mergeCollector, ok := collec.(collect.MergeableCollector); ok {
-			allCollectors, err = mergeCollector.Merge(allCollectors)
-			if err != nil {
-				msg := fmt.Sprintf("failed to merge collector: %s: %s", mergeCollector.Title(), err)
-				opts.CollectorProgressCallback(opts.ProgressChan, msg)
+				newDesiredCollectors[cType] = append(newDesiredCollectors[cType], collector)
 			}
 		}
 	}
 
-	foundForbidden := false
-	for _, c := range allCollectors {
-		for _, e := range c.GetRBACErrors() {
-			foundForbidden = true
-			opts.ProgressChan <- e
+	var foundForbidden bool
+	for k, v := range newDesiredCollectors {
+		// run merge on v if mergeable
+		if mergeCollector, ok := v[0].(collect.MergeableCollector); ok {
+			fmt.Println(newDesiredCollectors[k])
+			newList, err := mergeCollector.Merge(v)
+			if err != nil {
+				msg := fmt.Sprintf("failed to merge collector: %s: %s", mergeCollector.Title(), err)
+				opts.CollectorProgressCallback(opts.ProgressChan, msg)
+			}
+			newDesiredCollectors[k] = newList
+			fmt.Println(newDesiredCollectors[k])
+		}
+		foundForbidden = false
+		for _, daCollec := range v {
+			for _, e := range daCollec.GetRBACErrors() {
+				foundForbidden = true
+				opts.ProgressChan <- e
+			}
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


diamonwiggins:
A couple things to discuss here:

1. I found that attempting to do the merge while the resultant list of collectors structs is being built from the user provided spec would cause the resultant list to potentially need to be modified by the `merge` function while it's still being constructed. That's the way it [was mocked out initially](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go#L78-L104) as we thought it would be easier to do the merges as we were building the resultant list to reduce the amount of comparisons we'd need to make. For example:

```
kind: SupportBundle
metadata:
  name: bundle1
spec:
  collectors:
    - clusterResources:
        namespaces: ["default"]
    - clusterResources:
        namespaces: ["kube-system"]
    - clusterInfo:
        collectorName: my-cluster-info
    - http:
        name: healthz
        get:
          url: http://api:3000/healthz
    - data:
        collectorName: my-password-dump
        name: data
        data: |
          my super secret password is abc123
          another redaction will go here
    - clusterResources: {}
```

As we're iterating through this spec, the first cluster resources definition is added to the `"toBeRun"` collectors, and so is the second, but its only until later when we've reached the final cluster resources definition, that we actually want to modify the resultant collectors to remove all the others that were specified earlier which feels strange.

Instead what i've implemented in my PR, is that the merge function expects to be passed not the resultant collector list as it's being built, but the final list of collectors `"toBeRun"`.

2. I think the problem with how i've implemented this may not be the right approach because it makes each collector's merge function responsible for potentially rebuilding the final `"toBeRun"` collector list in its entirety and not just modifying the part of the collector list its responsible for. Meaning the input to each collector's merge function should probably be a full list of all its **own** definitions from the user provided spec. 

`func (c *CollectClusterResources) Merge(allCollectors []Collector) ([]Collector, error)` 

becomes something like...

`func (c *CollectClusterResources) Merge(allclusterResourcesCollectors []CollectClusterResources) ([]CollectClusterResources, error)`

It would require a bit of a refactor, but perhaps this is a better approach as we look to have collectors built more closely to a plugin model which would imply a limited scope of responsibility for its merge function.

There's always the chance i'm over-complicating all this(very likely), and would love to hear what others think about what the best solution would be here :)

banjoh:
This is definitely an improvement to modifying the list of collectors in place. It still does feel a bit strange having an API where a `Collector` instance merges into itself. How about having an `AddSpec` API instead? This removes the notion of "merging". Instead, what we do is have each collector accept multiple spec objects and merge them. Your function signature would look like so

```go
func (c *CollectClusterResources) AddSpec(spec *troubleshootv1beta2.Collects) error
```

This way we do not have to manipulate the collector instances array. I think it also becomes easier to reason with the code

With the above approach, we would need to have one instance of each collector type. The `var allCollectors []collect.Collector` slice would contain only one collector typo instance. If we fancy, we could have each collector be a singleton, but that might be a bigger refactor (_I think_)

diamonwiggins:
@banjoh If i'm understanding you correctly it sounds akin to what i'm suggesting in my second point, but doing the "merge"/"add specs" before we build the result list of collector struct objects?

Also can you clarify what you mean here. Not sure I follow:

> With the above approach, we would need to have one instance of each collector type. The var allCollectors []collect.Collector slice would contain only one collector typo instance. If we fancy, we could have each collector be a singleton, but that might be a bigger refactor (I think)

diamonwiggins:
Agreed that `func (c *CollectClusterResources) AddSpec(spec *troubleshootv1beta2.Collects) error` is a better example of what I was trying to explain before. However, thinking through this more I like the idea of...

1. Build map `map[CollectorType][]troubleshootv1beta2.Collect` from the `[]*troubleshootv1beta2.Collect` provided
2. For each unique `CollectorType` run [GetCollector](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go#L86)
3. `GetCollector` would be modified to accept `[]*troubleshootv1beta2.Collect` and return `[]interface{}`
4. In `GetCollector` we can run the merge/add specs function which would return `[]interface{}` returning `[]collect.Collector` with one or more collectors. https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go#L85-L104 would then look more like

```
var allCollectors []collect.Collector
collectorMap := map[collectorType][]troubleshootv1beta2.Collect

// build map from user provided `[]*troubleshootv1beta2.Collect` here

for _, desiredCollectorList := range collectorMap {
	result := collect.GetCollector(desiredCollectorList, bundlePath, opts.Namespace, opts.KubernetesRestConfig, k8sClient, opts.SinceTime)
	allCollecters = append(allCollectors, result...)
}
```

This is probably overly complicated as well 😅 , but I like the idea of providing the full list of `desiredCollectors` to be run for the specific collector type so that `GetCollector`, which would include `Merge/AddSpecs`, can return everything at once.

banjoh:
> @banjoh If i'm understanding you correctly it sounds akin to what i'm suggesting in my second point, but doing the "merge"/"add specs" before we build the result list of collector struct objects?

Yes, its basically what you are suggesting. The only difference here is doing away with the logic of maintaining a list of collector struct objects. Instead, we'd have one collector object that manages all spec objects of its type within itself. Your https://github.com/replicatedhq/troubleshoot/pull/832#issuecomment-1309250937 echoes what I had in mind

banjoh:
> Also can you clarify what you mean here. Not sure I follow:

I was suggesting using a [singleton design pattern](https://en.wikipedia.org/wiki/Singleton_pattern) which seems to fit in here quite well.

This is how a typical collector would look like
```go
type CollectLogs struct {
	fields...
}

var singleton *CollectLogs

// GetCollectLogsInstance lazily creates a CollectLogs singleton instance. This singleton
// will be the only instance created throughout the lifetime of the application
// NOTE: Clients can create their own instances using CollectLogs{} if they choose to
func GetCollectLogsInstance(params...) {
	if singleton == nil {
		singleton = &CollectLogs{params...}
	}

	return singleton
}

// Implementation of the Collector interface
```

And this is how the `GetCollector` function would look like

```go
func GetCollector(params...) collect.Collector {
	switch {
	case collector.Logs != nil:
		// Return a singleton instance i.e the same object every time
		return GetCollectLogsInstance(params...)
	}
	...	// other collectors
}
```

and finally how it fits together
```go
// A map of unique collector instances. Since the collector instances are
// singletons, we will only ever have one instance of each collector type.
var allCollectors map[collect.Collector]interface{}{}

for _, collectSpec := range collectSpecs {
	// This function call will always return the same collector object of a specific collector type
	collector := collect.GetCollector(params...)
	collector.AddSpec(collectSpec)

	// Store the collector in a map so we can avoid duplicates
	allCollectors[collector] = nil
}

// Later on in the code
for collector := range allCollectors {
	// Check RBAC permissions
	collector.GetRBACErrors()

	// Collect data
	collector.Collect(opts.ProgressChan)
}
```

I might be overcomplicating things as well :D

banjoh:
As a side note, should we change the [GetCollector](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/collector.go#L54) return type to `collect.Collector`? It looks like it will always return a `collect.Collector`.

diamonwiggins:
I've held off on the larger change discussed that would have involved modifying the collector structs to take a complete list of all of its specs vs. just a single spec at a time. The further I got into the implementation, it was very clear that would be a significant refactor.

pkg/supportbundle/collect.go
```diff
@@ -73,70 +74,74 @@ func runCollectors(collectors []*troubleshootv1beta2.Collect, additionalRedactor
...
 	}
 
-	foundForbidden := false
-	for _, c := range allCollectors {
-		for _, e := range c.GetRBACErrors() {
-			foundForbidden = true
-			opts.ProgressChan <- e
+	var foundForbidden bool
+	for collectorType, collectors := range newDesiredCollectors {
+		if mergeCollector, ok := collectors[0].(collect.MergeableCollector); ok {
```
diamonwiggins, line 102: not in love with this, but i'm not sure of a better way to accomplish what I want. I would rather that the key in the `newDesiredCollectors` map could have a value that I could type check on, but I wasn't able to find a good way to do that.

pkg/supportbundle/collect.go
```diff
@@ -73,70 +74,74 @@ func runCollectors(collectors []*troubleshootv1beta2.Collect, additionalRedactor
...
+			newList, err := mergeCollector.Merge(collectors)
+			if err != nil {
+				msg := fmt.Sprintf("failed to merge collector: %s: %s", mergeCollector.Title(), err)
+				opts.CollectorProgressCallback(opts.ProgressChan, msg)
+			}
+			newDesiredCollectors[collectorType] = newList
```
banjoh: Isn't the aim to have one collector after a merge?

diamonwiggins: That proved to be a much larger refactor of the code considering the Collect and CheckRBAC functions also expect that you pass one collector to it at a time.

banjoh: Cool

I'm happy with this compromise approach. Only thing I would ask is to add some comments letting anyone know that this is a stop gap towards a future simpler implementation, however it would look like.

diamonwiggins:
@emosbaugh would love to get your thoughts on my approach before i write some tests

emosbaugh:
Seems like a pragmatic approach that was well thought through based on the comments. I think you should move forward and write the tests.

I wonder if it would be worth it to do a generic merge first for collectors that are equivalent? I imagine this may solve merge for many of the collector types.

diamonwiggins:
Converting back to draft. Forgot about preflight 🤦

emosbaugh (approved):

PR #834, opened 2022-11-14
fix(flag): fix wrong output filename

### LINKED ISSUE #825: RFE: meaningful default name to save the preflights checks

After run `kubectl preflight https://preflight.replicated.com` and choose the s option to save the default output seems to be "o". Could we change that for for example preflight_checks.txt or  preflight_checks_<timestamp>.txt? 

Screenshot 2022-11-07 at 10 39 23

DexterYan:
- fix the issue of  default output to be "o"
- add go test for flag

Fixes: #825

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/preflight/flags_test.go
+++ b/pkg/preflight/flags_test.go
@@ -19,42 +19,37 @@ func TestAddFlagsString(t *testing.T) {
 		want:    "",
 		wantErr: true,
 	}, {
-		name:    "expect output=empty, err=nil when output flag is set",
-		flag:    "output",
-		want:    "",
-		wantErr: false,
-	}, {
 		name:    "expect format=human, err=nil when format flag is set",
 		flag:    "format",
 		want:    "human",
 		wantErr: false,
 	}, {
-		name:    "expect collector-image=empty, err=nil when format collector-image is set",
+		name:    "expect collector-image=empty, err=nil when collector-image flag is set",
 		flag:    "collector-image",
 		want:    "",
 		wantErr: false,
 	}, {
-		name:    "expect collector-pullpolicy=empty, err=nil when format collector-pullpolicy is set",
+		name:    "expect collector-pullpolicy=empty, err=nil when collector-pullpolicy flag is set",
 		flag:    "collector-pullpolicy",
 		want:    "",
 		wantErr: false,
 	}, {
-		name:    "expect selector=empty, err=nil when format selector is set",
+		name:    "expect selector=empty, err=nil when selector flag is set",
 		flag:    "selector",
 		want:    "",
 		wantErr: false,
 	}, {
-		name:    "expect since-time=empty, err=nil when format since-time is set",
+		name:    "expect since-time=empty, err=nil when since-time flag is set",
 		flag:    "since-time",
 		want:    "",
 		wantErr: false,
 	}, {
-		name:    "expect since=empty, err=nil when format since is set",
+		name:    "expect since=empty, err=nil when since flag is set",
 		flag:    "since",
 		want:    "",
 		wantErr: false,
 	}, {
-		name:    "expect output=empty, err=nil when format output is set",
+		name:    "expect output=empty, err=nil when output flag is set",
 		flag:    "output",
 		want:    "",
 		wantErr: false,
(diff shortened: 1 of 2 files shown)


pkg/preflight/flags_test.go
```diff
@@ -0,0 +1,109 @@
...
+			AddFlags(&f)
+
+			if err := f.Parse(tt.args); err != nil {
+				if (err != nil) != tt.wantErr {
+					t.Errorf("AddFlags() error = %v, wantErr %v", err, tt.wantErr)
+				}
+			} else {
+				for i, flag := range tt.flags {
+					got, err := f.GetString(flag)
+					if (err != nil) != tt.wantErr {
```
banjoh: This is OK, but I like testify's interface better. This would look something like

```go
assert.Equalf(t, (err == nil), tt.wantErr, "AddFlags() error = %v, wantErr %v", err, tt.wantErr)
```

pkg/preflight/flags_test.go
```diff
@@ -0,0 +1,109 @@
...
+						t.Errorf("AddFlags() error = %v, wantErr %v", err, tt.wantErr)
+					}
+
+					if got != tt.want[i] {
+						t.Errorf("AddFlags() = %v, want %v", got, tt.want[i])
```
banjoh: testify me please :)

pkg/preflight/flags_test.go
```diff
@@ -0,0 +1,109 @@
...
+			resetFlags()
+			f := flag.FlagSet{}
+			AddFlags(&f)
+
+			if err := f.Parse(tt.args); err != nil {
+				if (err != nil) != tt.wantErr {
+					t.Errorf("AddFlags() error = %v, wantErr %v", err, tt.wantErr)
+				}
+			} else {
+				for i, flag := range tt.flags {
```
banjoh: I would do away with this for loop to simplify the test implementation by having `args`, `flags` and `want` as strings rather than slices. Our test inputs have one value each

pkg/preflight/flags_test.go
```diff
@@ -0,0 +1,109 @@
...
+		wantErr: false,
+	}}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resetFlags()
+			f := flag.FlagSet{}
+			AddFlags(&f)
+
+			if err := f.Parse(tt.args); err != nil {
```
banjoh: Minor comment: Perhaps this test is unnecessary. We are testing cobra's ability to parse commands. Unless your intention is to ensure cobra does not break its interface in the future, in which case the required test coverage would be much more than this :)

pkg/preflight/flags_test.go
```diff
@@ -0,0 +1,109 @@
+package preflight
```
banjoh, line 1: Big +1 on adding unit tests

pkg/preflight/flags_test.go
DexterYan: migrated into
```
assert.Equalf(t, (err != nil), tt.wantErr, "AddFlags() error = %v, wantErr %v", err, tt.wantErr)
```

pkg/preflight/flags_test.go
DexterYan: migrated to 
```
assert.Equalf(t, got, tt.want, "AddFlags() = %v, wantErr %v", got, tt.want)
```

pkg/preflight/flags_test.go
DexterYan: good point, I removed slices.

pkg/preflight/flags_test.go
DexterYan: Really good point! I have restructured the tests. My intention is about to check the add flags and default flag value. It does not need to use the cobra parse function. Thank you for pointing it out.

banjoh (dismissed):

banjoh:
You might want to squash your commits before merging.

chris-sanders:
> You might want to squash your commits before merging.

I believe it's actually required on the repo before merge.

DexterYan:
> > You might want to squash your commits before merging.
> 
> I believe it's actually required on the repo before merge.

Thank you for reminder. I have enabled squash and merge automatically.

xavpaice (approved):

PR #840, opened 2022-11-15
Remove ingress check from https://preflight.replicated.com

camilamacedo86:

Currently, when we install an k8s with kURL and we run kubectl preflight https://preflight.replicated.com it will fail:

> ------------
> Check FAIL
> Title: Ingress
> Message: Contour ingress not found!

Therefore,  Contour ingress does not seems a pre-requirement to kURL. So, should we have this check in the default example/test?


chris-sanders (approved):

camilamacedo86:
Hi @chris-sanders: 
 
We might want to revert this one :https://github.com/replicatedhq/kURL/issues/3734#issuecomment-1315507032

PR #842, opened 2022-11-15
Remove nodes check from https://preflight.replicated.com

camilamacedo86:

Currently, when we install an k8s with kURL and we run kubectl preflight https://preflight.replicated.com/ it will fail:

```
Check FAIL
Title: Must have at least 3 nodes in the cluster, with 5 recommended
Message: This application requires at least 3 nodes.
```

For some specific App that seems OK. However, it does not seems a kURL requirement and also, would it really be a requirement and/or recommendation to any app?  

### DIFF
--- a/examples/preflight/sample-preflight.yaml
+++ b/examples/preflight/sample-preflight.yaml
@@ -75,19 +75,6 @@ spec:
           - warn:
               message: Unable to determine the distribution of Kubernetes
     - nodeResources:
-        checkName: Must have at least 3 nodes in the cluster, with 5 recommended
-        outcomes:
-        - fail:
-            when: "count() < 3"
-            message: This application requires at least 3 nodes.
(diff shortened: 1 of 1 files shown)


chris-sanders (approved):

chris-sanders:
@camilamacedo86 this one doesn't merge cleanly with one of the others can you resolve the conflict?

camilamacedo86:
@chris-sanders,

Instead of remove the check we might want only make very clear that this spec is ONLY example. See; https://replicated.slack.com/archives/D048NEXPG4S/p1668527700566499
We have not a really good away to ensure in the long run that the example will be valid with kurl or any other replicated solution.

camilamacedo86:
Closing this one so that does not get merged.
We decide to just clarifies its purpose.

PR #843, opened 2022-11-15
Minimal memory requirement for kURL is 8GI

camilamacedo86:

Minimal requirement for kURL is 8GI and not 10GI.
See: https://kurl.sh/docs/install-with-kurl/system-requirements#minimum-system-requirements


chris-sanders (approved):

PR #845, opened 2022-11-15
doc/readme : make clear that the links are just examples

camilamacedo86:

Just ensure that users might not be misleading when they are trying to understand the replicated org solutions and does not know that those links are only preflight examples and has no relation with kURL or Kots. 

## Checklist

- [ ] New and existing tests pass locally with the changes.

- [ ] Yes
- [ ] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ curl https://krew.sh/preflight | bash
  and run:
  
 ```
-# It is ONLY an example. Do **not** use to validate cluster state.
+# It is ONLY an example. Do **not** use to validate real scenarios.
 kubectl preflight https://preflight.replicated.com
 ```
 
@@ -33,7 +33,7 @@ curl https://krew.sh/support-bundle | bash
  and run:
  
 ```
-# It is ONLY an example. Do **not** use to validate applications.
+# It is ONLY an example. Do **not** use to validate real scenarios.
 kubectl support-bundle https://support-bundle.replicated.com
 ```
 
--- a/README.md
+++ b/README.md
@@ -12,12 +12,13 @@ To run a sample preflight check from a sample application, install the preflight
 ```
 curl https://krew.sh/preflight | bash
 ```
- and run:
+ and run, where https://preflight.replicated.com provides an **example** preflight spec:
  
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,7 @@ curl https://krew.sh/preflight | bash
  and run:
  
 ```
+# It is ONLY an example. Do **not** use to validate cluster installs.
```
camilamacedo86:
```suggestion
# It is ONLY an example. Do **not** use to validate cluster state.
```

README.md
```diff
@@ -15,6 +15,7 @@ curl https://krew.sh/preflight | bash
  and run:
  
 ```
+# It is ONLY an example. Do **not** use to validate cluster state.
```
camilamacedo86, line 18:
```suggestion
# It is ONLY an example. Do **not** use to validate real scenarios.
```

README.md
```diff
@@ -32,8 +33,10 @@ curl https://krew.sh/support-bundle | bash
  and run:
  
 ```
+# It is ONLY an example. Do **not** use to validate applications.
```
camilamacedo86, line 36:
```suggestion
# It is ONLY an example. Do **not** use to validate real scenarios.
```

README.md
```diff
@@ -15,6 +15,7 @@ curl https://krew.sh/preflight | bash
  and run:
  
 ```
+# It is ONLY an example. Do **not** use to validate real scenarios.
```
xavpaice: I'd like to move this out of the command example and have it directly above.  e.g.:

"and run:" becomes

"and run, where  https://preflight.replicated.com provides an example preflight spec:"

we should also add a note to "To run a sample preflight check from a sample application,":

"To run a sample preflight check from a sample application (NOTE: this is an example. Do **not** use to validate real scenarios.):"

xavpaice (commented):
Great idea - as noted inline, I'd prefer to have the example note outside of copyable text if possible.

It's a nit, not blocking on that.

README.md
camilamacedo86: changed let me wdyt?

xavpaice (approved):
This is great, thanks!

Issue #808, opened 2022-10-28
Provide container images on release


chris-sanders:
**Describe the rationale for the suggested feature.**

While building release it wouldn't be a lot of extra effort to build and host minimal images on ghcr alongside the binary releasees. This provides an additional option and makes consuming Troubleshoot to make checks in-cluster a lot easier. For example if you wanted to use Helm Hooks to check before upgrading your application.

**Describe the feature**

Build images for each of the projects binaries on release so that they can be run in-cluster if desired without a user having to create and maintain their own troubleshoot images.

xavpaice:
https://github.com/replicatedhq/troubleshoot/blob/main/deploy/Dockerfile.troubleshoot has the current Dockerfile used to make the Docker images released with each tag.  These need testing.

xavpaice:
Tested the image on Dockerhub, it works fine and is updated.

PR #848, opened 2022-11-16
Add ghcr to package releases

chris-sanders:

From #808 we've added images but they are being uploaded to dockerhub. This adds them to ghcr because dockerhub now has pull restrictions that ghcr doesn't.

- [ ] Yes
- [x] No


chris-sanders:
I don't think I can fully tests this without making a release. I believe after the images have been pushed I also need to go assign them to this repository at the organization level and then they will show up on this repo's package list.

xavpaice (approved):

PR #857, opened 2022-11-20
feat: collecting and analyzing host packages

ricardomaraschini:

In some cases, it is important to collect what packages are installed in the Operating System. This PR adds a collector and an analyzer for these cases. Here is an example of collecting and installing packages:

```yaml
kind: HostPreflight
metadata:
  name: check-installed-packages
spec:
  collectors:
    - installedPackages:
        collectorName: "Installed packages"
```

For analyzing this PR also introduces the concept of a **templated** analyzer (an analyzer whose properties need to be templated out before running). Here is an analyzer to check whether a given package is present or not but only on an Ubuntu installation:

```yaml
spec:
  analyzers:
    - installedPackage:
        collectorName: "Installed packages"
        packageName: "dbus-user-session"
        exclude: '{{ne .Platform "ubuntu"}}'
        outcomes:
        - pass:
            when: "> 1.12.0"
            message: "dbus-user-session version > 1.12.0 is installed"
        - fail:
            message: "dbus-user-session version > 1.12.0 is required"
```

As not all installed packages comply with semantic versioning the analyzer attempts to use the `when` clause as a regex if the package is installed but does not comply with semantic versioning.

As now this PR does not implement any test (hence the Draft), the goal is to spark some discussions about how useful this can be and, especially, about the new "templated" analyzer concept.

## Extra

- If no outcome has been specified the analyzer passes if the package is installed (regardless of the version) and fails if not installed.
- This PR contains support for Debian and RedHat operating systems but it should be generic enough to support other flavours o the same family (tested in Ubuntu and Fedora).

## Checklist

- [X] New and existing tests pass locally with the changes.

- [ ] Yes
- [X] No


diamonwiggins:
Curious to know if you've had a look at https://kurl.sh/docs/create-installer/host-preflights/system-packages? 

https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/host_system_package.go
https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/host_system_packages.go

I still haven't digged too deep into this PR yet, but at surface levels seems like there's some overlap here. If that's the case, should make a decision if we want to build on the existing collector or deprecate it with this addition.

ricardomaraschini:
Closing this. I was not aware this feature already existed.

PR #874, opened 2022-11-26
Includes virtual memory parameters in Sysctl

crdant:

TL;DR
-----

Updates Sysctl collector and analyzer for virtual memory parameters

Details
-------

Adds supoort for virtual memory parameters to the Sysctl collector and analyzers. I uncovered this writing a pre-flight for a Helm chart that includes ECK as a subchart. Since ECK requires a specific minimum value for `vm.max_map_count` I wanted to use the Sysctl analyzer to check for the expected value, but wasn't able to because of the limited values it collected. I also learned that Sonarqube expects the same parameter to be increased, so it seemed like a general enough requirement to add it in.

The code updates the collector to collect values under `/proc/sys/vm`and adds tests to the analyzer to based on the ECK requirements. Making the tests pass required adding operators to the when expression, since the existing code only allowed for `=`, `==`, and `===`. The when expression now supports `>`, `<`, `>=`, and `<=`.

## Checklist

- [X] New and existing tests pass locally with the changes.

- [ ] Yes
- [X] No


diamonwiggins (approved):
lgtm

diamonwiggins:
going to go ahead and merge. be sure to update the docs at https://troubleshoot.sh as well - https://github.com/replicatedhq/troubleshoot.sh

PR #887, opened 2022-12-02
feat(redactors): Run redactors on an existing support bundle

### LINKED ISSUE #705: Add support to run redactors on an existing support bundle

If a support bundle with sensitive information is generated, it would be beneficial if there was a way to redact this information from the support bundle after the fact. At the moment a redactor can be added and another support bundle recreated, but there may be cases where a cluster may have mutated leading to losing helpful troubleshooting data such as logs (think log rotation).

**Describe the feature**

With a support bundle at hand, I would envision two ways of recreating the support bundle.
- In the application manager UI, add a way to reprocess an existing support bundle (run the redactors and analysers only) _(this is out of scope of troubleshoot because its part of KOTS, but adding here to describe a possible use case)_.
- On the CLI, run something akin to the following. Passing in a support bundle implies not to run collectors.
```
kubectl support-bundle ./support-bundle-with-new-redactor.yaml --support-bundle.tar.gz
```

banjoh:

Add a `redact` subcommand to `support-bundle` to allow running redactors on an existing bundle, creating a new redacted bundle as a result.

The command will be launched like so

```sh
support-bundle redact <redactors> --bundle support-bundle.tar.gz --output redacted-support-bundle.tar.gz
```

Fixes: #705

## Checklist

- [ ] Yes
- [x] No

## Additional notes
- Documentation to troubleshoot.sh will be done when we address https://github.com/replicatedhq/troubleshoot.sh/issues/414


chris-sanders (commented):
From the example, where is the new bundle written to? Are we overweighting the provided bundle?

I'd like to avoid commands being destructive (overwriting the input) if possible. I'd hate for a command to be run that can't be un-done. We could accept a new output name (probably should) and default to appending a suffix if not provided or something.

banjoh:
> From the example, where is the new bundle written to? Are we overweighting the provided bundle?
> 
> I'd like to avoid commands being destructive (overwriting the input) if possible. I'd hate for a command to be run that can't be un-done. We could accept a new output name (probably should) and default to appending a suffix if not provided or something.

The output will definitely be a new file. I'm with you here. We should not re-write the old file. I hint this in the usage description [here](https://github.com/banjoh/troubleshoot/blob/em/redactors-on-existing-sb/cmd/troubleshoot/cli/redact.go#L18). I've not yet settled on the naming convention though :)

chris-sanders:
Ah I see it now, great thanks!

README.md
```diff
@@ -12,13 +12,13 @@ To run a sample preflight check from a sample application, install the preflight
 ```
 curl https://krew.sh/preflight | bash
 ```
-and run, where https://preflight.replicated.com provides an **example** preflight spec:
- 
+ and run, where https://preflight.replicated.com provides an **example** preflight spec:
```
banjoh, line 15: README changes are spaces removed after formatting.

docs/preflight.md
```diff
@@ -17,7 +17,7 @@ preflight [url] [flags]
       --as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
       --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
       --as-uid string                  UID to impersonate for the operation.
-      --cache-dir string               Default cache directory (default "/Users/xavpaice/.kube/cache")
+      --cache-dir string               Default cache directory (default "/Users/evans/.kube/cache")
```
divolgin: i know you didn't add this, but this is a very strange default....

pkg/collect/result.go
```diff
@@ -315,3 +315,52 @@ func TarSupportBundleDir(bundlePath string, input CollectorResult, outputFilenam
...
+func CollectorResultFromBundle(bundleDir string) (CollectorResult, error) {
+	// Check directory exists
+	if _, err := os.Stat(bundleDir); os.IsNotExist(err) {
+		return nil, errors.Wrap(err, "bundle directory does not exist")
+	}
+
+	// Ensure this is indeed a support bundle directory
+	// All support bundles have a version.yaml file
+	// TODO: We might want to unmarshal the version.yaml file as well
+	_, err := os.Stat(filepath.Join(bundleDir, "version.yaml"))
```
divolgin: I don't think this check should be here.  It's just a bundle of files.  Nothing in it makes it a valid or an invalid bundle.

docs/preflight.md
xavpaice: eek.  That's autogenerated documentations!

pkg/collect/result.go
banjoh: Good point.

docs/preflight.md
banjoh: I modified the docs. It should be OK cause for now we manually generate the docs and update them. When the process becomes automated we might need some `sed s/old/new/g` magic

adamancini (approved):
I don't think I have any objections

divolgin (approved):

PR #888, opened 2022-12-05
patch cluster_resources.go

edgarlanting:
Proposed change for the json filename output in the bundle for `clusterRoleBindings` to all lowercase to be similar to `roles`, `clusterroles` & `rolebindings`.

I'm proposing this change in order to clean up file naming and use one specific convention.

## Checklist

- [-] Tests for the changes have been added (for bug fixes / features)
- [-] Any documentation required has been added/updated. For changes to https://troubleshoot.sh/ create a PR [here](https://github.com/replicatedhq/troubleshoot.sh/pulls)

- [ ] Yes
- [X ] No

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -294,8 +294,8 @@ func (c *CollectClusterResources) Collect(progressChan chan<- interface{}) (Coll
 
 	//Cluster Role Bindings
 	clusterRoleBindings, clusterRoleBindingsErrors := clusterRoleBindings(ctx, client)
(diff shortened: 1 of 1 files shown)


chris-sanders (commented):
This is a breaking change, is this important enough to make breaking?
I don't recall how long this has been available but since it was contributed by a community member (right?) we should assume this is in use. What is the reason to break that users current use of it?

edgarlanting:
> This is a breaking change, is this important enough to make breaking? I don't recall how long this has been available but since it was contributed by a community member (right?) we should assume this is in use. What is the reason to break that users current use of it?

This has been available since Oct 17th 2022 (Date merged). But will hold off this PR for now as it is not necessarily needed.

PR #894, opened 2022-12-08
Allow Preflight CLI to consume multiple specs as input 

diamonwiggins:

Updates the Preflight CLI to allow multiple specs via arguments. All specs will be concatenated into a single resultant spec which will be used for collection and analysis

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/cmd/preflight/cli/root.go
+++ b/cmd/preflight/cli/root.go
@@ -30,7 +30,7 @@ that a cluster meets the requirements to run an application.`,
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v := viper.GetViper()
-			return preflight.RunPreflights(v.GetBool("interactive"), v.GetString("output"), v.GetString("format"), args[0])
+			return preflight.RunPreflights(v.GetBool("interactive"), v.GetString("output"), v.GetString("format"), args)
 		},
 	}
 
--- a/pkg/preflight/concat.go
+++ b/pkg/preflight/concat.go
@@ -8,7 +8,6 @@ func ConcatPreflightSpec(target *troubleshootv1beta2.Preflight, source *troubles
 	newSpec := target.DeepCopy()
 	newSpec.Spec.Collectors = append(target.Spec.Collectors, source.Spec.Collectors...)
 	newSpec.Spec.RemoteCollectors = append(target.Spec.RemoteCollectors, source.Spec.RemoteCollectors...)
-	//newSpec.Spec.UploadResultsTo = append(target.Spec.UploadResultsTo, source.Spec.UploadResultsTo...)
 	newSpec.Spec.Analyzers = append(target.Spec.Analyzers, source.Spec.Analyzers...)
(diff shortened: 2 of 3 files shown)


diamonwiggins:
This is in draft because it is still lacking tests, but also because I've discovered a part of the `kind: Preflight` spec which can't be concatenated: `UploadResultsTo` - https://github.com/replicatedhq/troubleshoot/blob/main/pkg/apis/troubleshoot/v1beta2/preflight_types.go#L25

We'll need to decide what we want to do here. A couple options i've thought of so far:
1. Take the value from the first spec in which it is found :trollface: 
2. Update the Preflight spec and code to accept multiple locations. Send the results to all of them. 
3. Don't concat specs with `uploadResultsTo` and execute them separately
4. Move `UploadResultsTo` out of the Preflight spec and have it be a CLI flag

Option 2 & 4 could work depending on requirements, but is a breaking change. 

If there is any scenario in which one of the many potential specs that is provided only wants its analyzer results uploaded and not the results of all other specs gathered at processing time, Option 3 seems as though it would be best.

diamonwiggins:
After some additional discussion, it's not clear if `kind: Preflight` and the use cases surrounding it would benefit from any concatenating at all. The distributed use case with `kind: SupportBundle` where you are collecting and analyzing **separately maintained** components across an entire k8s cluster isn't quite present with `kind: Preflight` because a single spec is tightly coupled to an entire application you're deploying to the cluster.

It seems like the best path forward would be to:

1. Don't concat `kind: Preflight` and process each spec individually preventing collisions with `uploadResultsTo` although that's very unlikely to happen.
2. Concat `kind: HostPreflight` as expected with this work
3. Possible Future Item: Remove `uploadResultsTo` from `kind: Preflight` and make it a cli flag instead? Do they need to be tightly coupled together as they are today?

Outstanding questions:
1. Do we envision some sort of microservices pattern emerging with Preflights where individual spec files are maintained in an application? If so, concatenating would be beneficial, but brings back the discussion of what we do with `uploadResultsTo`.

divolgin:
> After some additional discussion, it's not clear if kind: Preflight and the use cases surrounding it would benefit from any concatenating at all

I think these tools could benefit from some consistency.  If `support-bundle` allows multiple specs, then `preflight` should do so as well.  But I agree that this functionality is somewhat extraneous to troubleshoot in general.

> Don't concat kind: Preflight and process each spec individually preventing collisions with uploadResultsTo although that's very unlikely to happen.

Internally this can be implemented either way to solve the problem of "consume multiple specs as input".  As you have pointed out, `uploadResultsTo` should only apply to the spec where it's defined.  On the flip side, each preflight run will execute collectors, which will be expensive.  So this decision boils down to which of these two problems you want to deal with.

> Remove uploadResultsTo from kind: Preflight and make it a cli flag instead

We'll have to start a new project if we can't rely on this one being somewhat backwards compatible.  :)

divolgin (approved):

PR #898, opened 2022-12-09
PR for issue #878 - Limit logfile size for support bundle

### LINKED ISSUE #878: Max size of pod logs
**Internal Tracking**: [#55381](https://app.shortcut.com/replicated/story/55381)

### Problem to solve
Setting a max size limit on the pod logs. This is to ensure that the support bundle is not too large, the existing limit of max age or max lines would not necessarily help with this as there could be lines which are extremely long and thus increase the bundle size.

### Impact
This could result in large support bundles which would be harder to obtain from customers. Also this was possible in both Ship and native schedulers so should exist to keep feature parity 

### Proposed Solution
Add `limits.maxSize` or `limits.maxBytes` field to the pod logs parameters

edgarlanting:

In order to limit the maximum size of a gathered logfile in a support bundle, this PR addresses setting a hard limit of a maximum size of 5MB to a pod log. This is needed so we won't end up with support-bundles that are too large to send/share and information in the pod logs that is not usable due to being too old or no longer relevant.

Fixes: #878

## Checklist

TODO: Specify that the maximum size of a pod log can now _only_ be 5MB

- [ ] Yes
- [x] No

In essence I don't believe limiting the logfile size can be seen as a breaking change.


edgarlanting:
Maximum allowed size of pod logs has been implemented. Currently working on tests

edgarlanting:
Current status:

Did a minor rewrite of the `MaxBytes` implementation in `logs.go`. I am however still having issues implementing properly running tests but there seems to be an issue with the implementation logic. 

Next steps: 

Review complete implementation and make changes where needed.

pkg/collect/logs.go
```diff
@@ -264,6 +264,18 @@ func setLogLimits(podLogOpts *corev1.PodLogOptions, limits *troubleshootv1beta2.
 	} else {
 		podLogOpts.TailLines = &limits.MaxLines
 	}
+
+	defaultMaxBytes := int64(5000000)
+	if limits.MaxBytes == 0 {
+		podLogOpts.LimitBytes = &defaultMaxBytes
+		return
+	}
```
banjoh, line 272: This potion is not necessary. The code below handles this case.

pkg/collect/logs.go
```diff
@@ -264,6 +264,18 @@ func setLogLimits(podLogOpts *corev1.PodLogOptions, limits *troubleshootv1beta2.
...
+
+	if limits.MaxBytes == 0 {
```
banjoh, line 274: What is you have a negative value (`limits.MaxBytes = -200`)

edgarlanting: This does not affect the working, logs are created regardless of the negative value.

pkg/collect/logs_test.go
```diff
@@ -63,19 +74,26 @@ func Test_setLogLimits(t *testing.T) {
 			actual := corev1.PodLogOptions{}
 			setLogLimits(&actual, test.limits, convertMaxAgeToTime)
 
-			if test.expected.TailLines != nil {
-				req.NotNil(actual.TailLines)
-				assert.Equal(t, *test.expected.TailLines, *actual.TailLines)
+			if test.expected.LimitBytes != nil {
+				req.NotNil(actual.LimitBytes)
```
banjoh: While here, could you please change using `require` to `assert`? The former is meant to ensure a mandatory precondition is set, not assert a test case. In practice, if `require` were to fail, the test run stops at that point, which we would not want. The preference is to run through all test assertions and report all failures.

The change in question would be `req.NotNil(actual.LimitBytes)` to `assert.NotNil(t, actual.LimitBytes)`. Please to the same for all other places that have `req.<function>`

edgarlanting: Thats indeed exactly what we talked about earlier, thanks @banjoh. So you would ask yourself when would a test stop, after the 1st, 2nd or 3rd? which is something you indeed do not want. We do want all tests to be ran through.

Next steps would then be to rewrite using `assert..` instead of the currently specified `req..` and make sure we have a proper function testset for Troubleshoot logs.

PR #916, opened 2022-12-16
Use GenerateName for runpod collector

### LINKED ISSUE #895: Multiple runPod specs with the same name overwrite each other's data
**Bug Description**

If two or more collector specs include the `runPod` collector with the same `name`, the second one to run will overwrite the data from the first.

Although it's not possible right now, if two run at the same time, with the same pod name, there will be a collision.

**Expected Behavior**

Each collector has a unique data output.

Each `runPod` collector starts a pod with a unique name.

**Steps To Reproduce**

Include the following in a spec:

```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - runPod:
        name: "run-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
    - runPod:
        name: "run-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
            args: ["-w", "5", "www.reddit.com"]
```

There may be other collectors that have the same problem, we have merely identified this with `runPod` at this point in time.

**Additional Context**

0.49.1

diamonwiggins:

Uses `generateName` to give a unique name to the Pods that are created by the `runPod` collector. This will prevent conflicts in the event that an old runPod collector is still running from a failed run, or if one day the `runPod` collector were to be run conurrently

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -333,7 +333,7 @@ the %s Admin Console to begin analysis.`
 			return nil
 		}
 
-		fmt.Printf("%s\n", response.ArchivePath)
+		fmt.Printf("\n%s\n", response.ArchivePath)
 		return nil
 	}
 
--- a/examples/preflight/e2e.yaml
+++ b/examples/preflight/e2e.yaml
@@ -44,7 +44,7 @@ spec:
               message: You have at least 5 replicas
     - textAnalyze:
         checkName: Said hi!
-        fileName: /static-hi.log
+        fileName: /static-hi*.log
         regex: 'hi static'
         outcomes:
           - fail:
--- a/pkg/collect/run_pod.go
+++ b/pkg/collect/run_pod.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"sync"
 	"time"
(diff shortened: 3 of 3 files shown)


diamonwiggins:
@xavpaice leaving this in draft atm because this would be a breaking change as the destination log file is the name of the pod which will now not have a consistent name as before. Thoughts? Do we just update our documentation to reflect this and proceed? New text analyzer would now need to glob like so...

```
kind: SupportBundle
metadata:
  name: bundle1
spec:
  collectors:
    - runPod:
        name: "runpod-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
  analyzers:
    - textAnalyze:
        checkName: mycheck
        fileName: /runpod-ping/runpod-ping*.log
        regex: 'something'
        outcomes:
          - pass:
              when: "false"
              message: "it passed"
          - fail:
              when: "true"
              message: "it failed"
```

edit. the e2e tests actually caught this :)

diamonwiggins:
Actually not even sure now If you can currently glob the filename based on the test still failing. will come back to this at another time

xavpaice:
No matter what you do here, if you don't save the output from each runPod collector into the unique pod's name, we can't avoid duplicates overwriting that file.  I therefore don't see a choice but to use the wildcard (if that's possible) in the analyzer, and though it's officially a breaking change it's also correcting a bug we simply haven't identified previously.

The only other way I can think to get around it is to do some duplicate detection and resolution during the merge of the specs, to highlight that there's a problem with multiple runPod collectors colliding and maybe rename one or both - in which case, one of the analyzers would run, but maybe not on the correct output file!

xavpaice:
Just tested the tests using main, and it looks like we'd need to introduce a separate change for having wildcard support in the textAnalyze code.

PR #918, opened 2022-12-19
rename the troubleshoot label to "support-bundle"

### LINKED ISSUE #913: `supportbundle-spec` label and `support-bundle-spec` data key in a secret-wrapped support bundle is inconsistent
**Bug Description**

```
apiVersion: v1
kind: Secret
metadata:
  name: flannel-troubleshoot-spec
  labels:
    troubleshoot.io/kind: supportbundle-spec
stringData:
  support-bundle-spec: |
```

having both of these right next to each other is jarring and I think a potential source of confusion, since we don't have validation on the contents of Secrets, and we are potentially doing automation based on this label.  let's at least try to be consistent.   

**Expected Behavior**

perhaps something like `troubleshoot.io/kind: support-bundle` or `troubleshoot.io/kind: support-bundle-kind` would be a little more consistent with the data key

adamancini:

rename the troubleshoot.io label `supportbundle-spec` into `support-bundle` to be more consistent with the data key `support-bundle-spec`.

resolves #913 

## Checklist

  - [x] https://github.com/replicatedhq/replicated-docs/pull/832
  - [x] https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/
  - [x] specs consumed as Secrets by kURL project need to be updated for the new label

- [x] Yes
- [ ] No

This changes the default label used by Troubleshoot to discover specs in the cluster.  Existing specs that specified `troubleshoot.io/kind=supportbundle-spec` will need to update to the new label `troubleshoot.io/kind=support-bundle`.

This change will not cause troubleshoot to cease to function, since an alternative label can be specified at the CLI


xavpaice (approved):
LGTM.  Please ensure the commit message is clear this is a breaking change

adamancini:
@xavpaice there are a few docs that need to be updated but this is a change to existing, so the docs exist already - just need to be corrected and I think I started linking those places.  and "breaking" might be a hard call;  the only thing that would change is that the default label that we use for discovery is changing.  so old specs (of which there should be very few) may need to have their labels updated.  kURL components that ship troubleshoot manifests are probably the only thing really impacted at this stage.

PR #920, opened 2022-12-20
feat(cluster_resources): increase default client burst and qps

DexterYan:

- Increase default QPS from 5 to 100
- Increase default Burst from 10 to 100
- Add UserAgent Header to identify it is from Troubleshoot once it is increasing the number of queries

Shortcut:
- https://app.shortcut.com/replicated/story/22748/preflights-are-very-slow-on-openshift

Issues: 
- https://github.com/replicated-collab/settlemint-replicated/issues/87
- https://github.com/replicated-collab/schrodinger-replicated/issues/62

Reference:
- https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/k8sutil/k8sutil.go#L109

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/preflight/collect.go
+++ b/pkg/preflight/collect.go
@@ -11,6 +11,7 @@ import (
 	analyze "github.com/replicatedhq/troubleshoot/pkg/analyze"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"github.com/replicatedhq/troubleshoot/pkg/collect"
+	"github.com/replicatedhq/troubleshoot/pkg/version"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
@@ -130,6 +131,10 @@ func Collect(opts CollectOpts, p *troubleshootv1beta2.Preflight) (CollectResult,
 	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterInfo: &troubleshootv1beta2.ClusterInfo{}})
 	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterResources: &troubleshootv1beta2.ClusterResources{}})
 	collectSpecs = collect.EnsureClusterResourcesFirst(collectSpecs)
+
+	opts.KubernetesRestConfig.QPS = 100
+	opts.KubernetesRestConfig.Burst = 100
+	opts.KubernetesRestConfig.UserAgent = fmt.Sprintf("ReplicatedTroubleshoot/%s", version.Version())
 
 	k8sClient, err := kubernetes.NewForConfig(opts.KubernetesRestConfig)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/preflight/collect.go
```diff
@@ -131,6 +132,10 @@ func Collect(opts CollectOpts, p *troubleshootv1beta2.Preflight) (CollectResult,
 	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterResources: &troubleshootv1beta2.ClusterResources{}})
 	collectSpecs = collect.EnsureClusterResourcesFirst(collectSpecs)
 
+	opts.KubernetesRestConfig.QPS = 100
```
banjoh: I'd create constants out of these magic numbers `const DEFAULT_CLIENT_QPS = 100` for example

```go
pkg
   constants
       constants.go
```

banjoh (commented):
Are we able to have these numbers configurable? While looking into options, you could also check if the kube config has these configuration parameters.

pkg/preflight/collect.go
```diff
@@ -131,6 +132,10 @@ func Collect(opts CollectOpts, p *troubleshootv1beta2.Preflight) (CollectResult,
...
+	opts.KubernetesRestConfig.Burst = 100
+	opts.KubernetesRestConfig.UserAgent = fmt.Sprintf("ReplicatedTroubleshoot/%s", version.Version())
```
adamancini: I guess my only suggestion might be to define this string as a constant and consume it from a variable, but I'm not sure where that should live

pkg/supportbundle/collect.go
```diff
@@ -77,6 +77,10 @@ func runCollectors(collectors []*troubleshootv1beta2.Collect, additionalRedactor
 	collectSpecs = collect.EnsureCollectorInList(collectSpecs, troubleshootv1beta2.Collect{ClusterResources: &troubleshootv1beta2.ClusterResources{}})
 	collectSpecs = collect.EnsureClusterResourcesFirst(collectSpecs)
 
+	opts.KubernetesRestConfig.QPS = 100
+	opts.KubernetesRestConfig.Burst = 100
+	opts.KubernetesRestConfig.UserAgent = fmt.Sprintf("ReplicatedTroubleshoot/%s", version.Version())
```
adamancini: same thing here as my previous note ^

pkg/preflight/collect.go
adamancini: oh probably `constants.go` 😂

pkg/preflight/collect.go
DexterYan: Thank you @banjoh! I have added constants.go

pkg/preflight/collect.go
DexterYan: Thank you @adamancini ! Yes, I have added constants.go and changed to
```
opts.KubernetesRestConfig.QPS = constants.DEFAULT_CLIENT_QPS
opts.KubernetesRestConfig.Burst = constants.DEFAULT_CLIENT_BURST
opts.KubernetesRestConfig.UserAgent = fmt.Sprintf("%s/%s", constants.DEFAULT_CLIENT_USER_AGENT, version.Version())
```

banjoh (approved):
Ship it

PR #925, opened 2022-12-21
Implementation of maximum pod logsize - Fixes issue 878

### LINKED ISSUE #878: Max size of pod logs

edgarlanting:

This PR introduces a new option to limit the size of a pod log when added to the bundle. This will make sure the support bundle will not grow to an unacceptable size and thus might contain information that is too old.

The maximum size of a pod log in a bundle is set to 5MB, and can be changed if we decide upon the need.

Changes include the sample support bundle, log collecting & corresponding tests. For more details please see the related codebase changes.

A Troubleshoot docs PR has been created at: https://github.com/replicatedhq/troubleshoot.sh/pull/457

Fixes #878 

## Checklist

- [X] Any documentation required has been added/updated. 

- [ ] Yes
- [X] No

Note: As this _only_ introduces a maximum size to the pod logs collected, it will still collect pod logs in a bundle.


pkg/collect/logs.go
```diff
@@ -264,6 +264,18 @@ func setLogLimits(podLogOpts *corev1.PodLogOptions, limits *troubleshootv1beta2.
 	} else {
 		podLogOpts.TailLines = &limits.MaxLines
 	}
+
+	defaultMaxBytes := int64(5000000)
+	if limits.MaxBytes == 0 {
+		podLogOpts.LimitBytes = &defaultMaxBytes
+		return
+	}
```
banjoh: These lines are not necessary

pkg/collect/logs_test.go
```diff
@@ -7,13 +7,13 @@ import (
...
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	testclient "k8s.io/client-go/kubernetes/fake"
 )
 
 func Test_setLogLimits(t *testing.T) {
+	maxBytes := int64(5000000)
```
banjoh, line 16: You can change this to a non default value (e.g 200) to test that case instead

pkg/collect/logs_test.go
```diff
@@ -58,23 +69,27 @@ func Test_setLogLimits(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			req := require.New(t)
 
 			actual := corev1.PodLogOptions{}
 			setLogLimits(&actual, test.limits, convertMaxAgeToTime)
 
+			if test.expected.LimitBytes == &maxBytes {
```
banjoh: You can change this line to `if test.expected.LimitBytes != nil {`

pkg/collect/logs_test.go
edgarlanting: I should have documented that in the PR comment, thanks for highlighting. I'll make a small change.

pkg/collect/logs_test.go
edgarlanting: Done

divolgin (dismissed):

pkg/collect/cluster_resources.go
```diff
@@ -169,6 +169,7 @@ func (c *CollectClusterResources) Collect(progressChan chan<- interface{}) (Coll
 		for _, container := range allContainers {
 			limits := &troubleshootv1beta2.LogLimits{
 				MaxLines: 500,
+				MaxBytes: 5000000,
```
divolgin, line 172: One of these limits seems redundant now.  Which one of them will be honored?

pkg/collect/logs.go
```diff
@@ -286,6 +286,12 @@ func setLogLimits(podLogOpts *corev1.PodLogOptions, limits *troubleshootv1beta2.
 	} else {
 		podLogOpts.TailLines = &limits.MaxLines
 	}
+
+	if limits.MaxBytes == 0 {
+		podLogOpts.LimitBytes = &limits.MaxBytes
+	} else {
+		podLogOpts.LimitBytes = &limits.MaxBytes
+	}
 }
```
xavpaice, line 298: what does this conditional do?

xavpaice (changes requested):
one query

pkg/collect/logs.go
edgarlanting: This was a temporary workaround during testing..This has been resolved.

xavpaice (approved):
LGTM.

Future improvement: add a test for a non-default maxBytes.

PR #926, opened 2022-12-22
chore: Add CLI flags to enable CPU & memory profiling

banjoh:

This PR add CLI flags that allow us to collect [CPU & memory profiling data](https://go.dev/doc/diagnostics#profiling). Here is how to use the flags

Run support bundle and with CPU & memory profile flags
```sh
./bin/support-bundle examples/support-bundle/sample-supportbundle.yaml --cpuprofile=cpu.prof --memprofile=mem.prof
```

Visualize using [pprof](https://github.com/google/pprof/blob/main/doc/README.md)
```sh
go tool pprof -http=":8000" cpu.prof

go tool pprof -http=":8001" mem.prof
```

## Checklist

- [ ] Yes
- [x] No


cmd/troubleshoot/cli/root.go
```diff
@@ -80,11 +82,3 @@ func initConfig() {
 	viper.SetEnvPrefix("TROUBLESHOOT")
 	viper.AutomaticEnv()
 }
-
```
banjoh, line 84: Some clean up. This function is not in use.

cmd/analyze/cli/root.go
```diff
@@ -28,11 +30,14 @@ func RootCmd() *cobra.Command {
 			}
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			v := viper.GetViper()
+			return util.ProfiledRunE(cmd, args, func(cmd *cobra.Command, args []string) error {
+				v := viper.GetViper()
 
-			logger.SetQuiet(v.GetBool("quiet"))
+				logger.SetQuiet(v.GetBool("quiet"))
+				fmt.Println(v.GetString("cpuprofile"))
```
xavpaice: was this left in deliberately?

xavpaice (changes requested):
one print that I suspect was left in accidentally.

I have a query, not blocking: personally I find the nested returns in the cli package to be difficult to read.  I wonder if there's a longer but maybe more readable approach we could use?  Just a thought.

cmd/analyze/cli/root.go
banjoh: This was left here accidentally.

Removed now

banjoh:
> I have a query, not blocking: personally I find the nested returns in the cli package to be difficult to read. I wonder if there's a longer but maybe more readable approach we could use? Just a thought.

I see your point. My first approach was inline with that. I included the boilerplate code required to setup and tear down the profiling environment in all the 4 `RunE` functions. It worked but in my opinion, it added quite a bit of duplicate code that made it slightly harder to read code that implements actual functionality. I found it better to "hide away" these kind of non-functional features to make it easier  "see the forest from the trees" (if that makes sense :) ).

One thing I did not explore was using `PreRun` & `PostRun` functions. I'll do that to see if the code becomes easier to understand.

banjoh:
> One thing I did not explore was using `PreRun` & `PostRun` functions. I'll do that to see if the code becomes easier to understand.

I tried this approach which I think looks cleaner @xavpaice

adamancini:
@banjoh only nitpick I have is that when I go to use the new flags, if I don't specify a file path, a weird little shell bug happens:

```
 ✔  moira  /tmp/foo  ~/go/src/github.com/replicatedhq/troubleshoot/bin/support-bundle ~/go/src/github.com/replicatedhq/troubleshoot/examples/support-bundle/sample-supportbundle.yaml --cpuprofile --memprofile

 Collecting support bundle ⠦ cluster-info2022/12/29 14:30:22 An analyzer failed to run: open /var/folders/5t/4b638lkj34x81k36l95zw9ph0000gn/T/supportbundle182731251/support-bundle-2022-12-29T14_30_10/cluster-resources/ingress/default.json: no such file or directory
support-bundle-2022-12-29T14_30_10.tar.gz
 ✔  moira  /tmp/foo  ls -lha
total 264
-rw-r--r--  1 ada   wheel    13K Dec 29 14:30 --memprofile
drwxr-xr-x  4 ada   wheel   128B Dec 29 14:30 .
drwxrwxrwt  8 root  wheel   256B Dec 29 14:29 ..
-rw-r--r--  1 ada   wheel   114K Dec 29 14:30 support-bundle-2022-12-29T14_30_10.tar.gz
 ✔  moira  /tmp/foo  less --memprofile                                                                  
There is no memprofile option ("less --help" for help)
Missing filename ("less --help" for help)
 ✔  moira  /tmp/foo  less \-\-memprofile
There is no memprofile option ("less --help" for help)
Missing filename ("less --help" for help)
 ✔  moira  /tmp/foo  
```

it's like `--cpuprofile --memprofile` gets parsed as `--cpuprofile='--memprofile'` and this file gets written, but then the shell has a hard time interacting with it because it looks like a long flag


maybe if there's no `=/path/to/file` in the argument, we should default to $pwd that way the parser doesn't try to do something weird?

banjoh:
> @banjoh only nitpick I have is that when I go to use the new flags, if I don't specify a file path, a weird little shell bug happens:
> 
> ```
>  ✔  moira  /tmp/foo  ~/go/src/github.com/replicatedhq/troubleshoot/bin/support-bundle ~/go/src/github.com/replicatedhq/troubleshoot/examples/support-bundle/sample-supportbundle.yaml --cpuprofile --memprofile
> 
>  Collecting support bundle ⠦ cluster-info2022/12/29 14:30:22 An analyzer failed to run: open /var/folders/5t/4b638lkj34x81k36l95zw9ph0000gn/T/supportbundle182731251/support-bundle-2022-12-29T14_30_10/cluster-resources/ingress/default.json: no such file or directory
> support-bundle-2022-12-29T14_30_10.tar.gz
>  ✔  moira  /tmp/foo  ls -lha
> total 264
> -rw-r--r--  1 ada   wheel    13K Dec 29 14:30 --memprofile
> drwxr-xr-x  4 ada   wheel   128B Dec 29 14:30 .
> drwxrwxrwt  8 root  wheel   256B Dec 29 14:29 ..
> -rw-r--r--  1 ada   wheel   114K Dec 29 14:30 support-bundle-2022-12-29T14_30_10.tar.gz
>  ✔  moira  /tmp/foo  less --memprofile                                                                  
> There is no memprofile option ("less --help" for help)
> Missing filename ("less --help" for help)
>  ✔  moira  /tmp/foo  less \-\-memprofile
> There is no memprofile option ("less --help" for help)
> Missing filename ("less --help" for help)
>  ✔  moira  /tmp/foo  
> ```
> 
> it's like `--cpuprofile --memprofile` gets parsed as `--cpuprofile='--memprofile'` and this file gets written, but then the shell has a hard time interacting with it because it looks like a long flag
> 
> maybe if there's no `=/path/to/file` in the argument, we should default to $pwd that way the parser doesn't try to do something weird?

This is indeed a weird case. I'm not sure how to solve it. Any approach I'm taking has some form of compromise I don't like

- Checking if the value is a valid file path. That's a hard one cause any string can be a valid file path, including `--memprofile`
- Checking if the value passed in is not one of the cli flags. Hard as well, cause I can pass in `--cpuprofile=cpuprofile`
- Checking the equal sign is also not very intuitive cause `support-bundle --cpuprofile cpu.prof` ought to work just like `support-bundle --cpuprofile=cpu.prof` according to how CLI applications behave.
- Stripping off `--` from the file name. Users will get confused

I'm open to suggestions here

xavpaice:
Regards the CLI switches being unable to handle a default when no filename is specified, I believe that is a limitation of the CLI library we're using and not one which can be solved in this PR.  Other flags with a filename would have a default and always write to the default if the switch is not supplied, but for this application we don't want anything output if we do not supply the flag.  The alternative I can think of is we have default paths which can be overridden with the flags, but default also to not use the profiling unless we add a 3rd switch `--enable-profiling` or some such.

Alternately, we live with this limitation and make sure help text is clear on how to use these flags.

banjoh:
> Alternately, we live with this limitation and make sure help text is clear on how to use these flags.

Agreed. This is how [pflag library](https://github.com/spf13/pflag#command-line-flag-syntax) is designed to work. If it becomes an issue needing fixing, we can address it in another GH issue.

cmd/util/profilers.go
```diff
@@ -0,0 +1,66 @@
...
+		defer f.Close()
+		runtime.GC() // get up-to-date statistics
+		if err := pprof.WriteHeapProfile(f); err != nil {
+			return fmt.Errorf("could not write memory profile: %w", err)
+		}
+	}
+
+	// Stop CPU profiling if it was started
+	if cpuProfileFile != nil {
+		pprof.StopCPUProfile()
```
divolgin: This will never happen if there is an error writing memory profile.  Ideally these two should be independent.

divolgin (dismissed):

diamonwiggins (approved):
lgtm

xavpaice (approved):

PR #928, opened 2022-12-23
fix: collect rbac permissions error

### LINKED ISSUE #929: the `--collect-without-permissions` flag is not respected
**Bug Description**

When running preflights with `--collect-without-permissions=false`, the process will still run even if the user does not have the proper permissions to access the collected resources.

```
 * cannot collect cluster-resources: action "list" is not allowed on resource "namespaces" at the cluster scope
 * cannot collect cluster-resources: action "list" is not allowed on resource "nodes" at the cluster scope
 * cannot collect cluster-resources: action "list" is not allowed on resource "customresourcedefinitions" at the cluster scope
 * cannot collect cluster-resources: action "list" is not allowed on resource "storageclasses" at the cluster scope
  Running Preflight Checks ⠏ 
```

**Expected Behavior**

Previous versions would output an error if the user did not have the required permissions and the program would exit:
```
 * cannot collect cluster-resources: action "list" is not allowed on resource "namespaces" at the cluster scope
 * cannot collect cluster-resources: action "list" is not allowed on resource "nodes" at the cluster scope
 * cannot collect cluster-resources: action "list" is not allowed on resource "customresourcedefinitions" at the cluster scope
 * cannot collect cluster-resources: action "list" is not allowed on resource "storageclasses" at the cluster scope
Error: failed to collect in cluster: insufficient permissions to run all collectors
```

**Steps To Reproduce**

Run the preflight binary with `--collect-without-permissions=false` as a user that does not have cluster-wide permissons and utilize the `clusterResources` collector and a trivial analyzer.

`/tmp/preflight.yaml`:
```
kind: Preflight
metadata:
spec:
  collectors:
    - clusterResources: {}
  analyzers:
    - clusterVersion:
        outcomes:              
            - fail:
                when: "< 1.16.0"
                message: The application requires at Kubernetes 1.16.0 or later
            - pass:
```

Run command:
```
./bin/preflight /tmp/preflight.yaml --collect-without-permissions=false
```

**Additional Context**

- Troubleshoot version: v0.50.0
- Operating system
- Operating system version

cbodonnell:

This PR fixes an issue where the `Collect(...)` function was not returning the expected error in cases where a collector failed because of RBAC permissions.  This is because `foundForbidden` was being reset to `false` on each iteration of the for loop, so even if it had previously been set to `true`, the final result may be incorrect.

This can be reproduced by running the preflight binary with the `--collect-without-permissions=false`.  The the expected behavior is the the program exits with: `Error: failed to collect in cluster: insufficient permissions to run all collectors`, but because of this bug, it does not.

Fixes: #929

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/preflight/collect.go
+++ b/pkg/preflight/collect.go
@@ -172,7 +172,6 @@ func Collect(opts CollectOpts, p *troubleshootv1beta2.Preflight) (CollectResult,
 			allCollectors = append(allCollectors, collectors...)
 		}
 
-		foundForbidden = false
 		for _, collector := range collectors {
 			for _, e := range collector.GetRBACErrors() {
 				foundForbidden = true


xavpaice (approved):
Thanks for finding this!

PR #943, opened 2023-01-04
feat(support-bundle): silent result if ceph or longhorn addon not enabled

### LINKED ISSUE #659: Support-bundle cli logs failures for ceph even though rook is not enabled
## Description

When generating a support bundle via CLI, the command logs errors/failures trying to collect and analyze rook ceph even though rook is not enabled in a cluster.

## Reproduction Steps

1. Deploy a kURL cluster without rook
2. run `kubectl support-bundle https://kots.io`

## Expected Behavior

We should not attempt to collect and analyze data for addons that are not enabled.

## Customer Impact

This is confusing for the customer. It adds noise and shows a failure that is a false negative.

## Possible Solution

Edit the Ceph and Longhorn collectors to first determine if that component is installed.  If not, do not attempt any collection, simply log a message to the effect that the collector was skipped.

DexterYan:

- if `ceph` is not enabled as add-on, return empty analyzer result silently without returning error 
- if `longhorn` is not enabled as add-on, return empty analyzer result silently without returning error 
- do not save those empty results into `analysis.json`
- if  `--debug` is supplied, error will be logged

Fixes: #659

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/analyze/analyzer.go
+++ b/pkg/analyze/analyzer.go
@@ -441,10 +441,9 @@ func Analyze(analyzer *troubleshootv1beta2.Analyze, getFile getCollectedFileCont
 		if err != nil {
 			return nil, err
 		}
-		if result == nil {
-			return nil, nil
+		if result != nil {
+			result.Strict = analyzer.CephStatus.Strict.BoolOrDefaultFalse()
 		}
-		result.Strict = analyzer.CephStatus.Strict.BoolOrDefaultFalse()
 		return []*AnalyzeResult{result}, nil
 	}
 	if analyzer.Longhorn != nil {
--- a/pkg/analyze/analyzer.go
+++ b/pkg/analyze/analyzer.go
@@ -441,6 +441,9 @@ func Analyze(analyzer *troubleshootv1beta2.Analyze, getFile getCollectedFileCont
 		if err != nil {
 			return nil, err
 		}
+		if result == nil {
+			return nil, nil
+		}
 		result.Strict = analyzer.CephStatus.Strict.BoolOrDefaultFalse()
 		return []*AnalyzeResult{result}, nil
 	}
--- a/pkg/analyze/ceph.go
+++ b/pkg/analyze/ceph.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"github.com/replicatedhq/troubleshoot/pkg/collect"
+	"github.com/spf13/viper"
 )
 
 type CephHealth string
@@ -104,6 +105,12 @@ func cephStatus(analyzer *troubleshootv1beta2.CephStatusAnalyze, getCollectedFil
 	collected, err := getCollectedFileContents(fileName)
 	analyzeResult := &AnalyzeResult{}
 	if err != nil {
+
+		v := viper.GetViper()
+		if v.GetBool("debug") {
+			return nil, errors.Wrap(err, "failed to read collected ceph status")
+		}
+
 		return analyzeResult, nil
 	}
 
--- a/pkg/analyze/ceph.go
+++ b/pkg/analyze/ceph.go
@@ -102,8 +102,9 @@ type PgMap struct {
 func cephStatus(analyzer *troubleshootv1beta2.CephStatusAnalyze, getCollectedFileContents func(string) ([]byte, error)) (*AnalyzeResult, error) {
 	fileName := path.Join(collect.GetCephCollectorFilepath(analyzer.CollectorName, analyzer.Namespace), "status.json")
 	collected, err := getCollectedFileContents(fileName)
+	analyzeResult := &AnalyzeResult{}
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to read collected ceph status")
+		return analyzeResult, nil
 	}
 
 	title := analyzer.CheckName
@@ -111,7 +112,7 @@ func cephStatus(analyzer *troubleshootv1beta2.CephStatusAnalyze, getCollectedFil
 		title = "Ceph Status"
 	}
 
-	analyzeResult := &AnalyzeResult{
+	analyzeResult = &AnalyzeResult{
 		Title:   title,
 		IconKey: "rook", // maybe this should be ceph?
 		IconURI: "https://troubleshoot.sh/images/analyzer-icons/rook.svg?w=11&h=16",
--- a/pkg/analyze/ceph.go
+++ b/pkg/analyze/ceph.go
@@ -9,7 +9,7 @@ import (
 	"github.com/pkg/errors"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"github.com/replicatedhq/troubleshoot/pkg/collect"
-	"github.com/spf13/viper"
+	"github.com/replicatedhq/troubleshoot/pkg/types"
 )
 
 type CephHealth string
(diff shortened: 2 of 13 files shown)


DexterYan:
You can run `./bin/support-bundle spec.yaml` to verify the result

`spec.yaml`
```
kind: SupportBundle
metadata:
  name: collector-sample
spec:
  collectors:
    - ceph: {}
    - longhorn: {}

  analyzers:
    - clusterVersion:
        outcomes:
          - fail:
              when: "< 1.16.0"
              message: The Admin Console requires at least Kubernetes 1.16.0
          - pass:
              message: Your cluster meets the recommended and required versions of Kubernetes
    - cephStatus: {}
    - longhorn: {}
```

pkg/convert/supportbundle.go
```diff
@@ -70,7 +70,7 @@ func FromAnalyzerResult(input []*analyze.AnalyzeResult) []*Result {
 	result := make([]*Result, 0)
 	for _, i := range input {
 		// Continue on nil result to prevent panic
-		if i == nil {
+		if i == nil || i.Title == "" {
```
banjoh: I assume `i.Title == ""` is meant to catch the empty Result case. I think an analyser returning a `nil` result is a better approach here to indicate that there is no analysis result.

pkg/analyze/ceph.go
```diff
@@ -102,16 +103,23 @@ type PgMap struct {
 func cephStatus(analyzer *troubleshootv1beta2.CephStatusAnalyze, getCollectedFileContents func(string) ([]byte, error)) (*AnalyzeResult, error) {
 	fileName := path.Join(collect.GetCephCollectorFilepath(analyzer.CollectorName, analyzer.Namespace), "status.json")
 	collected, err := getCollectedFileContents(fileName)
+	analyzeResult := &AnalyzeResult{}
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to read collected ceph status")
+
```
banjoh, line 106: I think we should differentiate other `error`s from `file-not-found` errors. `file-not-found` errors will be error logged in some cases (e.g what this PR proposes) while others will be propagated to the caller.

What you want to do is
- Create a new error type e.g `NotFoundError` and add it to a new types package (`pkg/types/types.go`)
```go
type NotFoundError struct {
    Name string
}

func (e *NotFoundError) Error() string { return e.Name + ": not found" }

// Comparison
if errors.Is(err, NotFoundError) {
    // err, or some error that it wraps, is a not-found error
}
```
For more on custom errors read https://go.dev/blog/go1.13-errors 

- Return or wrap not-found errors as `NotFoundError` [here](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/preflight/analyze.go#L55) and [here (you want to check file existence I believe, but maybe io.ReadFile returns an error that can be inspected)](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/download.go#L298)
- Log not found errors and return `nil, nil`, but return all other errors like before. Remove `v.GetBool("debug")`, its setup by library clients

banjoh (commented):
Approach is right, but I added a few comments on bits I think should change. Also remember to add unit tests

pkg/convert/supportbundle.go
DexterYan: I have removed `i.Title == ""` and return a nil analysis  result

pkg/analyze/ceph.go
DexterYan: Thank you @banjoh ! It is a great suggestion. I have added NotFoundError as type of new error and use it for file missing and other add-on not installed.

In #659, it required to hide on stdout for 
```
Longhorn: list nodes.longhorn.io: the server could not find the requested resource
rook ceph tools pod not found
```
So I added a block in collect to silent the error if it is `NotFoundError`.

However, I am not 100% sure that those changes did match you suggestions? Could you help me to review it again? If it matches, I will write those tests for it.

pkg/analyze/analyzer.go
```diff
@@ -441,6 +441,9 @@ func Analyze(analyzer *troubleshootv1beta2.Analyze, getFile getCollectedFileCont
 		if err != nil {
 			return nil, err
 		}
+		if result == nil {
```
banjoh: I'm guessing this needs to return a slice with a nil result else [this](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/download.go#L42) part with panic

Like so

```go
		if result != nil {
			result.Strict = analyzer.CephStatus.Strict.BoolOrDefaultFalse()
		}
		return []*AnalyzeResult{result}, nil
```

pkg/collect/ceph.go
```diff
@@ -215,11 +212,7 @@ func findRookCephToolsPod(ctx context.Context, c *CollectCeph, namespace string)
 		return &pods[0], nil
 	}
 
-	if v.GetBool("debug") {
-		return nil, errors.New("rook ceph tools pod not found")
-	}
-
-	return nil, nil
+	return nil, &types.NotFoundError{Name: "rook ceph tools pod"}
```
banjoh: You can just log that the pod was not found here using `logger` package. The log message will be silenced if `--debug` flag was not set. No need to return an error

pkg/collect/longhorn.go
```diff
@@ -66,13 +66,7 @@ func (c *CollectLonghorn) Collect(progressChan chan<- interface{}) (CollectorRes
...
 	nodes, err := client.Nodes(ns).List(ctx, metav1.ListOptions{})
 	if err != nil {
-
-		v := viper.GetViper()
-		if v.GetBool("debug") {
-			return nil, errors.Wrap(err, "list nodes.longhorn.io")
-		}
-
-		return output, nil
+		return nil, &types.NotFoundError{Name: "list nodes.longhorn.io"}
```
banjoh: Same here. Just log the message using `logger.Printf`.

Only return the `NofFoundError` if the function is used in multiple places and callers need to differentiate errors. This code here is collector logic which the collector will dictate on what to do with certain errors

pkg/analyze/analyzer.go
DexterYan: Thank you. Yes, that's right, I have added this

pkg/collect/ceph.go
DexterYan: We have `logger.SetQuiet(v.GetBool("quiet"))` in `cmd/troubleshoot/cli/root.go`, however, I didn't see any flag for quiet in support-bundle, so I have changed it to `logger.SetQuiet(!v.GetBool("debug"))`

pkg/collect/longhorn.go
DexterYan: changed.
However, when I try to write the test case for longhorn and ceph collect, I didn't any test function for ceph or longhorn Collect function. So I skipped them. I have added a test for ceph anazlyer.

pkg/analyze/ceph_test.go
```diff
@@ -5,16 +5,21 @@ import (
...
 func Test_cephStatus(t *testing.T) {
 	tests := []struct {
-		name           string
-		analyzer       troubleshootv1beta2.CephStatusAnalyze
-		expectResult   AnalyzeResult
-		filePath, file string
+		name            string
+		analyzer        troubleshootv1beta2.CephStatusAnalyze
+		expectResult    AnalyzeResult
+		expectNilResult bool
```
banjoh: Nit pick: I'd rather changing `expectResult` to a pointer and setting it to `nil` i.e

```go
[]struct {
		...
		expectResult    *AnalyzeResult
		...
}
```

then the test would be

```go
{
    name:            "pass case when get file returns not found error",
    analyzer:        troubleshootv1beta2.CephStatusAnalyze{},
    expectResult:    nil,    # set nil explicitly for ease of code reading
    getFile: func(n string) ([]byte, error) {
        return nil, &types.NotFoundError{}
    },
}

...

assert.Equal(t, test.expectResult, actual)
```

pkg/collect/longhorn.go
```diff
@@ -65,7 +65,8 @@ func (c *CollectLonghorn) Collect(progressChan chan<- interface{}) (CollectorRes
 	// collect nodes.longhorn.io
 	nodes, err := client.Nodes(ns).List(ctx, metav1.ListOptions{})
 	if err != nil {
-		return nil, errors.Wrap(err, "list nodes.longhorn.io")
+		logger.Printf("list nodes.longhorn.io not found")
```
banjoh: Not all errors will be "pod not found". There are many other [API errors](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L47) and runtime errors that can occur.

You want something like this

```go
	if err != nil {
		if apiErr, ok := err.(*apiErrors.StatusError); ok {
			if apiErr.ErrStatus.Code == http.StatusNotFound {
				logger.Printf("list nodes.longhorn.io not found")
				return nil, nil
			}
		}
		return nil, errors.Wrap(err, "list nodes.longhorn.io")
	}
```

Add this import as well
```go
	apiErrors "k8s.io/apimachinery/pkg/api/errors"
```

banjoh (commented):
All other changes look OK. I'll approve your PR once this longhorn bit is done.

pkg/analyze/ceph_test.go
DexterYan: thank you! Really good suggestion, nice and simple, I have learned from that!

pkg/collect/longhorn.go
DexterYan: updated accordingly

banjoh (approved):

Issue #906, opened 2022-12-13
Kubectl preflight --output=file doesn't send output to file


z4ce:
**Bug Description**
running `kubectl preflight --format=json --interactive=false --output=/tmp/output` does not result in the file /tmp/output being created with the output.

**Expected Behavior**

`/tmp/output` has the output from the preflight run

**Steps To Reproduce**
run:
`kubectl preflight --format=json --interactive=false --output=/tmp/output` 
notice /tmp/output is not created

**Additional Context**
```
kubectl preflight version
Replicated Preflight 0.49.1
```
MacOS Ventura

xavpaice:
verified, using the command below and v0.51.1

```
kubectl preflight --format=json --interactive=false --output=/tmp/output https://raw.githubusercontent.com/replicatedhq/troubleshoot/main/config/samples/troubleshoot_v1beta2_preflight.yaml
```

CpuID:
https://github.com/replicatedhq/troubleshoot/pull/951 fixes this 👍 

test coverage to ensure no future regressions for it are in https://github.com/replicatedhq/troubleshoot/pull/949 also

Issue #905, opened 2022-12-13
Kubectl preflight --format=yaml results in error that yaml is an unknown output format


z4ce:
**Bug Description**

Running with `kubectl preflight --format=yaml --interactive=false ...`
results in
```
name: cluster-resources    status: running         completed: 0    total: 2
name: cluster-resources    status: completed       completed: 1    total: 2
name: cluster-info         status: running         completed: 1    total: 2
name: cluster-info         status: completed       completed: 2    total: 2
Error: unknown output format: "yaml"
```
**Expected Behavior**
Output should be:
```yaml
pass:
    - title: Required Kubernetes Version
       message: Your cluster meets the recommended and required versions of Kubernetes.
```

**Steps To Reproduce**
Run the preflight plugin with `--format=yaml` worth noting that `--format=json` does work.

**Additional Context**
```
kubectl preflight version
Replicated Preflight 0.49.1
```
MacOS Ventura

xavpaice:
Verified with Troubleshoot v0.51.1 on LInux.

Command used:

```
kubectl preflight --format=yaml --interactive=false --output=/tmp/output https://raw.githubusercontent.com/replicatedhq/troubleshoot/main/config/samples/troubleshoot_v1beta2_preflight.yaml
```

CpuID:
Just a thought: do we want to maintain some consistency with how the output format is specified between tooling?

`kubectl` opts for `-o` or `--output` for `json`/`yaml` etc for basic commands like `get`.

`kubectl support-bundle analyze` also has `--output string   output format: json, yaml` which follows the above approach.

But `kubectl preflight` has 2 CLI args with different usage:

```
--format string                  output format, one of human, json, yaml. only used when interactive is set to false (default "human")
...
-o, --output string                  specify the output file path for the preflight checks
```

...as much as it would be nice to change `--format` to be `--output` for consistency, that would break some backwards compatibility? And what would we then use to supersede the old `--output`? Possibly `-f, --filename`?

CpuID:
YAML output format implemented in https://github.com/replicatedhq/troubleshoot/pull/940 - still under `--format` for now. Do we create a separate GitHub issue relating to my thoughts above about the CLI flag name...?

banjoh:
> YAML output format implemented in #940 - still under `--format` for now. Do we create a separate GitHub issue relating to my thoughts above about the CLI flag name...?

I think we can address this when looking at #933 and #934.

PR #949, opened 2023-01-09
adding test coverage for preflight.RunPreflights()

CpuID:

adding test coverage for `preflight.RunPreflights()`

[TDD](https://en.wikipedia.org/wiki/Test-driven_development) to work on https://github.com/replicatedhq/troubleshoot/issues/906 and verify the fix is successful once it's ready

this also adds coverage around https://github.com/replicatedhq/troubleshoot/issues/905 which is also beneficial

note: this currently requires a Kubernetes cluster to execute, so I've hidden it behind a build tag for local use. there's a couple of approaches to doing a mocked K8s API (eg. the `fake` packages, or a K3s in Docker etc) but both of those need further refactoring to be useful. leaving that as a future task to implement, then we can get this running as part of our GitHub Actions also...

## Checklist

- [ ] Yes
- [x] No


CpuID:
Before the fix for https://github.com/replicatedhq/troubleshoot/issues/906 is implemented:

```
❯ go test --tags=integration -run TestRunPreflights                                                                                 15s 13:40:09
--- FAIL: TestRunPreflights (10.93s)
    run_test.go:188: interactive: false format human :: Output file /var/folders/jf/wgflclmd6v97df6wvnt0twsc0000gn/T/preflight_out_test_23e4a1becfffe752b6640635086f4cf2 was expected, does not exist
    run_test.go:188: interactive: false format json :: Output file /var/folders/jf/wgflclmd6v97df6wvnt0twsc0000gn/T/preflight_out_test_17a16e135ce52f83dce3595a8b8021bb was expected, does not exist
    run_test.go:188: interactive: false format yaml :: Output file /var/folders/jf/wgflclmd6v97df6wvnt0twsc0000gn/T/preflight_out_test_22b56bafff2fcc70712ce1b101818fc7 was expected, does not exist
FAIL
exit status 1
FAIL	github.com/replicatedhq/troubleshoot/pkg/preflight	11.845s
```

CpuID:
before adding this test to the e2e tests run in the `Makefile`: https://github.com/replicatedhq/troubleshoot/actions/runs/3870599610
benchmark 5mins 47sec

CpuID:
@xavpaice gave me a few ideas which I'm trying out quickly :P to try get these tests running in our GitHub Actions quickly

CpuID:
Got it working, the test-integration job fails for the right reasons now (as #906 isn't resolved)
benchmark 6min 25sec, so it added 40~ seconds or so to do it this way...

not the end of the world, we might be able to optimize it more by combining test and test-integration but I'll leave that as a future improvement.

CpuID:
ready for review, but don't merge it until I have a fix ready for #906

pkg/preflight/troubleshoot_v1beta2_preflight_gotest.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: troubleshoot.sh/v1beta2
```
banjoh: We started storing test fixtures https://github.com/replicatedhq/troubleshoot/tree/main/testdata

pkg/preflight/run_test.go
```diff
@@ -0,0 +1,219 @@
...
+		"k8s.io/client-go/kubernetes/fake"
+		k8sruntime "k8s.io/apimachinery/pkg/runtime"
+		discoveryfake "k8s.io/client-go/discovery/fake"
+	*/)
+
+func TestRunPreflights(t *testing.T) {
+	t.Parallel()
+
+	// A very simple preflight spec (local file)
+	_, testFilename, _, status := runtime.Caller(0)
```
banjoh: I'd use https://github.com/replicatedhq/troubleshoot/blob/main/internal/testutils/utils.go#L12 instead. Feel free to add more test utility functions there.

Makefile
```diff
@@ -45,6 +45,12 @@ ffi: fmt vet
 test: generate fmt vet
 	go test ${BUILDFLAGS} ./pkg/... ./cmd/... -coverprofile cover.out
 
+# Go tests that require a K8s instance
+# TODOLATER: merge with test, so we get unified coverage reports? it'll add 21~sec to the test job though...
+.PHONY: test-integration
+test-integration:
+	go test --tags "integration exclude_graphdriver_devicemapper exclude_graphdriver_btrfs" ./pkg/... ./cmd/...
```
banjoh: It would be better to us `go test -v` for more verbose output in the CI

banjoh (commented):
This is a good start and a welcome change as well.

I highly recommend following the same table tests structure we have with other tests in the project.

pkg/preflight/run_test.go
```diff
@@ -0,0 +1,219 @@
...
+		//
+		ExpectedError      error
+		ExpectedOutputFile string
+		// May be in stdout or file, depending on above value
+		ExpectedOutputContent string
+	}
+
+	tests := []preflightTests{}
+	// TODO: test interactive true as well
+	for _, interactive := range []bool{false} {
```
banjoh: I did not review this implementation in detail in case you want to revise it to follow other table tests in the project.

CpuID:
> I highly recommend following the same table tests structure we have with other tests in the project.

thx I'll take a closer look 👍

pkg/preflight/run_test.go
CpuID: I need to pass in the actual filename, not the content of the file for test purposes here so that function won't cover it

but https://github.com/replicatedhq/troubleshoot/blob/main/internal/testutils/utils.go#L21 does save me some repetition, so I will use that :) thx

Makefile
CpuID: I didn't bother with `-v` as we weren't using it on `test`, but it probably makes a bit more sense if we're being selective with tags re what to hit... so we know what's actually "tested"

I'm really tempted to see if just starting a K3s instance in `test` is net faster CI job duration wise... in that case I might end up removing the build tag and consolidating things there, TBD and I may do it as a separate change later.

pkg/preflight/run_test.go
CpuID: yea I think I will, thx. main observations so far:

- use of `github.com/stretchr/testify/assert` for less repetition, which makes sense eg. `assert.Equal()`
- the way the structs are defined is a bit less noisy in other tests as they are written now

Makefile
CpuID: @banjoh take a look at https://github.com/replicatedhq/troubleshoot/actions/runs/3882319713/jobs/6622331875 - I feel like that's way too noisy having `-v` enabled... thoughts?

CpuID:
> ready for review, but don't merge it until I have a fix ready for https://github.com/replicatedhq/troubleshoot/issues/906

still applies...

CpuID:
https://github.com/replicatedhq/troubleshoot/pull/951 fixes https://github.com/replicatedhq/troubleshoot/issues/906 - and I confirmed (in a throwaway branch) that the tests pass with that fix too :)

pkg/preflight/run_test.go
```diff
@@ -0,0 +1,216 @@
...
+			os.Stderr = origStderr
+
+			if tt.wantErr {
+				assert.Error(t, tErr)
+			} else {
+				require.NoError(t, tErr)
+			}
+
+			useBufOut := string(bufOut[:nOut])
+			//useBufErr := string(bufErr[:nErr])
```
banjoh, line 191: You might want to clean this up

banjoh (approved):
LGTM

pkg/preflight/run_test.go
CpuID: keeping that there incase someone in future sees weird options with redirection and wonders "how can I see what's in stderr", they can just uncomment it fairly easily 👍

PR #965, opened 2023-01-13
Proposal: consolidated Troubleshoot CLI and API

### LINKED ISSUE #934: Design proposal for a consolidated Troubleshoot CLI
Create a design proposal to clarify and detail the required changes to the project to implement #933

adamancini:

WIP Proposal for a consolidated `troubleshoot` CLI/API interface

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/design/proposal-troubleshoot-consolidated-cli.md
+++ b/design/proposal-troubleshoot-consolidated-cli.md
@@ -2,18 +2,41 @@
 
 ## Goals
 
-As Troubleshoot grows and gains new features, some of which involves flags for setting options,
-and to make it easier to add additional subcommands that don't belong underneath either
-`support-bundle` or `preflight` binaries, we would like to consolidate all of the Troubleshoot
-commands under one binary/plugin.
+As Troubleshoot grows and gains new features, some of which involves flags for setting options, and to make it easier to add additional subcommands that don't belong underneath either `support-bundle` or `preflight` binaries, we would like to consolidate all of the Troubleshoot commands under one binary/plugin.
+
+There is discussion about changing the behaviour of `preflight`, considering that preflights and support-bundles utilize the same specs - the same collectors and analyzers - and only differ in what is returned to the user post-analysis.  For this design proposal, `support-bundle` and `preflight` may be condensed into a single `troubleshoot` binary/plugin.
 
 ## Non-Goals
 
+
 ## Background
+
+
+
 
 ## High Level Design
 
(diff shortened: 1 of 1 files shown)


diamonwiggins:
It would probably be helpful to mention just how similar the `collect.go` code is between `support-bundle` and `preflight`. 

https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go
https://github.com/replicatedhq/troubleshoot/blob/main/pkg/preflight/collect.go

I see that as another big benefit here as every time we make a change to the functions in one of these files, we likely have to copy/paste it to the other. Overall, may be worth detailing how much code we can get rid of by bringing `pkg/preflight` and `pkg/supportbundle` together.

design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+
+- use a spec to return a go/no-go preflight outcome
+
+  `kubectl troubleshoot --preflight spec.yaml`
+
+- use a support bundle tarball to execute `sbctl` and shell into a support bundle
+
+  `kubectl troubleshoot --shell support-bundle-12-12-2001.tar.gz`
```
diamonwiggins, line 42: This raises a good question of what should be flags, and what should be subcommands. I think `shell` as it could have additional options for itself, would need to remain its own subcommand right?

design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,56 @@
...
+
+  `kubectl troubleshoot supportbundle.yaml secrets/default/kotsadm-appslug-supportbundle`
+
+  `kubectl troubleshoot https://kots.io`
+
+
+
+- use a spec to return a go/no-go preflight outcome
+
+  `kubectl troubleshoot --preflight spec.yaml`
```
diamonwiggins, line 38: Similiar to my other comment about what should be a flag vs. subcommand, we should probably take a look at the current CLI options between the `support-bundle` and `preflight` binary to ensure nothing is radically different. I get the feeling that we should be fine here.

However, should generating a support bundle be the default behavior or should there be a flag to `--generate-support-bundle` and a preflight run(just generating analysis.json) is default behavior? I _think_ the default option should maybe be something that doesn't always generate an archive, but need to think about it more.

xavpaice:
Please see https://github.com/replicatedhq/troubleshoot/pull/1144 for an updated version (not from a fork)

PR #971, opened 2023-01-18
Standardize Cluster Resources Collector File Paths

### LINKED ISSUE #960: Standardize Cluster Resources Collector
Per the design in #893, this task is to make the following changes in preparation for integrating sbctl to the Troubleshoot repo:

* In `pkg/collect` standardize the root directory location of the Cluster Resources collector with a variable instead of repeatedly specifying it as a string - https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/cluster_resources.go#L121.
* Import the directory the Cluster Resources collector uses from `pkg/collect` for usage in the APIs - https://github.com/replicatedhq/sbctl/blob/main/pkg/api/server.go#L220. This is an example of the way we can start making the sbctl code more efficient with this change.

Definition of done:

* [ ] clusterResources specifies the root directory as a var in pkg/collect (use package `constants` for the source)
* [ ] sbctl pkg/api/server.go imports Troubleshoot's constants package and uses that to determine the file locations

diamonwiggins:

Setting constant values for cluster resources directories and subdirectories, so they can easily be reused

## Checklist

- [ ] Yes
- [x] No


diamonwiggins:
A couple questions:

1. It felt like I'd expect to find these constants a bit closer to the actual code and not in the separate `pkg/constants.go`? Curious on other's thoughts. I don't have too strong an opinion on it.
2. Per #960 the only requirement here was cluster resources' root directory, but I ended up adding the others as well to see if that would clean up the code in other places. Worth keeping around?

CpuID:
> **It felt like I'd expect to find these constants a bit closer to the actual code and not in the separate pkg/constants.go? Curious on other's thoughts. I don't have too strong an opinion on it.

I tend to agree, for this kind of thing closer to the code itself (for easier readability/trying to grok what it's doing etc). If the constants were being used in a whole bunch of files, having them in `constants.go` might be more reasonable, but that could be an easy future optimization as necessary.

> Per Standardize Cluster Resources Collector #960 the only requirement here was cluster resources' root directory, but I ended up adding the others as well to see if that would clean up the code in other places. Worth keeping around?

deferring to others for opinions 👍

DexterYan:
I am wondering is it worthy to create a cluster_constants.go under /pkg/constants folder to store them? In the meanwhile, since those constants are not like default configuration constants which can be put close to the code or may not in constants.go

diamonwiggins:
> I am wondering is it worthy to create a cluster_constants.go under /pkg/constants folder to store them? In the meanwhile, since those constants are not like default configuration constants which can be put close to the code or may not in constants.go

makes sense to me

adamancini (commented):
lgtm

banjoh:
> It felt like I'd expect to find these constants a bit closer to the actual code and not in the separate pkg/constants.go? Curious on other's thoughts. I don't have too strong an opinion on it.

The reason for creating the `constants` package is to avoid circular dependencies cause constants end up being sprawled all over the codebase.

diamonwiggins:
> > It felt like I'd expect to find these constants a bit closer to the actual code and not in the separate pkg/constants.go? Curious on other's thoughts. I don't have too strong an opinion on it.
> 
> The reason for creating the `constants` package is to avoid circular dependencies cause constants end up being sprawled all over the codebase.

didn't think about the circular dependency bit. good call out

banjoh (approved):
LGTM

PR #972, opened 2023-01-19
Deduplication for In-Cluster Collectors

### LINKED ISSUE #915: Generic Collector Deduplication
Although some collectors will have special logic for dedup/merge similar to the [Cluster Resources Collector](https://github.com/replicatedhq/troubleshoot/pull/832), most every collector spec will benefit from a basic dedup that checks to ensure that the same entry isn't added twice. This is critical as the recently added features to discover SupportBundle specs within Secrets/Configmaps in the cluster become more utilized.

**Describe the feature**
If the below spec is given as input twice, it should only run once.

```
kind: SupportBundle
metadata:
  name: bundle1
spec:
  collectors:
    - clusterInfo: {}
    - http:
        name: healthz
        get:
          url: http://api:3000/healthz
    - data:
        collectorName: my-password-dump
        name: data
        data: |
          my super secret password is abc123
          another redaction will go here
    - runPod:
        name: "run-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
```

This change should be made for both the Support Bundle and Preflight CLIs

`pkg/supportbundle`
As desired collectors are being added in [runCollectors](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go#L88-L99) and [runHostCollectors](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/collect.go#L28-L34) ensure the destination list doesn't have duplicates

`pkg/preflight`
Same as above in [CollectHost](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/preflight/collect.go#L89-L95) and [Collect](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/preflight/collect.go#L142-L153)

**Additional context**

diamonwiggins:

Introduces a function which can be used to do a simple dedup on a list of `troubleshootv1beta2.Collect` objects. Because no sorting of keys occurs in the collect object's spec, only exact matches will be deduped.

This spec will now be deduped, so that it only runs the `runPod` collector once.
```yaml
kind: SupportBundle
metadata:
  name: bundle1
spec:
  collectors:
    - runPod:
        name: "runpod-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
    - runPod:
        name: "runpod-ping"
        namespace: default
        podSpec: 
          containers:
          - name: run-ping
            image: busybox:1
            command: ["ping"]
```

## Checklist

- [ ] Yes
- [x] No


pkg/collect/collector.go
```diff
@@ -189,3 +190,49 @@ func getCollectorName(c interface{}) string {
...
+
+// deduplicates a list of troubleshootv1beta2.Collect objects
+// marshals object to json and then uses its string value to check for uniqueness
+// there is no sorting of the keys in the collect object's spec so if the spec isn't an exact match line for line as written, no dedup will occur
+func DedupCollectors(allCollectors []*troubleshootv1beta2.Collect) []*troubleshootv1beta2.Collect {
+	uniqueCollectors := make(map[string]bool)
+	finalCollectors := []*troubleshootv1beta2.Collect{}
+
+	for _, collector := range allCollectors {
+		data, _ := json.Marshal(collector)
```
banjoh: I think we should
* Check the returned error. If there is an error we perhaps should not remove the collector and let the code downstream handle it. We may have some wise person crafting a collector like below (not a good idea). You never know 😄 - defensive coding 🦾  
```go
type CollectG struct {
	Test     string
	Callback chan int
}
```
*  Check for a `nil` collector since you are dealing with pointers. A `nil` value will panic the application

pkg/collect/collector_test.go
```diff
@@ -364,3 +365,83 @@ pwd=somethinggoeshere;`,
...
+				},
+			},
+			want: []*troubleshootv1beta2.Collect{
+				{
+					ClusterInfo: &troubleshootv1beta2.ClusterInfo{},
+				},
+			},
+		},
+		{
+			name: "multiple cluster resources with matching namespace lists",
```
banjoh, line 392: Just a comment. No change required.

If you reverse the list of namespaces to `[]string{"namespace2", "namespace1"}` then the specs are different, but I guess that's OK. Its better having some duplicates compared to removing a collector that shouldn't have been removed.

pkg/collect/collector_test.go
```diff
@@ -364,3 +365,83 @@ pwd=somethinggoeshere;`,
...
 	}
 }
+
+func TestCollector_DedupCollectors(t *testing.T) {
+	tests := []struct {
+		name       string
+		Collectors []*troubleshootv1beta2.Collect
+		want       []*troubleshootv1beta2.Collect
+	}{
+		{
```
banjoh, line 375: Maybe add a test case of our fictional collector that does not marshal to json?

pkg/collect/collector.go
diamonwiggins: @banjoh where are you saying I need to check for nil here? with the `json.Marshal`? I thought I had read it handled nil pointers already, but could be wrong. A couldn't get a quick test to panic either.

pkg/collect/collector_test.go
diamonwiggins: right, yeah eventually we could do sorting of fields to garuantee this, but not going to address that with this PR

pkg/collect/collector_test.go
diamonwiggins: You have a recommendation on the "fictional collector" would look like since I'm using `troubleshootv1beta2.Collect` in the test?

pkg/collect/collector.go
banjoh: > ```go
> type CollectG struct {
> 	Test     string
> 	Callback chan int
> }
> ```

You are right. `troubleshootv1beta2.Collect` type will always be valid JSON cause we generate GVK types from these structs that should always marshal/unmarshal successfully.

As for failing to unmarshal, https://goplay.tools/snippet/UcYIpk7HQLn is an example

banjoh (approved):

PR #974, opened 2023-01-19
WIP - add proposal for utilising cluster-resoruces abstraction in existing analyzers. 

danj-replicated:
This PR introduces a new design proposal for utilising the recently merged cluster-resoruces abstraction in exinsting analysers.


design/proposal-enable-cluster-resorces-abstraction-for-existing-analyzers.md
```diff
@@ -0,0 +1,87 @@
...
+
+Currently a user wanting to analyze a resource collected by the cluster-resources collector has to manually specify the file location in one of the existing collectors.
+
+## High-Level Design
+
+This proposal suggests adding a new field to the collectors that need easy access to cluster-resource files; such as the text analyzer, jsonpath analyzer or yaml analizer to utilize the recently merged FindResource() function.
+
+## Detailed Design
+
+currently FindResource() returns an interface{}, but it would be trivial, since it's currently only implimented in an example collector, to change it to return a `[]byte` or `map[string][]byte` and use it as a conditional drop-in for os.ReadFile() or similar.
```
chris-sanders, line 21: Can/should we just strive to implement the [Reader Interface](https://github.com/golang/go/blob/2ebe77a2fda1ee9ff6fd9a3e08933ad1ebaea039/src/io/io.go#L83) so it's a drop in replacement? I think that's what the `[]byte` suggestion is but just being explicit about the interface would be good.

chris-sanders: I'm unclear what we would do if multiple files match, maybe that's not an issue today as we're trying to match a single file on disk and multiple would be an error?

chris-sanders (commented):
I like where this is going, one clarification / question in-line.

design/proposal-enable-cluster-resorces-abstraction-for-existing-analyzers.md
danj-replicated: yes good point. that would actually simplify the FindResource() code a bit as well, currently it inherits the file opener of the analyzer calling it. 

Allowing it top open it's own files and implimenting Read() would be a decent improvement.

danj-replicated: re: multiple files, I'm not sure multple _files_ matching is going to be a problem, multiple resources in one file might be though.

PR #989, opened 2023-01-30
feat(spec): get support bundle secret matching label selector

DexterYan:

- get support bundle secret matching label selector

feature: [sc-65659](https://app.shortcut.com/replicated/story/65659/troubleshoot-call-troubleshoot-to-search-for-labels-rather-than-provide-secret-uri)

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/specs/secrets.go
+++ b/pkg/specs/secrets.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/troubleshoot/pkg/k8sutil"
+	corev1 "k8s.io/api/core/v1"
+	kuberneteserrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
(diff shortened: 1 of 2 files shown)


banjoh (commented):
This is my understanding of what we want to achieve.

Have a client of troubleshoot (KOTS in this case) discover support bundle specs with labels (default `troubleshoot.io/kind=support-bundle`) and merge a downloaded spec from a defined `uri` to these discovered specs. For cli invocations, we by default discover the specs, download specs from defined `uri` links, and merge them together. I'm I correct?

I'm probably missing the big picture (cause I've not looked at how KOTS is using troubleshoot), but I'm failing to see how these changes are related to the problem we are trying to address.

This [code](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L140-L173) is where the cli implements part of the logic above.

pkg/specs/secrets.go
```diff
@@ -51,3 +53,21 @@ func LoadFromSecretMatchingLabel(client kubernetes.Interface, labelSelector stri
...
+	if err != nil {
+		return nil, err
+	}
+
+	for _, secret := range secrets.Items {
+		_, ok := secret.Data[key]
+		if !ok {
+			continue
+		}
+		return &secret, nil
```
diamonwiggins, line 69: How is this function intended to be used? This will just return the first secret found matching the label correct? Is that what we intend?

diamonwiggins:
I guess I'd also be curious if there's a way to get consistency here between KOTS and the Troubleshoot CLI, so that we don't need both `GetSecretMatchingLabel` and `LoadFromSecretMatchingLabel`. The intention was that the latter would also be used in KOTS, but I didn't do a good job of verifying that it would be able to be used there 😅 .

banjoh:
> I'm probably missing the big picture (cause I've not looked at how KOTS is using troubleshoot), but I'm failing to see how these changes are related to the problem we are trying to address.

Looking at https://github.com/replicatedhq/kots/pull/3548/files PR, I think I see what the intention is here. KOTS would select ONE secret it needs to add the support bundle label to (retroactively I believe?). The question here is how accurate filtering would be. But I guess that can be left to the caller. If this is the intention, I think this code belongs in KOTS. Its up to the application to decide how and which secret to attach a label. The logic isn't very "troubelshoot" centric in any case.

DexterYan:
After discussing in slack channel , the current solution to avoid the KOTS select multiple secrets which has more than one app installed in the same cluster, is to add two more labels
```
kots.io/app: <app-slug>
troubleshoot.io/kind=kurl
```
I will update the pr to reflect it. However, seems the function is more stick to kots beheviour, which maybe is not suitable to add it in troubleshoot. I will figurate it out.

PR #996, opened 2023-02-01
 feat: add the option to provide notes and update ceph analyser to use it

camilamacedo86:

This PR adds the option to introduce note info for the outcomes.
Also, we are updating the ceph analyser to use it to address the use case which motivated this change and can be used to verify the implementation.

### Use case:

If we check that the ceph status is NOT OK we recommend users check the link https://rook.io/docs/rook/v1.4/ceph-common-issues.html.

However, we need to clarify to users that the page should be checked according to the rook version installed 
on the host. Therefore, we are updating it to use the latest page instead of an old version and adding the note for URI. 

See how it is today before the change:

Screenshot 2023-02-01 at 08 42 17

- [ ] Yes
- [x] No


banjoh (commented):
Instead of adding a new optional field to all analysers, can't we just use the current `message` field?

If its a formatting issue in the TUI, I think we should address that instead.

camilamacedo86:
Hi @banjoh,

Thank you for check this one:

> Instead of adding a new optional field to all analysers, can't we just use the current message field?

The problem would be that would not provide a good result in the UI and would bring confusion.
See in the example scenario that ideally we would like to add observations/notes at the bottom. 

However, I will check if for this specific case we cannot add the note in the URI field instead of create a new one.

camilamacedo86:
Closing this one since it would change the API

Issue #1001, opened 2023-02-01
Support Bundle concat code should be consistent with Preflight concat code


diamonwiggins:
**Describe the rationale for the suggested feature.**

A recent bug was discovered in the concat code of Preflight - https://github.com/replicatedhq/troubleshoot/pull/998. Although there isn't a known issue in the support bundle code, we should update it for consistency.

https://github.com/replicatedhq/troubleshoot/blob/main/pkg/supportbundle/supportbundle.go#L237
https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L101-L105

**Describe the feature**
Refactor the Support Bundle Concat logic to be consistent with the changes in #998 

**Describe alternatives you've considered**

Describe alternative solutions here. Include any workarounds you've considered.

**Additional context**

Add additional context about the feature request.  If the change is substantial, consider attaching files to the issue  outlining architectural changes, data flows, file formats etc., anything that helps describe the requested change.

yangmeilly:
this issue has been annihilated by https://github.com/replicatedhq/troubleshoot/pull/1002

xavpaice:
#933 should resolve this as various parts of the codebase are consolidated.  The main issue here is about duplicating the concat code between both `preflight` and `support-bundle`, but that duplication sits in many places in the codebase.

I'll close this since we have #1002 handling the specific issue, and #933 to tidy up in general.

PR #998, opened 2023-02-01
fix: address runtime error of nil pointer about concatenating preflig…

yangmeilly:
…ht spec with hostpreflight spec in preflight run.go

## Description, Motivation and Context

This pr is intend to fix a nil pointer error about concatenating spec in preflight module.

**Steps To Reproduce bug**
 **cmd : preflight examples/preflight/host/cpu.yaml  examples/preflight/node-resources.yaml**

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x120 pc=0x10268e9bc]

goroutine 1 [running]:
github.com/replicatedhq/troubleshoot/pkg/preflight.ConcatPreflightSpec(0x0, 0x14000918180)
        /Users/yangmeili/work/tmp/troubleshoot/pkg/preflight/concat.go:9 +0x5c
       ... (omitting other error stacks )

## Checklist

- [x] New and existing tests pass locally with introduced changes.
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] The commit message(s) are informative and highlight any breaking changes
- [ ] Any documentation required has been added/updated. For changes to https://troubleshoot.sh/ create a PR [here](https://github.com/replicatedhq/troubleshoot.sh/pulls)

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #1002, opened 2023-02-02
refactor: keep support bundle concat logic to be consistent with Preflight concat

yangmeilly:

This PR is set to refactor the Support Bundle Concat logic to be consistent with the changes which commiteed in  https://github.com/replicatedhq/troubleshoot/pull/998 about Preflight concat 

more information reference to https://github.com/replicatedhq/troubleshoot/issues/1001

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/supportbundle/supportbundle_test.go
+++ b/pkg/supportbundle/supportbundle_test.go
@@ -3,6 +3,8 @@ package supportbundle
 import (
 	"reflect"
 	"testing"
+
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 )
 
 func Test_LoadAndConcatSpec(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/supportbundle/supportbundle_test.go
```diff
@@ -42,5 +44,15 @@ func Test_LoadAndConcatSpec(t *testing.T) {
 	if reflect.DeepEqual(fullbundle, bundle3) == false {
 		t.Error("Full bundle and concatenated bundle are not the same.")
 	}
+	// add nil pointer test case
```
banjoh: +1 for adding test coverage

* I'd extract this into a separate test function `Test_LoadAndConcatSpec_WithNil` for example
* Also add a case where both function arguments are `nil` to ensure it test no panic is caused. The current implementation handles that well

yangmeilly: copy that, and changes in https://github.com/replicatedhq/troubleshoot/pull/1002/commits/c7cf8f6649142835600a96985fe7b404d6fd9968

banjoh (approved):

PR #1003, opened 2023-02-02
CONTRIBUTING.md: adding a quick copy paste local dev setup

CpuID:

adding a quick copy paste local dev setup 

uses Docker/colima with local K8s

rationale: I want to be able to code in airgap environments (eg. on planes), and/or with ~0 latency, and not rely on VMs or remote K8s clusters etc :)

open to suggestions re different approaches or shuffling the logic I've documented here to places like the `Makefile` etc

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,42 @@ To get started we recommend:
 
 6. Install [golangci-lint] linter and run `make lint` to execute additional code linters.
 
+### Quick Start Local Development (MacOS)
+
+First install [colima](https://github.com/abiosoft/colima):
+
+```
+brew install colima
+```
+
+Start a colima VM with a single node Kubernetes cluster:
+
+```
+colima start --kubernetes
+```
+
+Then spawn a Docker container with a Go toolchain, with access to the Kubernetes cluster:
+
+```
+docker run -it --rm --network host -v $(pwd):/src -v $HOME/.kube:/root/.kube golang:1.19.5-bullseye bash
+```
+
+A quick test (in the Docker container):
+
+```
+make preflight support-bundle
+bin/preflight config/samples/troubleshoot_v1beta2_preflight.yaml
+bin/preflight examples/preflight/sample-preflight.yaml
+```
+
+If you need `kubectl` in the Docker container for diagnostics:
+
+```
+curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+rm -f kubectl
+```
+
 ### Testing
 
 To run the tests locally run the following:


banjoh:
Personally, I usually use [docker desktop](https://docs.docker.com/desktop/install/mac-install/) and run [k3d](https://k3d.io/v5.4.6/) to create a k8s cluster. I've not used `calima` before.

As for the dev environment steps you documented, they look quite involving for someone new to the project. Perhaps we can have the documentation be something shorter like

* Launch a k8s cluster. We use `calima` or `docker desktop` + `k3d`.
* Ensure you can `kubectl` to your cluster
* Run troubleshoot like this... (some commands)

CONTRIBUTING.md
```diff
@@ -32,6 +32,42 @@ To get started we recommend:
 
 6. Install [golangci-lint] linter and run `make lint` to execute additional code linters.
 
+### Quick Start Local Development (MacOS)
+
+First install [colima](https://github.com/abiosoft/colima):
```
chris-sanders, line 37: I'm not sure if we should be solving the "get a local kubernetes" problem in the contributing documentation for troubleshoot. I'm also not familiar with this project and I would be curious if we were to make recommendations down to this level why we pick this over the dozens of other options. In particular, it would seem [kind](https://kind.sigs.k8s.io/) being an upstream sig is a fairly strong option.

CONTRIBUTING.md
```diff
@@ -32,6 +32,42 @@ To get started we recommend:
...
+Start a colima VM with a single node Kubernetes cluster:
+
+```
+colima start --kubernetes
+```
+
+Then spawn a Docker container with a Go toolchain, with access to the Kubernetes cluster:
+
+```
+docker run -it --rm --network host -v $(pwd):/src -v $HOME/.kube:/root/.kube golang:1.19.5-bullseye bash
```
chris-sanders, line 52: Can this somehow be tied to the src tree so we don't have a chance of the go version drifting from what's actually in troubelshoot?

CONTRIBUTING.md
```diff
@@ -32,6 +32,42 @@ To get started we recommend:
...
+```
+
+A quick test (in the Docker container):
+
+```
+make preflight support-bundle
+bin/preflight config/samples/troubleshoot_v1beta2_preflight.yaml
+bin/preflight examples/preflight/sample-preflight.yaml
```
chris-sanders, line 60: I'd say an example of each of the binaries would be good. I'm guessing support-bundle works too?

chris-sanders (changes requested):
I like making things very straight forward, I think we need a proposal doc and discussion before picking a recommended test cluster if we decide to do that. I don't want to get in the business of telling people how to run a cluster so they can work on troubleshoot and if we are going to help people onboard at that level it should be carefully considered which method the project is going to offer advice on.

CONTRIBUTING.md
CpuID: good idea

CONTRIBUTING.md
CpuID: > Personally, I usually use [docker desktop](https://docs.docker.com/desktop/install/mac-install/) and run [k3d](https://k3d.io/v5.4.6/) to create a k8s cluster. I've not used `calima` before.
> 
> As for the dev environment steps you documented, they look quite involving for someone new to the project. Perhaps we can have the documentation be something shorter like
> 
>     * Launch a k8s cluster. We use `calima` or `docker desktop` + `k3d`.
> 
>     * Ensure you can `kubectl` to your cluster
> 
>     * Run troubleshoot like this... (some commands)

@banjoh I'd love to have a version in the docs that's executable/copy paste from "nothing to workable dev env", rather than having to interpret steps and go through some kind of setup more than that.

how we get there though, I'm completely open to feedback :)

CpuID:
> I like making things very straight forward, I think we need a proposal doc and discussion before picking a recommended test cluster if we decide to do that. I don't want to get in the business of telling people how to run a cluster so they can work on troubleshoot and if we are going to help people onboard at that level it should be carefully considered which method the project is going to offer advice on.

@chris-sanders yea I agree, I'm fine with a proposal and discussion about the "how" we do it rather than this approach specifically, for sure 👍 this was the approach I took so far, but definitely open to hearing how everyone else is approaching it

xavpaice:
stale PR, has merge conflicts.  I'll close this now since we're unlikely to merge it.

PR #1008, opened 2023-02-08
feat: use klog as the default logging library

### LINKED ISSUE #922: Add meaningful logging to library and CLI applications

When we run `preflight` or `support-bundle` (or, indeed, the other CLI components) there is a `--debug` switch, but it doesn't appear to do much.

Several times lately I have wanted to review what the process is doing, and where it is failing, but do not have the information without inserting print statements.

**Describe the feature**

When we specify `--debug` on the CLI, provide at least:
* A log message in stdout for each collector, analyzer and redactor running with start/end times
* same messages added to a log file in the support bundle
* a copy of the support-bundle spec that ran (post-merge) included in the support-bundle.  This may require redactors to run over the output.

Describe alternative solutions here. Include any workarounds you've considered.

TBA

**Additional context**

banjoh:

Use https://github.com/kubernetes/klog as the default logging library because of the following
- Allows users to pass verbosity levels using `-v` flag e.g `-v=2`. This allows us to instrument the library with logs based on importance. Less important log messages would have higher `-v` values where as more important messages would have lower`-v` values. Refer to [klog conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use) for more on verbosity levels
- Increase logging in the project
- Document logging conventions for the project (to be done in README docs as part of https://github.com/replicatedhq/troubleshoot/issues/1031)

Fixes: #922

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/cmd/troubleshoot/cli/root.go
+++ b/cmd/troubleshoot/cli/root.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"strings"
 
-	"github.com/go-logr/logr"
 	"github.com/replicatedhq/troubleshoot/cmd/util"
 	"github.com/replicatedhq/troubleshoot/internal/traces"
 	"github.com/replicatedhq/troubleshoot/pkg/k8sutil"
@@ -28,16 +27,11 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 			v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 			v.BindPFlags(cmd.Flags())
 
+			logger.SetupLogger(v)
+
 			if err := util.StartProfiling(); err != nil {
-				logger.Printf("Failed to start profiling: %v", err)
+				klog.Errorf("Failed to start profiling: %v", err)
 			}
-		},
-		PreRun: func(cmd *cobra.Command, args []string) {
-			v := viper.GetViper()
-			if !v.GetBool("debug") {
-				klog.SetLogger(logr.Discard())
-			}
-			logger.SetQuiet(!v.GetBool("debug"))
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v := viper.GetViper()
@@ -45,13 +39,13 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 			closer, err := traces.ConfigureTracing("support-bundle")
 			if err != nil {
 				// Do not fail running support-bundle if tracing fails
-				logger.Printf("Failed to initialize open tracing provider: %v", err)
+				klog.Errorf("Failed to initialize open tracing provider: %v", err)
 			} else {
 				defer closer()
 			}
 
 			err = runTroubleshoot(v, args)
-			if v.GetBool("debug") {
+			if v.GetBool("debug") || v.IsSet("v") {
 				fmt.Printf("\n%s", traces.GetExporterInstance().GetSummary())
 			}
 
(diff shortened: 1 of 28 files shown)


banjoh:
Left in draft as we wait for https://github.com/kubernetes/klog/pull/363#issuecomment-1422893691 to be addressed. Once that's done, I can add the rest of the implementation.

banjoh:
> 

Unblocked by splitting logs storage in suppor bundles to a separate issue (https://github.com/replicatedhq/troubleshoot/issues/1028)

pkg/collect/longhorn.go
```diff
@@ -69,8 +69,8 @@ func (c *CollectLonghorn) Collect(progressChan chan<- interface{}) (CollectorRes
 	if err != nil {
 		if apiErr, ok := err.(*apiErrors.StatusError); ok {
 			if apiErr.ErrStatus.Code == http.StatusNotFound {
-				logger.Printf("list nodes.longhorn.io not found")
-				return nil, nil
+				klog.Error("list nodes.longhorn.io not found")
+				return NewResult(), nil
```
banjoh, line 73: Returning `nil` here can lead to a panic since there is no error returned. Fixed when spotted.

adamancini:
@banjoh nice, taking a look

pkg/logger/logger.go
```diff
@@ -1,26 +1,59 @@
...
 
-func Printf(format string, args ...interface{}) {
+// SetQuiet enables or disables klog logger.
+func SetQuiet(quiet bool) {
 	if quiet {
-		return
+		klog.SetLogger(logr.Discard())
+	} else {
+		// Restore the default logger
+		klog.ClearLogger()
```
DexterYan, line 63: do we need to add `klog.ClearLogger()`? I saw SetupLogger has been used in all the root.go without any SetLogger before it if we are not use debug or v flag.

If we are going to use it not just in initial stage function, 
then I saw one notice in the klog docs about [ClearLogger](https://pkg.go.dev/k8s.io/klog/v2#ClearLogger) 
```
Modifying the logger is not thread-safe and should be done while no other goroutines invoke log calls, usually during program initialization.
```

banjoh: Good callout on thread safety. I'll update it.

> do we need to add klog.ClearLogger()? I saw SetupLogger has been used in all the root.go without any SetLogger before it if we are not use debug or v flag.

We do not need to `SetLogger()` before calling `ClearLogger()`. Think of `klog.ClearLogger()` as a way to restore klog to its defaults. I think this function should have been `klog.RestoreDefaultLogger()` or something similar.

DexterYan (approved):

cmd/troubleshoot/cli/root.go
```diff
@@ -79,7 +73,7 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 	cmd.Flags().String("since-time", "", "force pod logs collectors to return logs after a specific date (RFC3339)")
 	cmd.Flags().String("since", "", "force pod logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.")
 	cmd.Flags().StringP("output", "o", "", "specify the output file path for the support bundle")
-	cmd.Flags().Bool("debug", false, "enable debug logging")
+	cmd.Flags().Bool("debug", false, "enable debug logging. This is equivalent to --v=0")
```
diamonwiggins, line 76: I might just be confused, but from the comments you have in logger.go, 0 is least verbose and 2 would be most verbose correct? But here debug logging being enabled would be `-v=0`? Or am i incorrect and 0 is most verbose?

banjoh: That's correct. I wanted when a user provides `--debug` option, some logs are spit out but not at the highest log level. I believe users usually want to see some logs, but not everything. My 2 cnts :)

diamonwiggins: Gotcha

diamonwiggins (approved):

Issue #762, opened 2022-10-07
Collector & Analyzer for certificates


xavpaice:
**Describe the rationale for the suggested feature.**

Several support cases have come through where certificates have expired, we would like to be able to explain this at the analyzer stage so folks know immediately what to do about it.

**Describe the feature**

Ensure that the certificate expiry information is available in support bundles for:
* Kubernetes
* weave
* kURL registry
* Envoy

Create an analyzer that will highlight expired certificates from any of the above, along with a URL that will show instructions for remedy.

**Describe alternatives you've considered**

.

**Additional context**

Related Shortcut: https://app.shortcut.com/replicated/story/42738/add-collectors-and-analyzers-for-expired-certificates

Envoy: https://community.replicated.com/t/kurl-how-to-manually-rotate-an-expired-certificate-for-envoy/889

kURL Registry: https://community.replicated.com/t/what-to-do-if-the-kurl-registry-certificates-have-expired/955

Contour: https://community.replicated.com/t/kots-is-it-possible-to-configure-tls-certs-for-contour-in-the-embedded-installation/390/2

Checking the Kubernetes certs themselves is more difficult as it requires `sudo kubeadm certs check-expiration` which requires sudo.

**Definition of done**

- [x] Write a spec for the collectors and analysers - Done [here](https://github.com/replicatedhq/troubleshoot/issues/762#issuecomment-1285916327)
- [x] Develop an in-cluster collector
  - [x] Certs from secrets
  - [x] Certs from config maps
  - [ ] Certs from endpoints
  - [x] In-cluster kubernetes cluster certs - most are in `kube-*` config maps
- [x] Develop a host collector
  - [x] Kubernetes cluster certs from host paths i.e `/etc/kubernetes/pki`, `/var/lib/kubelet/pki` on a kubeadm-based cluster
  - [ ] Certs from endpoints
  - [x] Certs from a list of paths in the host e.g `/etc/myservice/*.crt`
- [x] Develop an analyser - reviews expiry amongst other things
- [ ] The default specs we provide in [troubleshoot-specs](https://github.com/replicatedhq/troubleshoot-specs) including analyzers for the above certificates with links to docs to fix them. Collect the following
  - [ ] Kubernetes (host & in-cluster collectors)
  - [ ] kURL registry
  - [ ] Contour/envoy
  - [ ]  KOTS Admin Console (kotsadm-tls)
[...]

PR #1034, opened 2023-02-23
incluster certificate collector

drohnow:
## Description: 
The is a new incluster certificate collector that collects certificate information from a named secret. This will allow the end-user to easily view important information about the SSL certificate their application depends on to function.  This collector could be helpful in easily surfacing information about the certificate during the troubleshooting process.

Summary of Change:  
Added a new file: /Users/davidr/go/src/github.com/drohnow/troubleshoot/pkg/collect/incluster-cert-collector.go
Added collector information to the following files:
/Users/davidr/go/src/github.com/drohnow/troubleshoot/pkg/collect/collector.go
/Users/davidr/go/src/github.com/drohnow/troubleshoot/pkg/apis/troubleshoot/v1beta2/collector_shared.go

This change is relevant to issue [762](https://github.com/replicatedhq/troubleshoot/issues/762)

## Checklist

- [x] New and existing tests pass locally with introduced changes 

- [ ] Yes
- [x] No


banjoh:
To ease testing this change, could you provide a sample support bundle spec similar to below?

```yaml
kind: SupportBundle
metadata:
  name: example
spec:
  collectors:
    - <collector>:... # Here is your collector spec
```

I'd like something I can run with `./support-bundle spec.yaml`

drohnow:
> To ease testing this change, could you provide a sample support bundle spec similar to below?
> 
> ```yaml
> kind: SupportBundle
> metadata:
>   name: example
> spec:
>   collectors:
>     - <collector>:... # Here is your collector spec
> ```
> 
> I'd like something I can run with `./support-bundle spec.yaml`

@banjoh here you go:
https://raw.githubusercontent.com/drohnow/troubleshoot-specs/main/in-cluster/inclustercertcollector.yaml

banjoh:
> @banjoh here you go: https://raw.githubusercontent.com/drohnow/troubleshoot-specs/main/in-cluster/inclustercertcollector.yaml

Pasting it here for posterity (your branch can vanish)

```yaml
kind: SupportBundle
metadata:
  name: default
spec:
  collectors:
    - inclustercertificate:
        name: registry-pki  #secretname where certificate exists
```

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+	Collector    *troubleshootv1beta2.InclusterCertificate
+	BundlePath   string
+	Namespace    string
+	ClientConfig *rest.Config
+	Client       kubernetes.Interface
+	Context      context.Context
+	RBACErrors
+}
+
+type Secret struct {
```
banjoh: Can we have the json key names in lower case or camel case?

`secret Name` => `name`

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+	Name      string `json:"secret Name,omitempty"`
+	Namespace string `json:"secret Namespace,omitempty"`
```
banjoh: `secret Namespace` -> `namespace`

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// Certificate Struct
+type ParsedCertificate struct {
+	CertName         string    `json:"Certificate Name"`
+	DNSNames         []string  `json:"DNS Names"`
+	IssuerCommonName string    `json:"Issuer"`
+	Organizations    []string  `json:"Issuer Organizations"`
+	CertDate         time.Time `json:"Certificate Expiration Date"`
```
banjoh: Can we have both dates here?

"notBefore": "2021-03-04T18:00:00Z",
"notAfter": "2021-06-02T18:00:00Z",

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+	Context      context.Context
+	RBACErrors
+}
+
+type Secret struct {
+	Name      string `json:"secret Name,omitempty"`
+	Namespace string `json:"secret Namespace,omitempty"`
+}
+
+// Certificate Struct
```
banjoh: Lower case/camel case without spaces is the way other collectors do it

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+type ParsedCertificate struct {
+	CertName         string    `json:"Certificate Name"`
+	DNSNames         []string  `json:"DNS Names"`
```
banjoh: Nit pick: I'd change this to `subjectAlternativeNames` instead as the X509 spec has it.

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+
+					data := string(cert)
+					var block *pem.Block
+
+					block, _ = pem.Decode([]byte(data))
+
+					//parsed SSL certificate
+					parsedCert, errParse := x509.ParseCertificate(block.Bytes)
+					if errParse != nil {
+						log.Println(errParse)
```
banjoh: I think we want to collect errors as well. Lets store this in an `errors` object in the json file

How about having the structure of the collected certificates look like this? Note that the certificate can be a chain.

```json
    {
        "source": {
                "secret": "<name of secret>"
        },
        "errors": [],          // List of errors that occurred when running the collector e.g host path not found, connection errors...
        "certificateChain" [   // Chain of certificates in order of their depth. The first one (0) should be the leaf cert.
            {
                "version": "",
                "serialNumber": "",
                "subject": "CN = registry.kurl.svc.cluster.local",
                "issuer": "CN = kubernetes",
                "notBefore": "2021-03-04T18:00:00Z",
                "notAfter": "2021-06-02T18:00:00Z",
                "isCA": false,
                "isValid": true,
                "subjectAlternativeNames": ["name1", "name2", "name3"],
            },
            ....
        ]
    },
```

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+type Secret struct {
+	Name      string `json:"secret Name,omitempty"`
+	Namespace string `json:"secret Namespace,omitempty"`
+}
+
+// Certificate Struct
+type ParsedCertificate struct {
+	CertName         string    `json:"Certificate Name"`
+	DNSNames         []string  `json:"DNS Names"`
+	IssuerCommonName string    `json:"Issuer"`
```
banjoh: Why not just collect the raw subject line? i.e the entire like `Subject: O=k3s, CN=k3s`

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+func (c *CollectInclusterCertificate) Collect(progressChan chan<- interface{}) (CollectorResult, error) {
+
+	output := NewResult()
+	// Json object initilization - start
+	var certInfo []ParsedCertificate
+	var certJson = []byte("[]")
+	errJson := json.Unmarshal(certJson, &certInfo)
+	if errJson != nil {
+		return nil, errors.Wrap(errJson, "failed to umarshal Json")
+	} // Json object initilization - end
```
banjoh: What's happening here? I'm not sure I follow the purpose of this bit

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+
+	certificate := SecretCertCollector(c.Collector.Name, c.Client)
+
+	results := certificate
```
banjoh: I believe we can do away with this line.

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+
+func SecretCertCollector(secretName string, client kubernetes.Interface) []byte {
+
+	currentTime := time.Now()
+	var certInfo []ParsedCertificate
+	var certJson = []byte("[]")
+	err := json.Unmarshal(certJson, &certInfo)
+	if err != nil {
+		log.Println(err)
+	}
```
banjoh: Are these lines needed?

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+
+	listOptions := metav1.ListOptions{}
+
+	secrets, _ := client.CoreV1().Secrets("").List(context.Background(), listOptions)
```
banjoh: How about the namespace? I think this is where we need to use it.

banjoh (changes requested):

pkg/collect/incluster-cert-collector.go
```diff
@@ -0,0 +1,124 @@
...
+
+	for _, secret := range secrets.Items {
+		if secretName == secret.Name {
+
+			for certName, cert := range secret.Data {
+				if certName[len(certName)-3:] == "crt" {
```
banjoh: We could ignore the key here and always parse the data (`cert`) to check if its a valid certificate then add it to your list of parsed certificates.

pkg/collect/incluster-cert-collector.go
drohnow: I have not addressed this item yet; I will look do so tomorrow.

pkg/collect/incluster-cert-collector.go
drohnow: Working in progress; not completed yet.

pkg/collect/incluster-cert-collector.go
drohnow: I believe they are; I was getting a json error before; I believe that it was due to attempting to write to an empty json object but cannot remember exactly now. This is what worked for me but perhaps there is a better way?

pkg/collect/incluster-cert-collector.go
drohnow: I was getting a json error before; I believe that it was due to attempting to write to an empty json object but cannot remember exactly now. This is what worked for me but perhaps there is a better way?

pkg/collect/incluster-cert-collector.go
drohnow: I think so but maybe we should walk through it together.

pkg/collect/incluster-cert-collector.go
drohnow: will work on tomorrow.

drohnow:
Hi @banjoh I have not cleaned up all of the naming convention just yet; I wanted to focus on the internal logic before I did so.

This is still a work in progress but I was able to accomplish collecting multiple configmaps & secrets certificates into a single certificates.json file

drohnow:
Here is sample spec:

kind: SupportBundle
metadata:
  name: test
spec:
  collectors:
    - certificate:
        secretName: envoycert
        configMapName: kube-root-ca.crt  #secretname where certificate exists
        secretName: kotsadm-tls  #secretname where certificate exists

In testing again, I am not able to have two named secrets or two names configmaps in the same spec; I think that the json file is being overwritten by the last entry.

drohnow:
@banjoh I was able to achieve multiple collection of secrets by using a list.

kind: SupportBundle
metadata:
  name: test
spec:
  collectors:
    - certificate:
         secretName: ["envoycert", "kotsadm-tls", "registry-pki"]


All certificate collection is collected within the same certificates.json file. Now I am going to do the same thing with configmaps.

drohnow:
Updated code to add configMapNames list as well; now we can collect multiple configmaps and secrets.

example spec:

kind: SupportBundle
metadata:
  name: test
spec:
  collectors:
    - certificate: 
         secretNames: ["envoycert", "kotsadm-tls", "registry-pki"]
         configMapNames: ["kube-root-ca.crt", "dagr-configmap"]

drohnow:
Note: I am not able to rename my InclusterCertificate struct located in ...collector.shared.go as we already have a struct called that (located in hostcollector_shared.go).

What would you like me to rename InclusterCertificate to?

image

image

drohnow:
To account for source (configmap/secret) to namespace mapping; I went ahead and deployed a map (secretName or configMapName: namespace). 

The spec looks as such:

image

note: quotes are not required.

With this implementation, we can collect certificates from multiple secrets and configmaps in named namespaces; the one issue I see is that if there are secretNames and or configMapNames with the same name; if you specify a namespace then only certificates(s) from the last one in the spec will be collected.  However, if you leave the namespace empty; then it will collect all secretsNames/configMapNames across all namespaces.

Let me know your thoughts.

pkg/collect/incluster-cert-collector.go
drohnow: @banjoh

certificateChain: Do you want to collect the actual certificates and show them is the certificates.json file within the support-bundle?  If so, I am not sure if we should do that; If I were an end user or a security team, I would not want my certificates as a complete object collected and presented in a document that could leave my organization.  I think parsing and pulling back relevant information from the certificate to be able to evaluate it is fine but we should not collect whole certificates.  

I think the crypto/x509 module might have a way for us to verify/validate a certificate chain.

banjoh: > @banjoh
> 
> certificateChain: Do you want to collect the actual certificates and show them is the certificates.json file within the support-bundle? If so, I am not sure if we should do that; If I were an end user or a security team, I would not want my certificates as a complete object collected and presented in a document that could leave my organization. I think parsing and pulling back relevant information from the certificate to be able to evaluate it is fine but we should not collect whole certificates.
> 
> I think the crypto/x509 module might have a way for us to verify/validate a certificate chain.

Collecting whole certificates is not the intention. We need to go through each certificate in the chain and extract what we need from it. The result would be a list of certificate dictionaries/maps with cert info. See `"certificateChain": []` above

banjoh:
> Note: I am not able to rename my InclusterCertificate struct located in ...collector.shared.go as we already have a struct called that (located in hostcollector_shared.go).
> 
> What would you like me to rename InclusterCertificate to?
> 
> image
> 
> image

`Certificates` would be a good name. Its anyway a collector meant to collect many certs. The other one only targets one certificate

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -237,6 +249,7 @@ type Collect struct {
 	Longhorn         *Longhorn         `json:"longhorn,omitempty" yaml:"longhorn,omitempty"`
 	RegistryImages   *RegistryImages   `json:"registryImages,omitempty" yaml:"registryImages,omitempty"`
 	Sysctl           *Sysctl           `json:"sysctl,omitempty" yaml:"sysctl,omitempty"`
+	Certificates     *Certificates     `json:"certificate,omitempty" yaml:"certificate,omitempty"`
```
banjoh: `json:"certificates"...`

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -533,6 +546,10 @@ func (c *Collect) GetName() string {
 		collector = "sysctl"
 		name = c.Sysctl.Name
 	}
+	if c.Certificates != nil {
+		collector = "inclustercertificate"
```
banjoh: `certificates`

drohnow:
> 

Done.

drohnow:
Latest spec:

image

pkg/collect/certificate.go
```diff
@@ -0,0 +1,246 @@
...
+
+						if parsedCert.Issuer.CommonName != "" {
+
+							certInfo = append(certInfo, ParsedCertificate{
+								CertificateSource: CertificateSource{
+									ConfigMapName: configMap.Name,
+									Namespace:     configMap.Namespace,
+								},
+								CertName:                certName,
+								Subject:                 parsedCert.Subject,
```
banjoh: `parsedCert.Subject.ToRDNSequence()`

drohnow:
Hi @banjoh Please take a look at my most recent changes as per our meeting on Friday.  

One item:  Subject:  parsedCert.Subject.ToRDNSequence(), gives us output as follows: 
     
         image

       I don't think we want all of this; I am under the impression that you just want the subject name (ie: parsedCert.Subject.CommonName); I have verified the the commonName is the same as in the value of the ToRDNSequence().

     image

  I can rename commonName --> subject and get rid of the "parsedCert.Subject.ToRDNSequence()".


image

drohnow:
Here is an example of certificates.json where we are collecting both configMaps and secrets [certificates.json.pdf](https://github.com/replicatedhq/troubleshoot/files/11023377/certificates.json.pdf)

Here is a sample spec file:

image

```yaml
kind: SupportBundle
metadata:
  name: test
spec:
  collectors:
    - certificates: 
        configMaps: 
          #configMap name: namespace
          kube-root-ca.crt: kube-system
          dagr-configmap: default
          kube-root-ca.crt: ""
        secrets: 
          #secret name: namespace
          envoycert: projectcontour
      #   kotsadm-tls: default
          registry-pki: kurl
```

pkg/collect/collector.go
```diff
@@ -171,6 +173,9 @@ func getCollectorName(c interface{}) string {
 	case *CollectSysctl:
 		collector = "sysctl"
 		name = v.Collector.Name
+	case *CollectCertificates:
+		collector = "certificate"
```
banjoh: Change to `collector = "certificates"`

pkg/collect/certificates.go
```diff
@@ -0,0 +1,238 @@
...
+	secrets, _ := client.CoreV1().Secrets(namespace).List(context.Background(), listOptions)
+
+	for _, secret := range secrets.Items {
+		if secretName == secret.Name {
+
+			for certName, certs := range secret.Data {
+
+				data := string(certs)
+
+				if strings.Contains(data, "BEGIN CERTIFICATE") && strings.Contains(data, "END CERTIFICATE") {
```
banjoh: Not needed anymore. `decodePem` will handle this

drohnow:
Hi @banjoh 
ok, I have same resource name (secret or configmap) in multiple namespaces cert collections working in the json format we want.  The certificates.json file shows that we collected two envoycerts in two different namespaces.
[certificates.json.pdf](https://github.com/replicatedhq/troubleshoot/files/11073040/certificates.json.pdf)

Here is what the spec.yaml looks like:

```
kind: SupportBundle
metadata:
  name: test
spec:
  collectors:
    - certificates: 
        secrets: 
          envoycert: ["default", "projectcontour"] 
          kotsadm-tls: ["default"]
```

drohnow:
@banjoh I changed the method we fetch secrets from "List" to "Get"

- from: secrets, _ := client.CoreV1().Secrets(namespace).List(context.Background(), listOptions)
- to: secret, _ := client.CoreV1().Secrets(namespace).Get(context.Background(), secretName, getOptions)

it seems more efficient to go right to the secret since we know it rather than pull back a list of secrets and then fetch the one we want.

pkg/collect/certificates.go
```diff
@@ -0,0 +1,221 @@
...
+		if errParse != nil {
+			trackErrors = append(trackErrors, errParse.Error())
+			continue
+		}
+
+		certInfo = append(certInfo, ParsedCertificate{
+			CertName:                certName,
+			Subject:                 parsedCert.Subject.ToRDNSequence().String(),
+			SubjectAlternativeNames: parsedCert.DNSNames,
+			Issuer:                  parsedCert.Issuer.CommonName,
```
banjoh: `Issuer: parsedCert.Subject.ToRDNSequence().String()`

drohnow:
[certificates.json.txt](https://github.com/replicatedhq/troubleshoot/files/11091033/certificates.json.txt)

DexterYan:
Hey @drohnow, 

For the secrets collector, I found if envoycert does not exist in default namespace
```
- certificates: 
        secrets: 
          envoycert: ["default", "projectcontour"] 
          kotsadm-tls: ["default"]
```
, then it will return a empty result like
```
[
	{
		"source": {},
		"errors": [],
		"certificateChain": []
	},
	{
		"source": {
			"secret": "envoycert",
			"namespace": "projectcontour"
		},
...
]
```
Is it expected? Or maybe can filter it out for those not existed results.

pkg/collect/certificates.go
```diff
@@ -0,0 +1,221 @@
...
+
+	for certName, certs := range secret.Data {
+		certInfo, _ := CertParser(certName, certs)
+
+		collection = append(collection, certInfo...)
+
+	}
+	results = append(results, CertCollection{
+		Source:           source,
+		Errors:           trackErrors,
```
DexterYan, line 194: For trackErrors field, it contains the error from CertParser in design, right?

drohnow: Hi @DexterYan Thank you for reviewing!  A TODO for myself tomorrow is to shore up the error collections; the CertParser for sure needs a bit of touch with regards to trackErrors; so I agree with your assessment.

drohnow:
> Hey @drohnow,
> 
> For the secrets collector, I found if envoycert does not exist in default namespace
> 
> ```
> - certificates: 
>         secrets: 
>           envoycert: ["default", "projectcontour"] 
>           kotsadm-tls: ["default"]
> ```
> 
> , then it will return a empty result like
> 
> ```
> [
> 	{
> 		"source": {},
> 		"errors": [],
> 		"certificateChain": []
> 	},
> 	{
> 		"source": {
> 			"secret": "envoycert",
> 			"namespace": "projectcontour"
> 		},
> ...
> ]
> ```
> 
> Is it expected? Or maybe can filter it out for those not existed results.

Hi @DexterYan Yah, I did not think of that use case; great catch!  I will add a conditional to be sure that the secret or configmap exists in the namespace before attempting to collect. I will add a test for this as well.  
(note, the code is already there and codified; I will add configmap collections to the code tomorrow; we disabled during review meeting to reduce noise).

Thank you!

pkg/collect/certificates.go
```diff
@@ -0,0 +1,230 @@
...
+
+// configmap certificate collector function will
+func configMapCertCollector(configMapName string, namespace string, client kubernetes.Interface) []CertCollection {
+
+	results := []CertCollection{}
+
+	getOptions := metav1.GetOptions{}
+
+	// Collect from configMaps
+	configMap, _ := client.CoreV1().ConfigMaps(namespace).Get(context.Background(), configMapName, getOptions)
```
banjoh: We can add some error handling here

```go
	secret, err := client.CoreV1().Secrets(namespace).Get(context.Background(), secretName, getOptions)
	if err != nil {
		trackErrors = append(trackErrors, err.Error())
		// results = append(results, CertCollection{
		// 	Source:           source,
		// 	Errors:           trackErrors,
		// 	CertificateChain: collection,
		// })
		// return results... No need to continue
	}
```

pkg/collect/certificates.go
```diff
@@ -0,0 +1,230 @@
...
+
+	results := []CertCollection{}
+
+	//listOptions := metav1.ListOptions{}
+	//secrets, _ := client.CoreV1().Secrets(namespace).List(context.Background(), listOptions)
+
+	// TODO: Handle RBAC errors. Not to be worked on yet
+	getOptions := metav1.GetOptions{}
+	// Collect from secrets
+	secret, _ := client.CoreV1().Secrets(namespace).Get(context.Background(), secretName, getOptions)
```
banjoh: same here

pkg/collect/certificates.go
```diff
@@ -0,0 +1,230 @@
...
+			configMapCollections := configMapCertCollector(configMapName, namespace, c.Client)
+			results = append(results, configMapCollections...)
+		}
+	}
+
+	certsJson, _ := json.MarshalIndent(results, "", "\t")
+
+	filePath := "certificates/certificates.json"
+
+	output.SaveResult(c.BundlePath, filePath, bytes.NewBuffer(certsJson))
```
banjoh: Could you please also add error handling here

```go
err := output.SaveResult(c.BundlePath, filePath, bytes.NewBuffer(certsJson))
if err != nil {
	return nil, err
}
```

pkg/collect/certificates.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// collect certificates from configMaps
+	for configMapName, namespaces := range c.Collector.ConfigMaps {
+		for _, namespace := range namespaces {
+			configMapCollections := configMapCertCollector(configMapName, namespace, c.Client)
+			results = append(results, configMapCollections...)
+		}
+	}
+
+	certsJson, _ := json.MarshalIndent(results, "", "\t")
```
banjoh: Could you please also add error handling here

```go
certsJson, err := json.MarshalIndent(results, "", "\t")
if err != nil {
	return nil, err
}
```

banjoh (commented):
The collector looks good now. I added a few minor comments on error handling

pkg/collect/certificates.go
drohnow: completed

drohnow:
@banjoh or @diamonwiggins. I believe that I have satisfied all PR recommendations.  I think this collector is ready to ship.

drohnow:
Added more clear language to error output when fail to find a configMap or secret.

pkg/collect/certificates_test.go
```diff
@@ -0,0 +1,223 @@
...
+	certParserTests := []struct {
+		Name      string
+		Certs     []byte
+		CertQty   int
+		IsSubject bool
+		IsCert    bool
+		IsValid   bool
+	}{
+		// Name, Certs, CertQty, IsSubject, IsCert, IsValid
+		{"Widgits", expiredCert, 1, true, true, false},
```
DexterYan, line 142: For the test case, I think we will put the fields to make it easier to read.
```
		{
			Name:      "Widgits",
			Certs:     expiredCert,
			CertQty:   1,
			IsSubject: true,
			IsCert:    true,
			IsValid:   false,
		},
```

And for name, it is usually a test name 
```
{
			name:              "without container name",
			withContainerName: false,
			collectorName:     "all-logs",
			createSymLinks:    true,
			want: CollectorResult{
				"all-logs/test-pod.log":                                                []byte("fake logs"),
				"all-logs/test-pod-previous.log":                                       []byte("fake logs"),
				"cluster-resources/pods/logs/my-namespace/test-pod/nginx.log":          []byte("fake logs"),
				"cluster-resources/pods/logs/my-namespace/test-pod/nginx-previous.log": []byte("fake logs"),
			},
		},
```
here, I think you are using the Name as a cert subject name

pkg/collect/certificates_test.go
```diff
@@ -0,0 +1,223 @@
...
+				isSubject := true
+				if e.IsSubject != isSubject {
+					t.Errorf("You expected that %s certificate contains %s in the Subject line but it was not present; here is the entire subject string: %v", e.Name, e.Name, cert.Subject)
+				}
+			}
+
+			// test checks for expected quantity of certificates in a certificate file
+			numCerts := len(results)
+			if numCerts != e.CertQty {
+				t.Errorf("expected %d certificates in slice but got %d", e.CertQty, numCerts)
```
DexterYan, line 183: we can change it to
```
assert.Equal(t, e.CertQty, len(results), "Expected %v, but got %v", e.CertQty, len(results))
```
And I think it is just one time comparing, may not need to be put into for loop, which will be the same result for each loop

pkg/collect/certificates_test.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, e := range certParserTests {
+
+		results, _ := CertParser(e.Name, e.Certs)
+
+		// tests if results contains a list of parsed certificates
+		var isCert bool
+		if len(results) == 0 {
+			isCert = false
+		} else {
+			isCert = true
```
DexterYan, line 157: For isCert, I think if we are comparing that e.CertQty and len(results), we may not need to use isCert, since the number of results does not match the number of wanted cert numbers, which indicate that the cert is not valid

Can put 
```
assert.Equal(t, e.CertQty, len(results), "Expected %v, but got %v", e.CertQty, len(results))
```
here and remove isCert

pkg/collect/certificates_test.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, e := range certDecoderTests {
+		results, _ := decodePem(e.certInput)
+		certCount := 0
+
+		for _, cert := range results.Certificate {
+			log.Println(cert)
+			certCount++
+		}
+		if certCount != e.certCount {
+			t.Errorf("cert count -- expected %d, but got %d", e.certCount, certCount)
```
DexterYan, line 219: can use assert.Equal(t, tt.want, got)

xavpaice:
The suggestions from Dexter's review are good, but maybe not totally required for this particular change. 

This change includes 249 commits, many of which do not have helpful commit messages.  I'd like to see these cleaned up.

To do the cleanup of the commits:

1. checkout prior to the last update from main (my bad!): `git checkout HEAD~1`
2. rebase the last 249 commits: `git rebase -i HEAD~249`
3. Make a nice commit message(s) for the ones you select
4. force push the branch

This will update the PR.

In order for tests to pass, you will need to run `make schemas` so that the schema is updated, and commit the resulting changes.

xavpaice:
This was replaced by #1112

PR #1042, opened 2023-03-03
fix: Use namespace from the collector spec for Ceph

### LINKED ISSUE #1041: Ceph collector ignores user provided namespace
**Bug Description**

The Ceph collector ignores the user provided namespace from the [collector spec](https://troubleshoot.sh/docs/collect/ceph/#namespace-optional) and only uses the namespace from the kubeconfig context or `---namespace` flag. This is a regression caused by the struct refactor done last year in #670

**Expected Behavior**
The below should look for a rook-ceph installation in `some-namespace` and not `rook-ceph`.

```
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - ceph:
         namespace: some-namespace
```

**Steps To Reproduce**
Attempt to use the Ceph collector for a Ceph installation in a namespace other than `rook-ceph

**Additional Context**

- Operating system
- Operating system version

diamonwiggins:

Correctly use namespace from the ceph collector spec and not from the `kubeconfig` or `--namespace` flag. This is a regression from the struct refactor for collectors

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/ceph.go
+++ b/pkg/collect/ceph.go
@@ -128,11 +128,11 @@ func (c *CollectCeph) IsExcluded() (bool, error) {
 func (c *CollectCeph) Collect(progressChan chan<- interface{}) (CollectorResult, error) {
 	ctx := context.TODO()
 
-	if c.Namespace == "" {
-		c.Namespace = DefaultCephNamespace
+	if c.Collector.Namespace == "" {
+		c.Collector.Namespace = DefaultCephNamespace
 	}
 
-	pod, err := findRookCephToolsPod(ctx, c, c.Namespace)
+	pod, err := findRookCephToolsPod(ctx, c, c.Collector.Namespace)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


banjoh (approved):
I think a bigger change will be needed for all collectors to support namespace precedence where the order would be (pick 3 if 2 or 1 do not exist and so forth)

1. `c.Collector.Namespace` -> defined per collector
2. `c.Namespace` -> Passed in from the environment/cli e.g [here in the CLI](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L270)
3. Defaults e.g `DefaultCephNamespace` for ceph. We also need to consider cases where the namespace comes from kubeconfig. I think this falls in this "default" precedence.

diamonwiggins:
> I think a bigger change will be needed for all collectors to support namespace precedence where the order would be (pick 3 if 2 or 1 do not exist and so forth)
> 
> 1. `c.Collector.Namespace` -> defined per collector
> 2. `c.Namespace` -> Passed in from the environment/cli e.g [here in the CLI](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L270)
> 3. Defaults e.g `DefaultCephNamespace` for ceph. We also need to consider cases where the namespace comes from kubeconfig. I think this falls in this "default" precedence.

Yeah, agreed. Will be sure to get an issue created for this today.

PR #1063, opened 2023-03-21
feat: Add partition type info to block device collector

### LINKED ISSUE #970: Add partition information to block device collector output and analyser

Some add-ons such as the [rook add-on](https://kurl.sh/docs/add-ons/rook#block-storage) require an unpartitioned block device for ceph to use. This information is currently not collected by the [block devices host collector](https://troubleshoot.sh/docs/host-collect-analyze/blockDevices/).

**Describe the feature**

- Collector
The collector parses `lsblk` output using [these specified columns](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/host_block_device.go#L29). We can extract more information from `lsblk` output by adding more flags that contain partition table information which the analyser would use to evaluate if a block device is partitioned or not. Here is the output from a partitioned and non-partitioned host.

_Partitioned_
```sh
$ sudo lsblk /dev/sdb --noheadings --bytes --pairs -o NAME,PTUUID,PTTYPE,PARTTYPE,PARTLABEL,PARTFLAGS
NAME="sdb" PTUUID="906db4f7-f33e-43e5-98f2-2218e41ba814" PTTYPE="gpt" PARTTYPE="" PARTLABEL="" PARTFLAGS=""
NAME="sdb1" PTUUID="906db4f7-f33e-43e5-98f2-2218e41ba814" PTTYPE="gpt" PARTTYPE="0fc63daf-8483-4772-8e79-3d69d8477de4" PARTLABEL="primary" PARTFLAGS=""
```

_Unpartitioned_
```sh
$ sudo lsblk /dev/sdb --noheadings --bytes --pairs -o NAME,PTUUID,PTTYPE,PARTTYPE,PARTLABEL,PARTFLAGS
NAME="sdb" PTUUID="" PTTYPE="" PARTTYPE="" PARTLABEL="" PARTFLAGS=""
```

- Analyser
Update the [block devices analyser](https://troubleshoot.sh/docs/host-collect-analyze/blockDevices/) to expose this information i.e whether a `disk` type device is partitioned or not. Also consider other device types e.g `raid`.

Leave things as they are and not catch this use case

banjoh:

A minor improvement in the block devices collector. We now collect the partition type information of a block device.

Fixes: #970

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/analyze/host_block_devices.go
+++ b/pkg/analyze/host_block_devices.go
@@ -116,7 +116,7 @@ func (a *AnalyzeHostBlockDevices) Analyze(getCollectedFileContents func(string) 
 func compareHostBlockDevicesConditionalToActual(conditional string, minimumAcceptableSize uint64, includeUnmountedPartitions bool, devices []collect.BlockDeviceInfo) (res bool, err error) {
 	parts := strings.Split(conditional, " ")
 	if len(parts) != 3 {
-		return false, fmt.Errorf("Expected exactly 3 parts, got %d", len(parts))
+		return false, fmt.Errorf("expected exactly 3 parts, got %d", len(parts))
 	}
 
 	rx, err := regexp.Compile(parts[0])
@@ -143,7 +143,7 @@ func compareHostBlockDevicesConditionalToActual(conditional string, minimumAccep
 		return count == desiredInt, nil
 	}
 
-	return false, fmt.Errorf("Unexpected operator %q", parts[1])
+	return false, fmt.Errorf("unexpected operator %q", parts[1])
 }
 
 func countEligibleBlockDevices(rx *regexp.Regexp, minimumAcceptableSize uint64, includeUnmountedPartitions bool, devices []collect.BlockDeviceInfo) int {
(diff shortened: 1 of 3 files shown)


emosbaugh (approved):

chris-sanders:
I'm not clear what the original goal/use case of this is. Are you considering that LVM won't show up like a normal partition? Should we be checking for PV's and LV's if we're checking for filesystem information? I'm only concerned if, depending on the use case, if you're going to end up implying that a disk is 'not used' just to find out it was actually used as part of LVM PV.

banjoh:
> I'm not clear what the original goal/use case of this is. Are you considering that LVM won't show up like a normal partition? Should we be checking for PV's and LV's if we're checking for filesystem information? I'm only concerned if, depending on the use case, if you're going to end up implying that a disk is 'not used' just to find out it was actually used as part of LVM PV.

The original issue, to filter out partitions from the list presented to rook/ceph, turned out to be a non-issue (see [comment](https://github.com/replicatedhq/troubleshoot/issues/970#issuecomment-1477829464)). The change in this PR was introduced as an improvement. I found myself needing to see the filesystem a partition had (`ceph_bluestore`, `ext4`, `crypto_LUKS` etc) when looking at the output of the collector. This is more informational than input to the analyser.

PR #1068, opened 2023-03-22
feat(collector): Add host collector to copy files to a bundle

### LINKED ISSUE #881: hostCollector: `copy` to copy files/folders from a host filesystem to the support bundle

Right now we have `copyFromHost` that runs as an in-cluster collector, which works great if pods can be scheduled, but in some cases of a node-down or cluster-down problem, it might be nice to be able to have a "simple" way to collect a specific file/path/glob-pattern from the host.

**Describe the feature**

A native Golang collector that can take a Go path that will copy the desired files into a destination folder in the support bundle.  Be able to report errors like "file not found" so a user has a little more information on the failure mode, and authors can trigger outcomes based on those conditions.  It should also preserve the original file mode and ownership information so we don't inadvertently change one of those things, which might be germane to a problem.

Right now my solution is to use `run` to execute some shell commands to get the output I want, which works but could fail for lots of reasons, since it relies on host dependencies.

**Additional context**

see https://github.com/adamancini/troubleshoot-specs/blob/cni/host/cni-conf.yaml#L12

banjoh:

Often we want to collect files from a node, just like we do within a cluster using the [copy-from-host collector](https://troubleshoot.sh/docs/collect/copy-from-host/). This collector, although doing what it should, would not work when a cluster is down. This PR introduces a new collector that copies files from a [host collector](https://troubleshoot.sh/docs/host-collect-analyze/overview/) that can be used to copy files into a bundle without requiring a k8s cluster. Here is an example spec

```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  hostCollectors:
    - copy:
        collectorName: copy-nginx-logs
        path: /var/log/nginx/access*  # glob syntax
```

The path parameter accepts a path conforming to glob syntax. All directories and files that match the path parameter are all collected in `<bundle>/host-collectors/<collector-name>/...`. If there is no collector name defined, the path would `<bundle>/host-collectors/copy/...`

The collection logic follows symlinks as well

Fixes: #881

## Checklist

- [ ] Yes
- [x] No


chris-sanders:
@banjoh maybe replace the template text with the requested motivation and context? :)

banjoh:
I'm working on documentation for troubleshoot.sh

banjoh:
Documentation PR https://github.com/replicatedhq/troubleshoot.sh/pull/484

diamonwiggins (approved):
lgtm

PR #1073, opened 2023-03-23
chore(ci): use fetch-depth instead of unshallow

twz123:

The fetch-depth parameter to the checkout action will provide a full checkout instead of a shallow one that only has the last commit. Using this makes the workflow files a bit more concise and removes an extra step.

## Checklist

- [ ] Yes
- [x] No


banjoh (approved):
LGTM

PR #1074, opened 2023-03-23
feat: Build and release binaries for Linux ARM 32-bit

### LINKED ISSUE #1069: Add linux/arm release

At k0s we support amd64, arm64 and arm and we need support-bundle into all our architectures, therefore we would like to have official releases for every architecture.

Currently we can compile it into arm without further changes by just changing the GOARCH and we haven't found any issue so far.

**Describe the feature**

Release binaries for linux/arm just like it's done already for linux/arm64 and linux/amd64.

We can build support bundle ourselves: https://github.com/k0sproject/k0s/pull/2902/files#diff-824296bdf39d6c59a3131a8c15d0031a1274bc0243c37a0941f48b6353e13cd8

This is not a blocker for us, we're asking for this because we think that:
1-  Releasing arm support-bundle in its own repo makes sense not just for us but for the project itself.
2- We've had a look at the build-test-deploy.yaml workflow and at goreleaser and we think this change wouldn't be very difficult or large. Also we don't think this will add a significant amount of maintenance in the long run.

We can make the PR for these changes ourselves if you agree that it makes sense.

twz123:

Fixes #1069:

* Add arm to GOARCHes in .goreleaser.yaml and exclude the windows/arm build
* Add arm/linux to goreleaser-test GitHub workflow job
* Add the linux/arm binaries to krew plugin manifests

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/build-test-deploy.yaml
+++ b/.github/workflows/build-test-deploy.yaml
@@ -232,6 +232,9 @@ jobs:
       matrix:
         goarch: [amd64, arm64]
         goos: [darwin, linux, windows]
+        include:
+          - goarch: arm
+            goos: linux
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/deploy/.goreleaser.yaml
+++ b/deploy/.goreleaser.yaml
@@ -13,7 +13,11 @@ builds:
       - windows
     goarch:
       - amd64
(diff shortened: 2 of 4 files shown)


banjoh (approved):
LGTM

PR #1075, opened 2023-03-23
change troubleshoot.io to troubleshoot.sh

### LINKED ISSUE #1070: default in-cluster label references a domain that is not registered
**Bug Description**

The default label we use to search for secrets and configMaps containing Troubleshoot specs in a cluster ("troubleshoot.io/kind=support-bundle") refers to a domain "troubleshoot.io" which is not registered.

**Expected Behavior**

The label should reference "troubleshoot.sh" which is registered for the Troubleshoot project.

**Steps To Reproduce**

n/a

**Additional Context**

Changes required:

- [x] Troubleshoot itself defaults and labels #1203 
- [x] #1204 
- [ ] https://github.com/replicatedhq/kURL/pull/4247 kURL change
- [x] https://github.com/replicatedhq/replicated-docs/pull/1039 docs (change KOTS to use Troubleshoot v0.69.0 first)
- [ ] Change the label that KOTS uses, and update it to Troubleshoot v0.69.0

grantmiller:

i don't believe that troubleshoot.io is right... should be troubleshoot.sh


chris-sanders:
I believe this relates to #1070 which we need to fix, but this just fixes the documentation string we need to also adjust the default search, yes?

xavpaice:
There's a number of changes that need to happen at the same time: see [this query](https://github.com/search?q=org%3Areplicatedhq+troubleshoot.io&type=code)

tl;dr:
* This change in Troubleshoot plus a rebuild of the schemas and docs
* search/replace in KOTS and kURL
* https://github.com/replicatedhq/troubleshoot.sh/pull/480 (docs)

xavpaice:
https://github.com/replicatedhq/kots/pull/3763

xavpaice:
https://github.com/replicatedhq/kURL/pull/4247

danj-replicated:
rather than treat this as a co-ordinated change, would it be better to address this in a similar way to how kubernetes treats label deprecations?

i.e: we add the new label while leaving the original one in place and set a deprecation/removal date that upstream dependants (kurl) must meet?

banjoh:
> rather than treat this as a co-ordinated change, would it be better to address this in a similar way to how kubernetes treats label deprecations?
> 
> i.e: we add the new label while leaving the original one in place and set a deprecation/removal date that upstream dependants (kurl) must meet?

My thoughts exactly.

In addition, since we (Replicated) maintain KOTS & kurl, we can make the changes there to deprecate the label. In KOTS side, we would need to update existing labels when users upgrade their KOTS deployments.

xavpaice:
I think if we're going to deprecate a label we need to work in this order:
* deprecate in Troubleshoot, search for both new and old for a time
* Adjust all the docs
* in KOTs, search for new and old.  Change what we store to the new label.
* In kURL, just change what we store to the new label.

Sound reasonable?

danj-replicated:
yeh sounds like a plan

banjoh:
Yep, that sounds like a good plan

banjoh:
> In kURL, just change what we store to the new label.

There will be a small window where new add-on versions installed using an old kURL installer won't work. It might be something we can have checked in [flannel's preflight](https://github.com/replicatedhq/kURL/blob/main/addons/flannel/0.21.3/host-preflight.yaml) (kurl-version > v2023.XX.XX) since its the only reference I see the label being used.

docs/preflight.md
```diff
@@ -17,7 +17,7 @@ preflight [url] [flags]
       --as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
       --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
       --as-uid string                  UID to impersonate for the operation.
-      --cache-dir string               Default cache directory (default "$HOME/.kube/cache")
+      --cache-dir string               Default cache directory (default "/Users/xavpaice/.kube/cache")
```
xavpaice:
```suggestion
      --cache-dir string               Default cache directory (default "$HOME/.kube/cache")
```

cmd/troubleshoot/cli/root.go
```diff
@@ -68,7 +68,7 @@ from a server that can be used to assist when troubleshooting a Kubernetes clust
 	cmd.Flags().Bool("redact", true, "enable/disable default redactions")
 	cmd.Flags().Bool("interactive", true, "enable/disable interactive mode")
 	cmd.Flags().Bool("collect-without-permissions", true, "always generate a support bundle, even if it some require additional permissions")
-	cmd.Flags().StringSliceP("selector", "l", []string{"troubleshoot.io/kind=support-bundle"}, "selector to filter on for loading additional support bundle specs found in secrets within the cluster")
+	cmd.Flags().StringSliceP("selector", "l", []string{"troubleshoot.sh/kind=support-bundle"}, "selector to filter on for loading additional support bundle specs found in secrets within the cluster")
```
xavpaice, line 71: if we're going to deprecate troubleshoot.io, but keep support for it, what should we have as the default here?  I think the new, but want feedback on that given that we want to support the old and the new label.  For one to be default, that means that folks using the other will need to supply it on the CLI, unless we add logic for multiple labels and have this default to both labels.

docs/support-bundle.md
```diff
@@ -17,15 +17,16 @@ support-bundle [urls...] [flags]
       --as string                      Username to impersonate for the operation. User could be a regular user or a service account in a namespace.
       --as-group stringArray           Group to impersonate for the operation, this flag can be repeated to specify multiple groups.
       --as-uid string                  UID to impersonate for the operation.
-      --cache-dir string               Default cache directory (default "$HOME/.kube/cache")
+      --cache-dir string               Default cache directory (default "/Users/xavpaice/.kube/cache")
```
xavpaice:
```suggestion
      --cache-dir string               Default cache directory (default "$HOME/.kube/cache")
```

docs/support-bundle.md
```diff
@@ -38,13 +39,14 @@ support-bundle [urls...] [flags]
       --redact                         enable/disable default redactions (default true)
       --redactors strings              names of the additional redactors to use
       --request-timeout string         The length of time to wait before giving up on a single server request. Non-zero values should contain a corresponding time unit (e.g. 1s, 2m, 3h). A value of zero means don't timeout requests. (default "0")
-  -l, --selector strings               selector to filter on for loading additional support bundle specs found in secrets within the cluster (default [troubleshoot.io/kind=support-bundle])
+  -l, --selector strings               selector to filter on for loading additional support bundle specs found in secrets within the cluster (default [troubleshoot.sh/kind=support-bundle])
```
xavpaice, line 42: as above, need to determine the correct default

xavpaice (changes requested):

xavpaice:
this has been replaced by #1203

PR #1088, opened 2023-03-31
fix/clusterPodStatuses: only process `when` conditional if specified

### LINKED ISSUE #1087: Unable to use `when` condition for `clusterPodStatuses` within Preflight. By doing so I can see error `Error invalid 'when' format:`
**Bug Description**

Unable to use `when` condition for `clusterPodStatuses` within Preflight. By doing so I can see error `Error invalid 'when' format:`

**Expected Behavior**

Include a description of the expected behaviour here.

**Steps To Reproduce**

Use preflight to run:

```ymal
kind: Preflight
metadata:
  name: kurl-builtin-oncluster
spec:
  collectors:
    - clusterResources: {}
    - clusterInfo: {}
  analyzers:
    - clusterPodStatuses:
        checkName: "Pod(s) healthy"
        exclude: '{{kurl or (not .IsPrimary) (not .IsUpgrade) }}'
        outcomes:
          - warn:
              when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
              message: "A Pod, {{ .Name }}, is unhealthy with a status of: {{ .Status.Reason }}. Restarting the pod may fix the issue."
          - pass:
              message: "All Pods are OK."
    - clusterPodStatuses:
        checkName: "Kurl Pod(s) healthy"
        exclude: '{{kurl or (not .IsPrimary) (not .IsUpgrade) }}'
        namespaces:
          - kurl
        outcomes:
          - fail:
              when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
              message: "A Pod, {{ .Name }}, in the kurl namespace is unhealthy with a status of {{ .Status.Reason }}. Restarting the pod may fix the issue."
          - pass:
              message: "All Pods in Kurl namespace are Running."
```

**Additional Context**

- Troubleshoot version. If you built from source, note that including the version of Go you used to build with. `v0.59.0`
- Operating system linux/arm64
- Operating system version

diamonwiggins:

Currently the `when` conditional for the `clusterPodStatuses` analyzer is expected for all outcome types and if not specified the following behavior is seen using the following spec:

```
kind: Preflight
metadata:
  name: kurl-builtin-oncluster
spec:
  collectors:
    - clusterResources: {}
    - clusterInfo: {}
  analyzers:
    - clusterPodStatuses:
        checkName: "Pod(s) healthy"
        outcomes:
          - warn:
              when: "!= Healthy" # Catch all unhealthy pods. A pod is considered healthy if it has a status of Completed, or Running and all of its containers are ready.
              message: "A Pod, {{ .Name }}, is unhealthy with a status of: {{ .Status.Reason }}. Restarting the pod may fix the issue."
          - pass:
              message: "All Pods are OK."
```

Screenshot 2023-03-31 at 10 36 27 AM

Because `when` isn't set for `pass` you see the above errors. With the changes in this PR, `when` is optional and the expected pattern of having a default outcome at the end of your outcome list is supported.

Screenshot 2023-03-31 at 10 36 02 AM

Fixes: #1087 

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/pkg/analyze/cluster_pod_statuses.go
+++ b/pkg/analyze/cluster_pod_statuses.go
@@ -104,33 +104,36 @@ func clusterPodStatuses(analyzer *troubleshootv1beta2.ClusterPodStatuses, getChi
 				continue
 			}
 
-			parts := strings.Split(strings.TrimSpace(when), " ")
-			if len(parts) < 2 {
-				println(fmt.Sprintf("invalid 'when' format: %s\n", when)) // don't stop
-				continue
-			}
-
-			operator := parts[0]
-			reason := parts[1]
+			operator := ""
+			reason := ""
 			match := false
-
-			switch operator {
-			case "=", "==", "===":
-				if reason == "Healthy" {
-					match = !k8sutil.IsPodUnhealthy(&pod)
-				} else {
-					match = reason == string(pod.Status.Phase) || reason == string(pod.Status.Reason)
+			if when != "" {
+				parts := strings.Split(strings.TrimSpace(when), " ")
+				if len(parts) < 2 {
+					println(fmt.Sprintf("invalid 'when' format: %s\n", when)) // don't stop
+					continue
 				}
-			case "!=", "!==":
-				if reason == "Healthy" {
-					match = k8sutil.IsPodUnhealthy(&pod)
-				} else {
-					match = reason != string(pod.Status.Phase) && reason != string(pod.Status.Reason)
-				}
-			}
+				operator = parts[0]
+				reason = parts[1]
 
-			if !match {
-				continue
+				switch operator {
+				case "=", "==", "===":
+					if reason == "Healthy" {
+						match = !k8sutil.IsPodUnhealthy(&pod)
+					} else {
+						match = reason == string(pod.Status.Phase) || reason == string(pod.Status.Reason)
+					}
+				case "!=", "!==":
+					if reason == "Healthy" {
+						match = k8sutil.IsPodUnhealthy(&pod)
+					} else {
+						match = reason != string(pod.Status.Phase) && reason != string(pod.Status.Reason)
+					}
+				}
+
+				if !match {
+					continue
+				}
 			}
 
 			r.InvolvedObject = &corev1.ObjectReference{
--- a/pkg/analyze/cluster_pod_statuses.go
+++ b/pkg/analyze/cluster_pod_statuses.go
@@ -3,7 +3,6 @@ package analyzer
 import (
 	"bytes"
 	"encoding/json"
-	"fmt"
 	"path/filepath"
 	"strings"
 	"text/template"
@@ -13,6 +12,7 @@ import (
 	"github.com/replicatedhq/troubleshoot/pkg/constants"
 	"github.com/replicatedhq/troubleshoot/pkg/k8sutil"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 )
 
 type AnalyzeClusterPodStatuses struct {
@@ -100,7 +100,7 @@ func clusterPodStatuses(analyzer *troubleshootv1beta2.ClusterPodStatuses, getChi
 				r.URI = outcome.Pass.URI
 				when = outcome.Pass.When
 			} else {
-				println("error: found an empty outcome in a clusterPodStatuses analyzer") // don't stop
+				klog.Error("error: found an empty outcome in a clusterPodStatuses analyzer\n")
 				continue
 			}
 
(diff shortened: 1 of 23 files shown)


camilamacedo86 (commented):
@diamonwiggins that is great. Could we try to return only one result when pods are Healthy instead of a list with all Healthy pods on the cluster? Could we just return the list of those that are unhealthy?

diamonwiggins:
> @diamonwiggins that is great. Could we try to return only one result when pods are Healthy instead of a list with all Healthy pods on the cluster? Could we just return the list of those that are unhealthy?

I think that's a good idea. Do you think we need to cover that here on this PR?

camilamacedo86 (dismissed):
Return many times the same success result seems a bug out of scope of this PR. See: https://github.com/replicatedhq/troubleshoot/issues/1089

I think would be nice it has tests but I am so happy with the fix already.
Thank you 👍

camilamacedo86 (approved):

camilamacedo86 (dismissed):
Terrific work 🥇 I added only one nit.
IF possible would be great have the mocks under a directory called testdata to follow the best practices with Golang to keep those. Otherwise, it has my LGTM

pkg/analyze/files/pods/default-unhealthy.json
```diff
@@ -0,0 +1,263 @@
+{
```
camilamacedo86, line 1: Could we add it inside of a directory called testdata?
Why Golang has a particular meaning for those and for we add mock data that is the best approach.

banjoh: +1

Specifically https://github.com/replicatedhq/troubleshoot/tree/main/testdata

```sh
[evans] $ go help test | grep -A1 testdata
The go tool will ignore a directory named "testdata", making it available
to hold ancillary data needed by the tests.
```

But it can be done in a separate PR. There are many other test data files in the [current directory](https://github.com/replicatedhq/troubleshoot/tree/main/pkg/analyze/files) used

banjoh:
> @diamonwiggins that is great. Could we try to return only one result when pods are Healthy instead of a list with all Healthy pods on the cluster? Could we just return the list of those that are unhealthy?

Looking at other status analysers (Deployment, StatefulSet...), it looks like the intention is to show the analysis of each resource. I however see your concern here @camilamacedo86. We can improve how these analysis results are reported back by adding a [title to each outcome](https://troubleshoot.sh/docs/analyze/outcomes/#title) like so. There is a bug (https://github.com/replicatedhq/troubleshoot/issues/1093) that needs to be fixed first before this is possible

```yaml
  analyzers:
    - clusterPodStatuses:
        checkName: "Pod(s) health status(es)"
        outcomes:
          - fail:
              title: "Pod {{ .Name }} is unable to pull images"
              when: "== ImagePullBackOff"
              message: "A Pod, {{ .Name }}, is unable to pull its image. Status is: {{ .Status.Reason }}"
          - warn:
              title: "Pod {{ .Name }} is unhealthy"
              when: "!= Healthy"
              message: "A Pod, {{ .Name }}, is unhealthy with a status of: {{ .Status.Reason }}. Restarting the pod may fix the issue."
          - pass:
              title: "Pod {{ .Name }} is healthy"
              message: "Pod {{ .Name }} is healthy"
```

This would render results like below

Screenshot 2023-04-03 at 11 53 46

banjoh:
> Could we just return the list of those that are unhealthy?

@camilamacedo86 I think if we ignore the `pass` condition, all the healthy ones are not shown in the analysis outcomes

pkg/analyze/cluster_pod_statuses_test.go
```diff
@@ -0,0 +1,266 @@
+package analyzer
+
+import (
```
banjoh, line 3: Good stuff with the tests!

Nit pick: Could you add/modify tests to use `!==`,  `==`, `===` and `=` conditionals for completeness? e.g `=== CrashLoopBackOff`

pkg/analyze/cluster_pod_statuses.go
```diff
@@ -104,33 +104,36 @@ func clusterPodStatuses(analyzer *troubleshootv1beta2.ClusterPodStatuses, getChi
...
-			switch operator {
-			case "=", "==", "===":
-				if reason == "Healthy" {
-					match = !k8sutil.IsPodUnhealthy(&pod)
-				} else {
-					match = reason == string(pod.Status.Phase) || reason == string(pod.Status.Reason)
+			if when != "" {
+				parts := strings.Split(strings.TrimSpace(when), " ")
+				if len(parts) < 2 {
+					println(fmt.Sprintf("invalid 'when' format: %s\n", when)) // don't stop
```
banjoh: Whilst here, shall we use `klog.Errorf` instead? There is another `println` somewhere in the file

banjoh (dismissed):

banjoh (approved):

PR #1098, opened 2023-04-04
fix: Discover specs from namespaces user is allowed

### LINKED ISSUE #1079: Discovering cluster specs when user is missing required permissions on at least one namespace does not find specs
**Bug Description**

_This is a possible regression which I have not yet tested_

If a cluster has multiple support bundle specs distributed across many namespaces, there is a potential of discovering a subset of these specs when using the `--load-cluster-specs` flag.

**Expected Behavior**

Discover all specs from namespaces the user is allowed to read secrets from and log namespaces they could not search for specs.

**Steps To Reproduce**

I have the following namespaces
```
kube-system
default
app-1    # has spec
backend-app-1     # has spec
admin-tools
```

A user has RBAC permissions to list `secrets` & `configmaps` from the following namespaces
```
app-1
default
```

If the user’s kubectl’s context (kubeconfig) is configured to use `default` or any other NS they have RBACs for, they can discover specs for that NS only if they run `./support-bundle --load-cluster-specs`. This is because [this logic](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L129) will fail on checking if the user is able to read secrets & configmaps from all namespaces, and default to searching for specs on their default kubeconfig namespace.

**Additional Context**

- Troubleshoot version. `main` branch

banjoh:

If a user has limited access to read secrets and config maps from certain namespaces in a cluster, we'd need to gracefully fail when forbidden errors are caught. We'll log them and continue searching for specs in other namespaces. With this bug fix, we ensure that all specs in the cluster, that the user can read, are discovered and and used when collecting a support bundle

Fixes: #1079

## Checklist

- [ ] Yes
- [x] No


banjoh:
To test this bug fix, you would need to do the following
* Create a user and assign limited RBAC permissions. You can borrow from  https://gist.github.com/banjoh/aa07ecc29802404de1dfb2fab029222a instructions
* Create a namespace the new user can create spec secrets & config maps
* Collect a support bundle using the new user's credentials

cmd/troubleshoot/cli/run.go
```diff
@@ -346,6 +259,162 @@ the %s Admin Console to begin analysis.`
 	return nil
 }
 
+// loadClusterSpecs loads the support bundle and redactor specs from the cluster
+// based on troubleshoot.io/kind=support-bundle label selector. We search for secrets
```
DexterYan, line 263: According to https://github.com/replicatedhq/kURL/pull/4247, it should be  troubleshoot.sh, right?

xavpaice: not until we've come to a conclusion on that path

DexterYan: awesome, the PR is all good now.

DexterYan (approved):

xavpaice (approved):

Issue #1111, opened 2023-04-11
Filter incoming yaml documents to `preflight` to only respond to valid preflights


xavpaice:
There are circumstances, e.g. when using the output of `helm template`, when many documents are fed to `preflight` and only a portion of them are valid for use as specs.  We should be able to filter incoming docs, via stdin or from a URI path (including local files) such that we only process docs that are:

* match both `apiVersion: troubleshoot.sh/v1beta*` and `kind: Preflight`
* or, a secret or configMap that has the label `troubleshoot.io/kind: preflight`(or updated by #1070)

The second option here extends the work in #1110 to match not just secrets stored in-cluster, but provided as yaml as manifests on the command line.

Definition of done:

- [ ] `preflight` can be passed multiple yaml documents of varying formats, and only processes those listed above
- [x] if multiple documents/specs/secrets are selected for processing, these are merged
- [ ] if there is a `kind: Prefight` document provided as well as a secret containing a spec, these are both selected for processing and merged.
- [ ] Docs updated

Out of scope:
- there is no requirement to support both URI and stdin on the same command-line call.

danj-replicated:
this will have to be addressed in https://github.com/replicatedhq/troubleshoot/issues/1109 to avoid erroring out whenever preflight can't decode input into valid preflight manifests

xavpaice:
Let's do the first iteration where we only respond to `kind: Preflight` and not the secrets/configMaps.

xavpaice:
I'm closing this as #1114 covers the minimum scope that turned out to be what we need

PR #1114, opened 2023-04-11
Add stdin and multidoc support to preflight.

### LINKED ISSUE #1109: Allow `preflight`binary to accept stdin as well as URI and files for spec input
The `preflight` binary currently does not accept stdin as a method to pass yaml documents containing Troubleshoot specs.

This task is to add that ability.

Definition of done:

- [x] `preflight` accepts stdin such that `cat mypreflight.yaml | kubectl preflight` produces the output of preflights from mypreflight.yaml
- [x] the content of stdin can be multiple yaml documents, each containing a valid preflight(s), results are from a merged spec
- [ ] Docs are updated in troubleshoot.sh and/or docs.replicated.com

Out of scope:
- Ability to mix stdin and other URI types of input

danj-replicated:

partially addresses #1111 
Fixes #1109 

Preflight can now accept stdin alongside existing documents like so:

```bash
helm template | preflight -
```

input documents are filtered for valid yaml with the key-value pair: `kind: Preflight`
## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/preflight/run.go
+++ b/pkg/preflight/run.go
@@ -76,6 +76,12 @@ func RunPreflights(interactive bool, output string, format string, args []string
 			}
 
 			preflightContent = b
+		} else if v == "-" {
+			b, err := io.ReadAll(os.Stdin)
+			if err != nil {
+				return err
+			}
+			preflightContent = b
 		} else {
 			u, err := url.Parse(v)
 			if err != nil {
--- a/pkg/preflight/run.go
+++ b/pkg/preflight/run.go
@@ -135,7 +135,7 @@ func RunPreflights(interactive bool, output string, format string, args []string
 
 			var parsedDocHead documentHead
 
-			err := yaml.Unmarshal([]byte(doc),&parsedDocHead)
+			err := yaml.Unmarshal([]byte(doc), &parsedDocHead)
 			if err != nil {
 				return errors.Wrap(err, "failed to parse yaml")
 			}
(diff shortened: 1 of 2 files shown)


pkg/preflight/run.go
```diff
@@ -118,27 +125,48 @@ func RunPreflights(interactive bool, output string, format string, args []string
...
+			preflightContent, err = docrewrite.ConvertToV1Beta2([]byte(doc))
+			if err != nil {
+				return errors.Wrap(err, "failed to convert to v1beta2")
+			}
+
+			troubleshootclientsetscheme.AddToScheme(scheme.Scheme)
+			decode := scheme.Codecs.UniversalDeserializer().Decode
+			obj, _, err := decode([]byte(preflightContent), nil, nil)
+			if err != nil {
+				return errors.Wrapf(err, "failed to parse %s", v)
```
banjoh, line 156: Decoding is bound to fail here since troubleshoot will not know how to decode custom CRDs

danj-replicated: I don't think custom CRDs will get to this point as we skip anything that's not `kind: Preflight`

banjoh:
I ran a test using `cat spec.yaml | ./bin/preflight -`. Works like a charm. This change also brings in a good side effect. The `preflight` binary now allows using a multidoc spec

banjoh:
`cat spec.yaml | ./bin/preflight - spec-2.yaml` works too

banjoh:
An integration test would be good now

xavpaice (approved):
Looks good, tests are written

PR #1121, opened 2023-04-17
Revert "feat: Add partition type info to block device collector"

### LINKED ISSUE #1122: blockDevices host preflight fails on older linux kernels
**Bug Description**

https://github.com/replicatedhq/troubleshoot/pull/1063 introduced a change to collect PTTYPE from lsblk in the blockDevices collector. PTTYPE column is not supported on older Linux kernels (confirmed on older releases of RHEL 7 and 8). When running preflight, the check fails with the following error:

```
2023-04-17 16:16:55+00:00 [FAIL] Block Devices: Analyzer Failed: analyze: failed to get collected file: file host-collectors/system/block_devices.json was not collected
```

**Expected Behavior**

The blockDevices should be backwards compatible with all previously supported kernels.

**Steps To Reproduce**

Using image id centos-7-v20210512 on GCP with a block device attached if you run the following spec you will get the error with the v0.60.0 release of troubleshoot. This worked previous to the v0.60.0 release.

```
kind: HostPreflight
metadata:
  name: kurl-builtin
spec:
  collectors:
  - blockDevices: {}

  analyzers:
  - blockDevices:
      includeUnmountedPartitions: true
      minimumAcceptableSize: 10737418240 # 1024 ^ 3 * 10, 10GiB
      outcomes:
      - pass:
          when: ".* == 1"
          message: One available block device
      - pass:
          when: ".* > 1"
          message: Multiple available block devices
      - fail:
          message: No available unformatted block devices were found
```

**Additional Context**

  v0.60.0
- Operating system
  RHEL variants
- Operating system version
  7 or 8

emosbaugh:
Reverts replicatedhq/troubleshoot#1063

PTYPE is not supported on older kernels

```
$ sudo lsblk --bytes --pairs -o NAME,KNAME,PKNAME,TYPE,MAJ:MIN,SIZE,FSTYPE,MOUNTPOINT,SERIAL,RO,RM,PTTYPE
lsblk: unknown column: PTTYPE
$ cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)
```

### DIFF
--- a/pkg/analyze/host_block_devices.go
+++ b/pkg/analyze/host_block_devices.go
@@ -116,7 +116,7 @@ func (a *AnalyzeHostBlockDevices) Analyze(getCollectedFileContents func(string) 
 func compareHostBlockDevicesConditionalToActual(conditional string, minimumAcceptableSize uint64, includeUnmountedPartitions bool, devices []collect.BlockDeviceInfo) (res bool, err error) {
 	parts := strings.Split(conditional, " ")
 	if len(parts) != 3 {
-		return false, fmt.Errorf("expected exactly 3 parts, got %d", len(parts))
+		return false, fmt.Errorf("Expected exactly 3 parts, got %d", len(parts))
 	}
 
 	rx, err := regexp.Compile(parts[0])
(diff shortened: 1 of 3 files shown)


xavpaice:
related: #970

chris-sanders:
Rather than just completely reverting this, can/should we just put a graceful fail in and reduce the data collected if the lsblk command doesn't support this column (or any column if that's something we can reasonably ascertain) ?

xavpaice (approved):

xavpaice:
I've approved this one as is - I would like to collect what the kernel allows, rather than fail the collection.  However, this revert is to overcome a regression as a result of the earlier change, and for that reason I'll approve/merge it and cut a release to unblock folks.

The original bug #970 should get some notes about this approach and the change be worked over there.

PR #1133, opened 2023-04-21
Use strings.TrimPrefix

### LINKED ISSUE #1129: strings.TrimLeft is used instead of TrimPrefix
**Bug Description**

The `PortForward` method incorrectly uses `strings.TrimLeft`. The "htps:/" argument is treated as a cutset (set of characters) and not as a prefix.

https://github.com/replicatedhq/troubleshoot/blob/6493064e4599adb2a4f1255d90052872e73f5d8c/pkg/k8sutil/portforward.go#L23

The `hostIP` will be set to `esis.example.com` instead of `thesis.example.com` for `config.Host` `https://thesis.example.com`.

Fix is to use `strings.TrimPrefix`.

AKARSHITJOSHI:
Fixes #1129

### DIFF
--- a/pkg/k8sutil/portforward.go
+++ b/pkg/k8sutil/portforward.go
@@ -20,7 +20,7 @@ func PortForward(config *restclient.Config, localPort int, remotePort int, names
 	}
 
 	path := fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward", namespace, podName)
-	hostIP := strings.TrimLeft(config.Host, "htps:/")
+	hostIP := strings.TrimPrefix(config.Host, "htps:/")
 	serverURL := url.URL{Scheme: "http", Path: path, Host: hostIP}
 	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: roundTripper}, http.MethodPost, &serverURL)
 


danj-replicated (changes requested):
Hi @AKARSHITJOSHI, Thanks for the contribution. 

Just looking over this change and I'm not sure it'll work as intended. 

`strings.TrimLeft` takes a cutset (you can think of this as a list of characters to remove), while `strings.TrimPrefix` takes the actual string to trim.

 `htps:/` won't match the scheme here and potentially cause an error by doubling up the scheme when we construct the URL on line 24.

here's an example of the outputs and resultant URL structs of what we have here: 

```go
oldURL := "https://172.0.0.1"

url1 := strings.TrimLeft(oldURL,"htps:/")
// 172.0.0.1

url2 := strings.TrimPrefix(oldURL,"htps:/")
// https://172.0.0.1

url3 := strings.TrimPrefix(oldURL,"https://")
// 172.0.0.1

fmt.Printf("%+v\n", url.URL{Scheme: "http", Path: "/", Host: url1 })
// {Scheme:http Opaque: User: Host:172.0.0.1 Path:/ RawPath: OmitHost:false ForceQuery:false RawQuery: Fragment: RawFragment:}

fmt.Printf("%+v\n", url.URL{Scheme: "http", Path: "/", Host: url2 })
// {Scheme:http Opaque: User: Host:https://172.0.0.1 Path:/ RawPath: OmitHost:false ForceQuery:false RawQuery: Fragment: RawFragment:}

fmt.Printf("%+v\n", url.URL{Scheme: "http", Path: "/", Host: url3 })
// {Scheme:http Opaque: User: Host:172.0.0.1 Path:/ RawPath: OmitHost:false ForceQuery:false RawQuery: Fragment: RawFragment:}
```

equally I'm not sure simply trimming `https://` as a fixed prefix is the best solution. Unless we can be certain that the URL returned from `config.Host` will  _always_ be https we should find a way to safely parse this URL in a scheme agnostic way.

banjoh:
Hi @AKARSHITJOSHI. Just checking if you needed to address comments made in the PR

AKARSHITJOSHI:
> Hi @AKARSHITJOSHI. Just checking if you needed to address comments made in the PR

Hi currently I'm not working on this . You can take it forward

xavpaice:
I'll close this PR now, since it's not active.  We can revisit the original issue separately.

PR #1144, opened 2023-05-04
Consolidated CLI design proposal

### LINKED ISSUE #934: Design proposal for a consolidated Troubleshoot CLI

xavpaice:

## Checklist

- [ ] Yes
- [x] No


docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+Available Commands:
+  help        This help text
+  collect     Run collectors, redactors and analyzers, store the result
+  completion  Generate the autocompletion script for the specified shell
+  analyze     Analyze an existing support bundle
+  redact      Run redactors across an existing support bundle
+  preflight   Run collectors, and analyzers, and provide a pass/fail preflight result with explanation
+  shell       Run sbctl shell using a support bundle
+  version     Print the current version and exit
```
danj-replicated: There's some inconsistency in the design here that I'd like to see clarified. 

I'm a fan of the subcommand structure listed in the example help text but the detailed design suggests that functionality lives under flags?

danj-replicated (dismissed):

docs/design/proposal-troubleshoot-consolidated-cli.md
xavpaice: let's get the most logical suggestion rolled consistently right through.  You should be able to add commits to the branch (it's not a fork).

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,199 @@
...
+## Non-Goals
+
+## Background
+
+## High Level Design
+
+Functions of `support-bundle`, `preflight`, `analyze`, `redact`, and `sbctl` binaries/tools should be rolled together into a single `troubleshoot` CLI plugin that can perform all necessary functions of the same.
+
+`troubleshoot` CLI plugin should be able to report on the version that is installed in the CLI, and any support bundle generated with `troubleshoot` should report its build/version inside the archive it generates.
+
```
xavpaice, line 29: should we reconsider the use of the word 'plugin', and maybe add that the `troubleshoot` binary would also be available as a Krew plugin for kubectl?

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,199 @@
...
+The Consolidated cli should be responsible for handling the end-to-end running of this process in the `cmd` leaving the importable `pkg` free to focus on defining the individual collect, redact analyze steps.
+
+this allows us to move functions out from behind `CollectSupportBundleFromSpec` so they can be easier called from other projects importing troubleshoot.
+
+those functions could be then called as:
+
+- `troubleshoot.Collect()`
+- `troubleshoot.Redact()`
+- `troubleshoot.Analyze()`
+
```
xavpaice, line 71: suggest in order to deprecate the `CollectSupportBundleFromSpec()` function, we replace it with a wrapper that calls the cmd - so that to someone importing the function it looks the same, but the logic inside is moved to a more suitable place

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,199 @@
...
+  This should not only clearly state any reasons for failing, but also use standardised exit codes that can be used by automation tools.
+
+- Interact with an existing support bundle with kubectl.
+
+  `troubleshoot inspect support-bundle-12-12-2001.tar.gz`
+
+  This should set the `KUBECONFIG` environment variable, as well as a `TROUBLESHOOT_SHELL` environment variable before spawning a subshell.
+  This is to allow prompts and shell environments to be able to detect that they're running in a troubleshoot spawned shell, much like tmux et-al.
+
```
xavpaice, line 126: do we also want to include a 'serve' type of option for this outside the shell?  Not sure there's big value in it, we should maybe just document that folks can use the same KUBECONFIG env.

xavpaice (commented):
Some minor comments, let's flesh this out a bit now with some technical detail on what functions need to move to where and how.

Particularly want to know about breaking changes (i.e. changes to functions that other projects are known to import) so that we can ensure these are agreed up-front.

Also; the existing CLI programs need to continue to work, let's identify what needs to happen to achieve that.

The goal here is that we agree on the desired outcome, the breaking changes that introduces, and the path to get to that point.

xavpaice (commented):
Some minor comments, let's flesh this out a bit now with some technical detail on what functions need to move to where and how.

Particularly want to know about breaking changes (i.e. changes to functions that other projects are known to import) so that we can ensure these are agreed up-front.

Also; the existing CLI programs need to continue to work, let's identify what needs to happen to achieve that.

The goal here is that we agree on the desired outcome, the breaking changes that introduces, and the path to get to that point.

docs/design/proposal-troubleshoot-consolidated-cli.md
danj-replicated: if we intend to have an end-to-end process function importable like this we could just leave this alone? 

my thoughts were that the end-to-end process should live in the `cmd` instead which ideally you don't import

docs/design/proposal-troubleshoot-consolidated-cli.md
danj-replicated: hmm. yeh I can't see much value for use by a human but it could be useful for writing automation against.

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,240 @@
...
+## High Level Design
+
+Functions of `support-bundle`, `preflight`, `analyze`, `redact`, and `sbctl` binaries/tools should be rolled together into a single `troubleshoot` CLI that can perform all necessary functions of the same.
+
+`troubleshoot` CLI should be able to report on the version that is installed in the CLI, and any support bundle generated with `troubleshoot` should report its build/version inside the archive it generates.
+
+## Detailed Design
+
+- TODO: specify the public APIs we expect users of Troubleshoot to consume, e.g.
```
xavpaice: looks like we can remove this TODO now :)

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,249 @@
...
+- Interact with an existing support bundle with kubectl.
+
+  `troubleshoot inspect support-bundle-12-12-2001.tar.gz`
+
+  This should set the `KUBECONFIG` environment variable, as well as a `TROUBLESHOOT_SHELL` environment variable before spawning a subshell.
+  This is to allow prompts and shell environments to be able to detect that they're running in a troubleshoot spawned shell, much like tmux et-al.
+
+  `troubleshoot inspect --non-interactive support-bundle-12-12-2001.tar.gz`
+
+  this should behave much like the `sbctl serve` command does today, printing the path to it's kubeconfig location to stdout or optionally take a flag to specify the kubeconfig location.
```
xavpaice: I don't follow what you mean by the flag to specify kubeconfig location - doesn't inspect make a temporary kubeconfig, and we'd not want to overwrite `.kube/config`?

does `--non-interactive` mean "no shell please"?  Maybe want to skip that from the first example above.

CpuID: > does --non-interactive mean "no shell please"?

that's how I interpret it 👍

CpuID: > this should behave much like the `sbctl serve` command does today, printing the path to it's kubeconfig location to stdout or optionally take a flag to specify the kubeconfig location.

the concept here makes sense to me 👍 either make a temp kubeconfig and output an `export KUBECONFIG=...` var, or allow a CLI arg to say where to write it out, and then output the `export KUBECONFIG=specified-location` instead

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,249 @@
...
+
+```
+Kubernetes diagnostics assistant
+
+Usage:
+  troubleshoot [command]
+
+Available Commands:
+  help        This help text
+  collect     Run collectors, redactors and analyzers, store the result
```
CpuID: a thought: should we always run analyzers AOT (ahead of time) and store the results...? or should we JIT execute analyzers whenever you load up a bundle? as that would allow running against the latest analyzers on old bundles etc.

obviously you can still use `analyze` to re-run against an existing bundle, but does it make sense to run it upfront at all every time...?

we could technically still have a `troubleshoot support-bundle` command that ran collectors, redactors and analyzers and gave you the interactive UI showing the state of analysis still in a single subcommand...

CpuID (changes requested):
couple of inline comments, otherwise looking good from my standpoint 👍 

I do wonder if we should JIT analyzers instead of AOT execute against them though... I feel like more recent analyzers every time might be better/safer/more informative, in theory...

docs/design/proposal-troubleshoot-consolidated-cli.md
danj-replicated: yeh so my thoughts here are:
- if you're a human running `kubectl inspect` it's very likely that you're going to want to run kubectl to interact with the bundle. thus defaulting to a shell provisioned with the suitable env-vars removes complexity for people just wanting something to work. 
- if you're writing automation around `kubectl inspect` you might not want it to spawn a nested shell, and extracting the kubeconfig from stdout generates extra handling for scripts. being able to tell troubleshoot where you want to store the kubeconfig allows us to guarantee a location for the kubeconfig that can be used for other tooling in a script / automation pipeline. (i.e a github action)

docs/design/proposal-troubleshoot-consolidated-cli.md
danj-replicated: so to maintain functionality for vendor portal we'd still need to run analyzers up front to generate the analasys.json.

it's worth thinking about the impact that JIT analasys would have on vendor portal

docs/design/proposal-troubleshoot-consolidated-cli.md
CpuID: yep, both use cases are valid 👍 and the UX feels right for each the way you've noted it (`shell` vs `serve` behaviour)

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,243 @@
...
+
+To this end we should create a new `pkg` that targets the functionality provided by the `collect` and `analyze` packages such as `collect.runHostCollectors()` and provide a stable API to be used by the CLI and other projects. At the same time we can unpublish functionality that should not be exposed (such as the individual collector and analyser functions) and mark code for deprecation.
+
+This new package should be kept as minimal as possible. and serve only as an interface to private functions in the other packages.
+
+Once the stable API is ready we can instruct projects like kurl to target that and work on removing code marked for deprecation.
+
+The functionality we want to expose via this api is:
+
+- `CollectBundle(spec{}) (bundlePath string, error)`
```
banjoh: _This is more of an implementation detail._

For the public API function parameters, we should lean towards having a `struct` as a single parameter rather than inlined parameters.

Preference
```go
type CollectOptions struct {
	SpecURIs []string
	AnotherArg string
	// Future args come here
}

func CollectBundle(opt CollectOptions) (bundlePath string, error)
```

instead of 
```go
func CollectBundle(specURIs string[], anotherArg string) (bundlePath string, error)
```

With this approach the function signature remains constant but we are able to pass in new parameters over time, future proofing the API.

banjoh: _Another implementation detail_

I think we should have all APIs accept a `context.Context`. This is a good golang pattern which we should follow

```go
RedactBundle(ctx context.Context, opt RedactOptions) error
```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,243 @@
...
+The existing package entrypoint we use from the support-bundle command is `supportbundle.CollectSupportBundleFromSpec()`
+which is an end-to-end function that runs collectors, redactors and then analysis.
+
+The Consolidated cli should be responsible for handling the end-to-end running of this process in the `cmd` leaving the importable packages free to focus on defining the individual collect, redact analyze steps.
+
+To this end we should create a new `pkg` that targets the functionality provided by the `collect` and `analyze` packages such as `collect.runHostCollectors()` and provide a stable API to be used by the CLI and other projects. At the same time we can unpublish functionality that should not be exposed (such as the individual collector and analyser functions) and mark code for deprecation.
+
+This new package should be kept as minimal as possible. and serve only as an interface to private functions in the other packages.
+
+Once the stable API is ready we can instruct projects like kurl to target that and work on removing code marked for deprecation.
```
banjoh: _Implementation detail_

Eventually, once the API is stable and all deprecated code removed, all internal logic should be moved from the `pkg` package to an [`internal` package](https://go.dev/doc/go1.4#internalpackages).

For reference of how this would look like, please take a look at https://github.com/derailed/k9s/tree/master/internal project that uses this pattern extensively.

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,243 @@
...
+The functionality we want to expose via this api is:
+
+- `CollectBundle(spec{}) (bundlePath string, error)`
+  - collect a support bundle from a spec.
+  - takes a parsed spec struct as an parameter.
+  - returns a path to the bundle directory and errors.
+  - to minimise IO and increase collection speed. redactors should be run inline, redacting data in memory before it's saved to a bundle.
+- `RedactBundle(bundlePath string) error`
+  - redact an already collected bundle, takes a path to the bundle as an parameter.
+  - returns any errors
```
mhrabovcin: Could this API accept list of redactors with non-hardcoded defaults. I think that currently there is no way to disable some default collectors, like IP addresses.

xavpaice: IP addresses was removed from the defaults, but the others could use more granular switches, for sure.

docs/design/proposal-troubleshoot-consolidated-cli.md
xavpaice: In k9s, this really only changes the import path rather than much else.  The package files still start with, e.g. `package client`.  Kubernetes itself doesn't use this pattern though there are many packages not intended to be imported by other projects.  Is there something about using the 'internal' directory other than the name making it clearer as to the intent of these packages?  I'd rather avoid complexity, but not sure if using 'internal' avoids complexity or creates it.

xavpaice:
Couple of usage suggestions in #1198 and #1199 that are likely able to be taken care of via this CLI.

docs/design/proposal-troubleshoot-consolidated-cli.md
banjoh: > Is there something about using the 'internal' directory other than the name making it clearer as to the intent of these packages? I'd rather avoid complexity, but not sure if using 'internal' avoids complexity or creates it.

Its [golang's way of enforcing](https://go.dev/doc/go1.4#internalpackages) (at a compiler level) the boundary of public vs private packages. In troubleshoot for example, `github.com/replicatedhq/troubleshoot/internal/traces` package cannot be imported by any other project besides troubleshoot. If KOTS for example tried to import this package, the code would not compile.

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,243 @@
...
+- `ArchiveBundle(bundlePath string, compressionMethod string, destination string) error`
+  - generates a tar archive of the bundle directory at the specified path with optional compression
+  - takes bundle path, compression method and destination as parameters.
+  - returns errors
+- `ParseSpecFile(specFile string) (spec{},error)`
+  - parses a spec file and returns an unmarshalled troubleshoot spec struct.
+  - could be used for linting
+- `ServeBundle(bundlePath string, kubeconfig string, port int)`
+  - starts a sbctl server using the specified bundle and port, outputting a kubeconfig at a specified location.
+
```
banjoh:
```suggestion
- `LoadFromStrings(rawSpecs ...string) (*TroubleshootV1beta2Kinds, error)`
  - accepts a list of valid yaml documents and extracts all troubleshoot objects from the documents. This includes `Secret` and `ConfigMap` objects that have troubleshoot specs.

```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,292 @@
...
+```go
+// Load loads specs defined by the LoadOptions struct and returns a TroubleshootKinds struct
+func Load(LoadOptions) (TroubleshootKinds,err){}
+kinds, err := Load(LoadOptions)
+
+// Bundle.Collect runs collections defined in TroubleshootKinds passed through CollectOptions
+func (bundle *Bundle) Collect(CollectOptions) (error) {}
+
+func (bundle *Bundle) Load(LoadBundleOptions) (error) {}
```
xavpaice: can we get a comment on what Bundle.Load does?

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// Note: this is almost identical to `CollectOptions` for now but remains separate to enable easier addition of redact specific options at a later date
+
+type ServeOptions struct {
+  Address string // address to listen on including port (0.0.0.0:8080)
+  ConfigPath string // optional path to store generated kubeconfig
+}
+```
+
```
xavpaice: would love to see an example of how a bundle is created with a spec - the flow from loading the spec, to having a Bundle that's been collected and is ready to redact, analyze, etc.

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,350 @@
...
+func (bundle *Bundle) Redact(RedactOptions) error {}
+
+// Bundle.Archive produces an archive from a bundle on disk with options passed in ArchiveOptions
+func (bundle *Bundle) Archive(ArchiveOptions) error {}
+
+// Bundle.Serve starts an sbctl-like server with options defined in ServeOptions
+func (bundle *Bundle) Serve(ServeOptions) error {}
+
+// AnalysisResults.Save saves the results of the analysis according to SaveAnalysisOptions
+func (results *AnalysisResults) Save(SaveAnalysisOptions) error {}
```
banjoh: Is this meant to save the results in an existing bundle?

Just wondering if `func (bundle *Bundle) Analyze(AnalyzeOptions) (AnalysisResults, error) {}` does this already, or whether it should save to a bundle.

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,350 @@
...
+  SearchCluster bool // toggle for searching cluster from context for troubleshoot objects
+  ProgressChan chan // a channel to write progress information to
+}
+
+type LoadBundleOptions struct {
+    Path string // Path to archive or directory of bundle files
+}
+
+type CollectOptions struct {
+  DefaultCollectors []Collector // list of collectors to run regardless of what's in the spec
```
banjoh: With the default collectors/redactors/analysers, I think we should not treat them in any special way. I see it working like so

```go
defaults := TroubleshootKinds{}
defaults.CollectorsV1Beta2 = append(kinds.CollectorsV1Beta2, other.CollectorsV1Beta2...)
defaults.AnalyzersV1Beta2 = append(kinds.CollectorsV1Beta2, other.AnalyzersV1Beta2...)
defaults.RedactorsV1Beta2 = append(kinds.CollectorsV1Beta2, other.RedactorsV1Beta2...)

kinds := TroubleshootKinds{}
kinds.Add(defaults)

opt := CollectOptions{Specs: kinds}
```

banjoh:
```suggestion
```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,350 @@
...
+}
+// Note: this is almost identical to `CollectOptions` for now but remains separate to enable easier addition of redact specific options at a later date
+
+type ServeOptions struct {
+  Address string // address to listen on including port (0.0.0.0:8080)
+  ConfigPath string // optional path to store generated kubeconfig
+}
+
+type AnalyzeOptions struct {
+  DefaultAnalyzers []Analyzer // list of default analyzers to run regardless of what's defined in spec.
```
banjoh:
```suggestion
```

docs/design/proposal-troubleshoot-consolidated-cli.md
banjoh: Troubleshoot library will expose a `specs.DefaultSpecs` variable somewhere which `support-bundle` & `preflight` binary would consume.

docs/design/proposal-troubleshoot-consolidated-cli.md
danj-replicated: hmm I was thinking it would save the results to a file in a specified format, but I wonder if we should leave that to the implementing application instead?

banjoh: Yer, I see what you mean. There are a number of use cases here.

* Analyse and save results in the bundle
* Re-analyse an existing bundle but do not modify the bundle (e.g after loading an existing bundle which has results already, you may want to retain the old ones as-is)

Maybe these decisions can be driven by `AnalyzeOptions` e.g

```go
type AnalyzeOptions struct {
	Specs TroubleshootKinds // list of specs to extract analysers from
	SaveToBundle bool // save the results to a bundle. Or should we just check if BundlePath is empty and infer from that?
	BundlePath string // path to save the bundle to
}
```

banjoh (dismissed):

banjoh (dismissed):

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
...
+
+For example, running a set of redactors or analyzers on an existing support bundle without having to re-run collection.
+
+### Troubleshoot top-level commands to subcommands
+
+All top level commands (`support-bundle`, `preflight`, `analyze`, `redact`, and `sbctl`) of the Troubleshoot project will now be subcommands e.g `sbctl` is planned to be `troubleshoot inspect`. Some flags and subcommands names may change to best fit completeness of the CLI interface e.g `--interactive=false` will be `--no-input`. Putting the CLI interface together should aim to follow best practises from well known guidelines such as https://clig.dev/ and other mature projects.
+
+- `troubleshoot inspect` - This subcommand will be equivalent to the `sbctl` command.
+- `troubleshoot collect` - This subcommand will be equivalent to the `support-bundle` command.
```
xavpaice: some discussion around this - let's keep it consistent and have this `troubleshoot support-bundle`

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
...
+- `ServeBundle()`
+  - serves a `bundle` object via a kubernetes API server for programmatic access by tools such as `kubectl` or the `client-go` library.
+
+### CLI Usage patterns
+
+This is a non-exhaustive list of usage patterns, but it should help guide how functionality from the current CLIs will be exposed in the new `troubleshoot` CLI.
+
+- generate a support bundle
+
+  `troubleshoot collect supportbundle.yaml`
```
xavpaice:
```suggestion
  `troubleshoot support-bundle supportbundle.yaml`
```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
...
+
+  `troubleshoot collect supportbundle.yaml secrets/default/kotsadm-appslug-supportbundle`
```
xavpaice:
```suggestion
  `troubleshoot support-bundle supportbundle.yaml secrets/default/kotsadm-appslug-supportbundle`
```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
...
+
+  `troubleshoot collect https://kots.io`
```
xavpaice:
```suggestion
  `troubleshoot support-bundle https://kots.io`
```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
...
+
+- if no bundle URI/location is specified, search for a spec in-cluster
+
+  `troubleshoot collect`
```
xavpaice:
```suggestion
  `troubleshoot support-bundle`
```

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
+# Consolidated Troubleshoot CLI
+
+## Goals
+
+- Consolidate all top level Troubleshoot commands (`support-bundle`, `preflight`, `analyze`, `redact`, and `sbctl`) into one exposing subcommands of all the functionality that was implemented by the previous commands.
+- Ensure functional backward compatibility of newly introduces interfaces (CLI and Public APIs)
+
+## Non-Goals
+
+- Maintain backward compatibility of the CLI interface and the public APIs (new interfaces will be created)
```
xavpaice: let's add a note to the effect that the existing CLI will remain supported until such time as the new CLI is in common use at least by the Replicated product set.

docs/design/proposal-troubleshoot-consolidated-cli.md
```diff
@@ -0,0 +1,166 @@
...
+- `troubleshoot inspect` - This subcommand will be equivalent to the `sbctl` command.
+- `troubleshoot collect` - This subcommand will be equivalent to the `support-bundle` command.
+- `troubleshoot preflight` - This subcommand will be equivalent to the `preflight` command.
+- `troubleshoot redact` - This subcommand will be equivalent to the `redact` command.
+- `troubleshoot analyze` - This subcommand will be equivalent to the `analyze` command.
+
+### Public APIs
+
+The existing package entrypoint we use from the support-bundle command is `supportbundle.CollectSupportBundleFromSpec()`
+which is an end-to-end function that runs collectors, redactors and then analysis.
```
xavpaice: let's be sure to mention that we do not intend to break this entrypoint immediately, but that it is intended that we use this API to allow future versions to make breaking changes to that function.

xavpaice (commented):
couple of notes, based on some discussion today

xavpaice:
Closing until we revisit the proposal

PR #1172, opened 2023-05-23
Capture `runPod` status and events before deleting the pod

diamonwiggins:

This change adds a function to capture the current status and events of a Pod created by the `runPod` collector. This is useful in the event you need to capture details about the scheduling of the Pod as well as other use cases. 

## Checklist

- [ ] Yes
- [x] No


diamonwiggins:
Screenshot 2023-05-22 at 8 41 36 PM
Screenshot 2023-05-22 at 8 42 09 PM

pkg/collect/run_pod.go
```diff
@@ -378,3 +393,37 @@ func RunPodLogs(ctx context.Context, client v1.CoreV1Interface, podSpec *corev1.
...
+	}
+
+	podBytes, err := json.MarshalIndent(podStatus, "", "  ")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal pod status")
+	}
+
+	podEventBytes, err := json.MarshalIndent(podEvents, "", "  ")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal pod status")
```
banjoh: `failed to marshal pod events`

pkg/collect/run_pod.go
```diff
@@ -93,11 +112,14 @@ func (c *CollectRunPod) Collect(progressChan chan<- interface{}) (CollectorResul
 
 	select {
 	case <-time.After(timeout):
-		return nil, errors.New("timeout")
-	case result := <-resultCh:
+		return result, errors.New("timeout")
+	case output := <-resultCh:
+		for k, v := range output {
```
banjoh: result.AddResult(output)

banjoh (commented):
All's good with the change introduced by this PR.

I think the goroutine used to wrap the timeout is unnecessary though. The change can be done in a separate PR if it feels like scope creep.

banjoh:
Some other improvements I've picked up
* The pod is not given enough time to run. We start cleaning up immediately (~ 1s after) the correct status is detected
* When we delete the pod, we probably should ignore the grace period and wait (with timeout) for the pod to get deleted. A new pod that gets created with the same name in the same namespace would fail if the old one is still being deleted.

(Since the collector is working fine ATM, I'd add these as improvement comments/TODOs in the code. If a bug was to be reported, such comments would be helpful in knowing where to start)

diamonwiggins:
> The pod is not given enough time to run. We start cleaning up immediately (~ 1s after) the correct status is detected

It looks like in `savePodLogs` for the `PodLogOptions` we use `Follow`, so perhaps that's why this is working as expected today? But I understand where you're coming from. I wonder if expected behavior for a successful `runPod` should be that the Pod phase is `Succeeded`? 🤔

banjoh:
> 

I ran the collector below and as you can see the pod is left terminating long after collection is done. I'd expect some cleaning up to take place before progressing. I ran this using your branch.

```yaml
kind: SupportBundle
metadata:
  name: example
spec:
  collectors:
    - runPod:
        collectorName: "my-counter"
        timeout: 5s
        podSpec:
          containers:
          - name: my-counter
            image: ubuntu
            command: ["/bin/bash", "-c", 'for i in {1..30}; do echo "count $i"; sleep 1; done']
```

```sh
[evans] $ ./bin/support-bundle spec.yaml -v0

 * failed to run collector: run-pod/my-counter: timeout
 Collecting support bundle ⠼ cluster-info
support-bundle-2023-05-23T17_35_49.tar.gz

Suceeded (S), eXcluded (X), Failed (F)
run-pod/my-counter (F) : 5,047ms
cluster-resources (S)  : 2,128ms
cluster-info (S)       : 4ms

In-cluster collectors : 488ms

Suceeded (S), eXcluded (X), Failed (F)
No analyzers executed

Duration: 7,686ms
✔ ~/repos/replicated/troubleshoot [pr/1172 L|…2⚑ 3]
[evans] $ k get pods
NAME         READY   STATUS        RESTARTS   AGE
my-counter   1/1     Terminating   0          16s
```

banjoh:
Hmm. Modifying the grace period to "force" deleting the pod immediately might be the trick here. I made this change [here](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/run_pod.go#L56)

```
    defer func() {
        if err := client.CoreV1().Pods(pod.Namespace).Delete(context.Background(), pod.Name, metav1.DeleteOptions{}); err != nil {
        g := int64(0)
        if err := client.CoreV1().Pods(pod.Namespace).Delete(context.Background(), pod.Name, metav1.DeleteOptions{
            GracePeriodSeconds: &g,
        }); err != nil {
            klog.Errorf("Failed to delete pod %s: %v", pod.Name, err)
        }
    }()
```


Results

```sh
[evans] $ ./bin/support-bundle spec.yaml -v0 && kubectl get pods

 * failed to run collector: run-pod/my-counter: timeout
 Collecting support bundle ⠴ my-counter
support-bundle-2023-05-23T17_40_09.tar.gz

Suceeded (S), eXcluded (X), Failed (F)
run-pod/my-counter (F) : 5,063ms
cluster-resources (S)  : 2,131ms
cluster-info (S)       : 2ms

In-cluster collectors : 511ms

Suceeded (S), eXcluded (X), Failed (F)
No analyzers executed

Duration: 7,735ms
No resources found in default namespace.
```

banjoh:
Here is the other use case where the spec name has the same name, leading to the same pod name. The error is debatable cause the collector should probably be named differently. Perhaps we should raise a warning/error when specs have the same name? Either way, here are the results

This is without the grace period patch. With the patch, the pod object gets deleted and the second pod gets scheduled as expected.

```yaml
kind: SupportBundle
metadata:
  name: example
spec:
  collectors:
    - runPod:
        collectorName: "my-counter"
        timeout: 5s
        podSpec:
          containers:
          - name: my-counter
            image: ubuntu
            command: ["/bin/bash", "-c", 'for i in {1..2}; do echo "count $i"; sleep 1; done']
    - runPod:
        collectorName: "my-counter"
        timeout: 5s
        podSpec:
          containers:
          - name: my-counter
            image: ubuntu
            command: ["/bin/bash", "-c", 'for i in {1..2}; do echo "count again $i"; sleep 1; done']
```

```sh
✔ ~/repos/replicated/troubleshoot [pr/1172 L|✚ 1…2⚑ 3]
[evans] $ ./bin/support-bundle spec.yaml -v0 && kubectl get pods

 * failed to run collector: run-pod/my-counter: failed to run pod: failed to create pod: object is being deleted: pods "my-counter" already exists
 Collecting support bundle ⠏ cluster-info
support-bundle-2023-05-23T17_48_39.tar.gz

Suceeded (S), eXcluded (X), Failed (F)
run-pod/my-counter (S) : 4,373ms
cluster-resources (S)  : 2,128ms
run-pod/my-counter (F) : 3ms
cluster-info (S)       : 2ms

In-cluster collectors : 598ms

Suceeded (S), eXcluded (X), Failed (F)
No analyzers executed

Duration: 7,133ms
NAME         READY   STATUS        RESTARTS     AGE
my-counter   1/1     Terminating   1 (1s ago)   5s
```

diamonwiggins:
@banjoh Let's address the collisions/pod delete stuff in a separate issue. Would you mind creating one?

pkg/collect/run_pod.go
```diff
@@ -48,6 +50,13 @@ func (c *CollectRunPod) Collect(progressChan chan<- interface{}) (CollectorResul
 		return nil, errors.Wrap(err, "failed to create client from config")
 	}
 
+	if c.Collector.CollectorName == "" {
+		c.Collector.CollectorName = "run-pod"
+		if c.Collector.Name != "" {
+			c.Collector.CollectorName = c.Collector.Name
+		}
+	}
+
```
diamonwiggins: @banjoh @xavpaice And last thing, on [troubleshoot.sh](https://troubleshoot.sh/docs/collect/run-pod/) for the `runPod` collector we document the output path to be `collector-name/collector-name.log`, but that hasn't been the case for likely a very long time due to a bug in whether or not the user specifies `collectorName` or `name` in the spec. This PR includes the changes to fix that, but it may break what has been expected behavior in `runPod` files being left at the root of the bundle. So we either need to leave as is and correct the documentation, or fix it. I lean towards the latter.

diamonwiggins: https://troubleshoot.sh/docs/analyze/regex/#example-analyzer-definition-for-regexgroups shows an example of both `name` and `collectorName` being used. Our docs are pretty confusing on this at the moment. I think this collector should be consistent with all the others where `collectorName` is used for any directory and file names

banjoh: > I think this collector should be consistent with all the others where collectorName is used for any directory and file names

Agreed.

banjoh: We will need to document the change as breaking by fixing something that was already broken, I guess :)

pkg/collect/run_pod.go
```diff
@@ -378,3 +391,37 @@ func RunPodLogs(ctx context.Context, client v1.CoreV1Interface, podSpec *corev1.
 
 	return ioutil.ReadAll(logs)
 }
+
+func savePodDetails(output CollectorResult, bundlePath string, clientConfig *rest.Config, pod *corev1.Pod, runPodCollector *troubleshootv1beta2.RunPod) (CollectorResult, error) {
+	client, err := kubernetes.NewForConfig(clientConfig)
```
banjoh: You could reuse an [already created client](https://github.com/replicatedhq/troubleshoot/blob/diamonwiggins/save-runpod-details/pkg/collect/run_pod.go#L48) instead of creating a new one

pkg/collect/run_pod.go
```diff
@@ -378,3 +391,37 @@ func RunPodLogs(ctx context.Context, client v1.CoreV1Interface, podSpec *corev1.
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed create client from config")
+	}
+
+	ctx := context.Background()
```
banjoh: Pass in this context as an argument to be provided by the caller. There is one [here already](https://github.com/replicatedhq/troubleshoot/blob/diamonwiggins/save-runpod-details/pkg/collect/run_pod.go#L46)

pkg/collect/run_pod.go
```diff
@@ -378,3 +391,37 @@ func RunPodLogs(ctx context.Context, client v1.CoreV1Interface, podSpec *corev1.
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal pod status")
+	}
+
+	podEventBytes, err := json.MarshalIndent(podEvents, "", "  ")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal pod events")
+	}
+
+	output.SaveResult(bundlePath, filepath.Join(runPodCollector.Name, fmt.Sprintf("%s.json", runPodCollector.CollectorName)), bytes.NewBuffer(podBytes))
```
banjoh: We ignore errors returned by these `SaveResult` quite often. I'd prefer we log these errors instead of "swallowing" them. I wonder whether we should have a `SaveResultLogError()` or `SaveResultSafe()` function that internally wraps `SaveResult` but logs any error and has no return value.

banjoh: > SaveResultSafe

I couldn't find a naming convention for this pattern so we'll probably have to invent something and document the function to say what its doing

banjoh: Perhaps we should rename `SaveResult` to `SaveResultErr` to indicate it returns an error?

pkg/collect/run_pod.go
diamonwiggins: Going to revert these changes here, and create a separate issue

diamonwiggins:
Should have addressed all the feedback. Will write a few a tests for `savePodDetails` and then re-request review.

banjoh (approved):

banjoh (approved):

PR #1174, opened 2023-05-24
feat: add k8s custom metrics collector

ahmedElqutb:

Here we introduce a collector that collects custom k8s metrics from `custom.metrics.k8s.io/v1beta1/`
and would save them in the bundle under the directory metrics:

```
metrics
|_ <resource_type> # can be Namespace, Pod, Node etc.
   |_ <metric_name> # raw metric name truncated from the resource name as per custom.metrics.k8s.io/v1beta1/ e.g. "namespaces/cpu_usage" would result in metric_name "cpu_usage"
      |_ <namespace>.json or <non_namespaced_object>.json # values for namespaced resources metrics are saved together in a file named after the namespace. For non-namespaced resources, each resource has their metric values in a separate file named after the resource. 
```

and the output is saved as:
      
```
[
  {
    "DescribedObject": {
      "Kind": "Pod",
      "Namespace": "kommander",
      "Name": "stack-alertmanager-0",
      "UID": "",
      "APIVersion": "/v1",
      "ResourceVersion": "",
      "FieldPath": ""
    },
    "Metric": {
      "Name": "",
      "Selector": null
    },
    "Timestamp": "2023-05-23T14:04:48Z",
    "WindowSeconds": null,
    "Value": "1m"
  },
]
```
      
the collector executes multiple metric requests that have the format

```
type MetricRequest struct {
	// Namespace for which to collect the metric values, empty for non-namespaces resources.
	Namespace string `json:"namespace,omitempty" yaml:"namespace,omitempty"`
	// ObjectName for which to collect metric values, all resources when empty.
	// Note that for namespaced resources a Namespace has to be supplied regardless.
	ObjectName string `json:"objectName,omitempty" yaml:"objectName,omitempty"`
	// MetricName name of the MetricValueList as per the APIResourceList from
	// custom.metrics.k8s.io/v1beta1
	ResourceMetricName string `json:"resourceMetricName,omitempty" yaml:"resourceMetricName,omitempty"`
}
```

this might cause some duplication in case the same request, for the same object, or specific objects and a wildcard request for the object's kind are present.

Co-authored-by: Martin Hrabovcin(@mhrabovcin) [mhrabovcin@users.noreply.github.com](mailto:mhrabovcin@users.noreply.github.com)

## Checklist

- [ ] Yes
- [x] No


go.mod
```diff
@@ -43,8 +43,8 @@ require (
 	k8s.io/api v0.27.2
 	k8s.io/apiextensions-apiserver v0.27.0
 	k8s.io/apimachinery v0.27.2
-	k8s.io/apiserver v0.27.2
-	k8s.io/cli-runtime v0.27.2
+	k8s.io/apiserver v0.27.1
```
banjoh: Please rebase your branch

banjoh:
@ahmedElqutb thanks for making this contribution. Is there an issue describing the motivation behind this collector? I'm curious to know what use cases you intent to achieve here.

ahmedElqutb:
> @ahmedElqutb thanks for making this contribution. Is there an issue describing the motivation behind this collector? I'm curious to know what use cases you intent to achieve here.

we are using a troubleshoot fork for collecting/describing customer clusters for debugging purposes and sometimes these clusters are in an airgapped environment so we don't have access to their thanos/grafana dashboards.
Thus it is very useful to have some metrics collected as part of the bundle instead of having to do it the manual way.

xavpaice:
Thanks for making this change, it's really appreciated.  I can see this being a useful collector for the project and I support adding it.

The failing test indicates that you need to run `make schemas` and commit the result - the changes to the API mean that some code needs to get regenerated.

We will need to add some docs, I see that you have added a detailed description in the PR already, please let us know if you would like to handle the docs for the collector in https://github.com/replicatedhq/troubleshoot.sh/tree/main/docs/source/collect or if we should do that separately.

ahmedElqutb:
> Thanks for making this change, it's really appreciated. I can see this being a useful collector for the project and I support adding it.
> 
> The failing test indicates that you need to run `make schemas` and commit the result - the changes to the API mean that some code needs to get regenerated.
> 
> We will need to add some docs, I see that you have added a detailed description in the PR already, please let us know if you would like to handle the docs for the collector in https://github.com/replicatedhq/troubleshoot.sh/tree/main/docs/source/collect or if we should do that separately.

That's great to hear! I'll provide the relevant docs in the next 2-3 days at most. Thanks!

ahmedElqutb:
@xavpaice here is the docs page as promised https://github.com/replicatedhq/troubleshoot.sh/pull/501

pkg/collect/k8s_metrics.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (c *CollectMetrics) IsExcluded() (bool, error) {
+	return isExcluded(c.Collector.Exclude)
+}
+
+func (c *CollectMetrics) Collect(progressChan chan<- interface{}) (CollectorResult, error) {
+	output := NewResult()
+	resultLists := make(map[string][]custom_metrics.MetricValue)
+	errorsList := make([]string, 0)
+	for _, metricRequest := range c.Collector.MetricRequests {
```
banjoh, line 48: Some logging here (using `klog.V(2).Info...`) would come in handy here. When we run `./support-bundle spec.yaml -v2` we usually print verbose output of what collectors/analysers/redactors are doing. `-v1` is verbose output of what the "framework" is doing i.e support bundle, preflight then `-v0` is for high level events.

PS: Instrumenting the library with more logs is work in progress. You can take a look at how https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/host_subnetavailable.go collector presents its progress as logs.

banjoh (dismissed):
LGTM besides that one comment on logging

pkg/collect/k8s_metrics.go
ahmedElqutb: sounds good, I added some logs to track progress of the requests and processing.

banjoh (approved):

PR #1185, opened 2023-05-29
feat(analyzer): fix clusterResource analyzer  always uses the first object found

### LINKED ISSUE #1184: `clusterResource` analyzer ignores the `name` property and always uses the first object found
**Bug Description**

Today the analyzer will always return the first result object in the list because the `name` property being provided by the end-user is ignored. This is due to a collision in the variable name used for name of the object being searched for and the actual object in the collected contents.

See - https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/kube_resource.go#L96-L103

**Expected Behavior**

The analyzer should return the unique object expected and not the first result

**Steps To Reproduce**

Using:
```
kind: Preflight
metadata:
  name: preflight
spec:
  collectors:
    - clusterResources: {}
  analyzers:
    - clusterResource:
        clusterScoped: false
        checkName: check-pvc
        kind: pvc 
        name: some-pvc
        namespace: default
        yamlPath: "spec.accessModes.[0]"
        expectedValue: ReadWriteOnce
        outcomes:
          - fail:
              when: "false"
              message: isFalse
          - pass:
              when: "true"
              message: isTrue
```

Even if `some-pvc` doesn't exist, as long as at least one PVC exists in the `default` namespace, and that PVC had an accessMode of `ReadWriteOnce` the analyzer would pass.

DexterYan:

- fix a variable name collision
- update tests to test no existed resource

Fixes: #1184

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/analyze/kube_resource_test.go
+++ b/pkg/analyze/kube_resource_test.go
@@ -10,12 +10,14 @@ import (
 
 func Test_clusterResource(t *testing.T) {
 	tests := []struct {
-		name     string
-		isError  bool
-		analyzer troubleshootv1beta2.ClusterResource
+		name           string
+		isError        bool
+		resourceExists bool
+		analyzer       troubleshootv1beta2.ClusterResource
 	}{
 		{
-			name: "namespaced resource",
+			name:           "namespaced resource",
+			resourceExists: true,
 			analyzer: troubleshootv1beta2.ClusterResource{
 				CollectorName: "Check namespaced resource",
 				Kind:          "Deployment",
@@ -24,7 +26,8 @@ func Test_clusterResource(t *testing.T) {
 			},
 		},
 		{
-			name: "check default fallthrough",
+			name:           "check default fallthrough",
+			resourceExists: true,
 			analyzer: troubleshootv1beta2.ClusterResource{
 				CollectorName: "Check namespaced resource",
 				Kind:          "Deployment",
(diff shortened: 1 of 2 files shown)


pkg/analyze/kube_resource_test.go
```diff
@@ -24,22 +26,34 @@ func Test_clusterResource(t *testing.T) {
...
+			resourceExists: true,
 			analyzer: troubleshootv1beta2.ClusterResource{
 				CollectorName: "Check namespaced resource",
 				Kind:          "Node",
 				ClusterScoped: true,
 				Name:          "repldev-marc",
 			},
 		},
+		{
+			name:           "not existed resource",
```
xavpaice:
```suggestion
			name:           "resource does not exist",
```

pkg/analyze/kube_resource_test.go
```diff
@@ -24,22 +26,34 @@ func Test_clusterResource(t *testing.T) {
...
+			resourceExists: false,
+			analyzer: troubleshootv1beta2.ClusterResource{
+				CollectorName: "Check namespaced resource",
+				Kind:          "Node",
+				ClusterScoped: true,
+				Name:          "not-existed-resource",
```
xavpaice:
```suggestion
				Name:          "resource-does-not-exist",
```

xavpaice (changes requested):
Couple of very minor suggestions, otherwise LGTM.

xavpaice (approved):

Issue #871, opened 2022-11-23
Validation of input specs before running troubleshoot operations


banjoh:
**Describe the rationale for the suggested feature.**

Input specs (often YAML) passed in to troubleshoot CLIs & library functions can often be erroneous. We need to provide a way to validate this input before running any operations. Any validation errors need to be clear and user friendly. At the moment, input validation is handled by collector/analyser/redactor implementation at a very basic level.

Yaml validation is not in itself enough. Below is an example of a case that yaml validation won't catch.

_Wrong indentations (Valid YAML, but invalid troubleshoot spec)_
```yaml
apiVersion: troubleshoot.sh/v1beta2
kind: HostPreflight
metadata:
  name: subnet-available
spec:
  collectors:
    # would output yes/no depending if there is a /22 available in 10.0.0.0/8
    - subnetAvailable:
      CIDRRangeAlloc: "10.0.0.0/8"
      desiredCIDR: 22
  analyzers:
    - subnetAvailable:
      outcomes:
        - fail:
          when: "no-subnet-available"
          message: failed to find available subnet
        - pass:
          when: "a-subnet-is-available"
          message: available /22 subnet found
```

... collector json object equivalent will be
```json
{
    "subnetAvailable": null,
    "CIDRRangeAlloc": "10.0.0.0/8",
    "desiredCIDR": 22
}
```

_Correct indentation_
```yaml
apiVersion: troubleshoot.sh/v1beta2
kind: HostPreflight
metadata:
  name: subnet-available
spec:
  collectors:
    # would output yes/no depending if there is a /22 available in 10.0.0.0/8
    - subnetAvailable:
        CIDRRangeAlloc: "10.0.0.0/0"
        desiredCIDR: 22
  analyzers:
    - subnetAvailable:
        outcomes:
          - fail:
              when: "no-subnet-available"
              message: failed to find available subnet
          - pass:
              when: "a-subnet-is-available"
              message: available /22 subnet found
```

... collector json object equivalent will be
```json
{
    "subnetAvailable": {
        "CIDRRangeAlloc": "10.0.0.0/0",
        "desiredCIDR": 22
    }
}
```

**Describe the feature**

Below are two approaches this feature could take
- Annotate structs with validation tags and ran inputs through a validation library such as [this one](https://github.com/go-playground/validator) to ensure they do not violate the required specification.
[...]

PR #1189, opened 2023-05-29
Add regex to `clusterResources` analyzer

### LINKED ISSUE #1180: Typo in PersistentVolumeClaim resource name in the `clusterResource` analyzer
**Bug Description**

In the `clusterResource` analyzer, when using `kind: PersistentVolumeClaim` it fails with:

```
Analyzer Failed

failed to find resource: failed to find resource
```

This is due to a typo in the resource names for the `clusterResource` analyzer here - https://github.com/replicatedhq/troubleshoot/blob/main/pkg/analyze/kube_resource.go#L25. The key name is incorrect which is what causes the issue. 

**Expected Behavior**

Using `kind: PersistentVolumeClaim` should not result in an error

**Steps To Reproduce**

```
    - clusterResource:
        checkName: Check PVC is RWX
        kind: PersistentVolumeClaim
        name: pvc-name
        namespace: some-namespace
        yamlPath: "spec.accessModes.[0]"
        expectedValue: ReadWriteMany
        outcomes:
          - fail:
              when: "false"
              message: The user provided PVC does not support ReadWriteMany (RWX) access mode.
          - pass:
              when: "true"
              message: The user provided PVC supports ReadWriteMany (RWX) access mode.
```

**Additional Context**

v0.64.0

- Operating system
- Operating system version

diamonwiggins:

* Similiar to the `textAnalyze` analyzer, adds support to evaluate the value at the specified `yamlPath` with `regex` or `regexGroups`
* fixes a typo in the shorthand name used for searching for cluster resource PVC objects #1180 

## Checklist

- [ ] Yes
- [ ] No


pkg/analyze/kube_resource.go
```diff
@@ -108,15 +109,28 @@ func FindResource(kind string, clusterScoped bool, namespace string, name string
...
+			IconURI: "https://troubleshoot.sh/images/analyzer-icons/text-analyze.svg",
+			IsFail:  true,
+			Message: "resource does not exist",
+		}, nil
 	}
 
 	actual, err := iutils.GetAtPath(selected, analyzer.YamlPath)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get object at path: %s", analyzer.YamlPath)
+		klog.Errorf("failed to get object at path: %s: %v", analyzer.YamlPath, err)
```
banjoh: Since this error catches an error with the jsonpath, wouldn't be better to return an error since its a syntax error with the spec?

pkg/analyze/kube_resource.go
```diff
@@ -125,15 +139,58 @@ func (a *AnalyzeClusterResource) analyzeResource(analyzer *troubleshootv1beta2.C
 		return nil, errors.Wrap(err, "failed to parse expected value as yaml doc")
 	}
 
-	result := &AnalyzeResult{
+	actualYAML, err := yaml.Marshal(actual)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal actual value")
+	}
+
+	if analyzer.ExpectedValue != "" {
```
banjoh, line 147: I think mentioning the order precedence of these parameters in the docs would come in handy.

pkg/analyze/kube_resource.go
diamonwiggins: how about now?

pkg/analyze/kube_resource.go
```diff
@@ -108,15 +109,28 @@ func FindResource(kind string, clusterScoped bool, namespace string, name string
...
+			IconURI: "https://troubleshoot.sh/images/analyzer-icons/text-analyze.svg",
+			IsFail:  true,
+			Message: "resource does not exist",
+		}, nil
 	}
 
 	actual, err := iutils.GetAtPath(selected, analyzer.YamlPath)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get object at path: %s", analyzer.YamlPath)
+		klog.Errorf("invalid yaml path: %s for kind: %s: %v", analyzer.YamlPath, analyzer.Kind, err)
```
banjoh, line 126: The message is clearer now.

I was coming from the POV of the cli, where if the spec is wrong i.e invalid regex, jsonpath etc, the cli would return an error code as per https://troubleshoot.sh/docs/preflight/exit-codes/.

I'm not so fussy but later on when we start [validating specs](https://github.com/replicatedhq/troubleshoot/issues/871), this bit will change.

banjoh (approved):

Issue #1186, opened 2023-05-29
e2e tests fail first time round with an error from runPod


xavpaice:
**Bug Description**

On new changes running tests, the e2e test fails with the error:

```
failed to run collector: run/static-hi: failed to run pod: failed to create pod: object is being deleted: pods "static-hi" already exists. 
```

**Expected Behavior**

The pod should be deleted after success, not just after failure.

**Steps To Reproduce**

Make a small change, watch the tests.

However, you could likely also produce a reproducer with a support bundle spec that includes runPod, and watch to see if the pod is cleaned up after the collector finishes.

**Additional Context**

Include the following information.
- Troubleshoot version. If you built from source, note that including the version of Go you used to build with.
- Operating system
- Operating system version
- Other details that might be helpful in diagnosing the problem

Most recent change in that part of the code: https://github.com/replicatedhq/troubleshoot/pull/1172 (might be irrelevant)

PR #1202, opened 2023-06-01
feat: add loader APIs to load specs from raw troubleshoot spec

### LINKED ISSUE #1014: Misleading/redundant usage of AddToScheme
**Bug Description**

Looking at those files

* cmd/collect/cli/run.go
* cmd/troubleshoot/cli/run.go
* pkg/collect/load.go
* pkg/supportbundle/load.go

they use this code pattern:

```go
import (
	"github.com/replicatedhq/troubleshoot/pkg/client/troubleshootclientset/scheme"
	troubleshootclientsetscheme "github.com/replicatedhq/troubleshoot/pkg/client/troubleshootclientset/scheme"
)

func someFunc(...) {
	// ...
	troubleshootclientsetscheme.AddToScheme(scheme.Scheme)
	// ...
}
```

This seems to be:

* misleading, since one would expect `scheme` to be an import of `k8s.io/client-go/kubernetes/scheme`.
* redundant, as the code as is is actually also part of the init function of the troubleshootclientset's scheme.

**Expected Behavior**

Depending on the desired behavior, either one of the following:

1. remove those lines, as they are redundant
2. fix the import, so that the troubleshoot scheme gets added to client-go: `import "k8s.io/client-go/kubernetes/scheme"`
3. fix the import usage, so that the lines look like `troubleshootclientsetscheme.AddToScheme(troubleshootclientsetscheme.Scheme)`

I think removal should be fine, since as it's right now, they're not doing anything. But maybe there are bugs that may be fixed by option 2.

**Additional context**

AFAICT, the `prefilght` command actually uses client-go's import, so maybe it's required there.

banjoh:

The change introduces a `loader` package that will contain loader public APIs. The aim of these APIs will be to, given any source of troubleshoot specs, fetch the specs and parse out all troubleshoot objects that can be extracted. This specific change contains one API to load specs.

`LoadSpecs(ctx context.Context, opt LoadOptions) (*TroubleshootKinds, error)`

Sources to load spec will come from `LoadOptions` parameter

At this point the API will expect a prepared array (slice) of strings containing valid yaml documents. The `LoadSpec` function will parse these documents and extract any troubleshoot specs. This includes specs inside `Secrets` or `ConfigMaps`.

**Follow-up work**
* This PR is a building block of the ongoing work with https://github.com/replicatedhq/troubleshoot/pull/1144. This piece focuses on standardising how troubleshoot specs are collected from various sources and parsed into GVK objects. Other changes that will be considered are
  * Loading specs from file paths and URLs
  * Loading specs from k8s clusters. This is the current ["discover cluster specs"](https://troubleshoot.sh/docs/support-bundle/discover-cluster-specs/) implementation we have in troubleshoot

* Preflight CLI is now using `LoadSpecs()`. Other parts of the project will be moved to consume these loader API

**Additional changes**
* Cleanup of `troubleshootclientsetscheme.AddToScheme(scheme.Scheme)` redundant calls is part of addressing #1014 issue.

Fixes: #1014

## Checklist

- [ ] Yes
- [x] No


banjoh:
I'll need to update https://github.com/replicatedhq/troubleshoot/pull/1144 to reflect this change

diamonwiggins (commented):
this lgtm so far, but I think it may be worth adding a preflight example that's multidoc to the rendered helm templates. I think traditionally our multidoc support has been a single SupportBundle or Preflight spec is expected first, and then only redactor specs afterwards, and with this code change I think we're making that more robust, so would be good to have a couple multidoc examples in the test. 

edit. an example of what i'm referring to - https://github.com/replicatedhq/troubleshoot/blob/main/sample-troubleshoot.yaml. For this PR it would maybe be two preflight specs in the same doc because Preflights don't run redaction. 

Multi-doc may not be relevant here in the context of helm because you'd probably create a separate key per spec, but it's technically possible I suppose.

banjoh:
> Multi-doc may not be relevant here in the context of helm because you'd probably create a separate key per spec, but it's technically possible I suppose.

I added support for splitting multidocs within `ConfigMap` and `Secret` data. This allows us to further load a spec like below. I believe this is what you meant, right?

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sb-spec
  labels:
    troubleshoot.io/kind: support-bundle
data:
  support-bundle-spec: |-
    kind: SupportBundle
    spec:
      collectors:
        - logs:
            name: all-logs
    ---
    kind: SupportBundle
    spec:
      collectors:
        - clusterResources: {}
    ---
    kind: SupportBundle
    spec:
      collectors:
        - clusterInfo: {}
```

There is an ambiguous scenario where the data of a `ConfigMap` can contain a spec like below. Note that the ` support-bundle-spec` now contains a redactor. In this implementation, the redactor will be parsed and loaded, but its a bit of a confusing use case.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: sb-spec-2-configmap
  labels:
    troubleshoot.io/kind: support-bundle
data:
  support-bundle-spec: |-
    kind: SupportBundle
    ---
    kind: Redactor
```

diamonwiggins (dismissed):
lgtm

diamonwiggins (approved):
tentatively approved. not sure if we have a flaky test which isn't related to https://github.com/replicatedhq/troubleshoot/issues/1186

banjoh:
> tentatively approved. not sure if we have a flaky test which isn't related to #1186

I think its a race condition related to the first `static-hi` pod clean up, cleaning up the service account that the 2nd `static-hi` pod needs. This is a theory, but https://github.com/kubernetes/kubernetes/issues/66689 issue has some context on pod creation having a potential race condition related to this missing service account

PR #1220, opened 2023-06-12
don't fail to collect cluster-resources if negotiating RBAC fails because of authorizer webhooks

### LINKED ISSUE #1167: clusterResources skips several parts for GKE clusters, without namespace specified
**Bug Description**

I have a GKE cluster, with the ability to run `kubectl get pods` on all namespaces (plus other permissions).

I expect to see, in a namespace that has deployments, pods, etc to see those listed.  I do not, I see a very minimal set of information.

If I specify the namespace, everything works as expected.

**Expected Behavior**

Given that my user has access to the resources required to collect from all namespaces, I expect to see information including deployments etc for all namespaces unless I specify otherwise.

**Steps To Reproduce**

Include the commands to reproduce the issue including any output. Any information that will help us to understand the problem is useful.  Feel free to paste long output into a [Github gist](https://gist.github.com) and include the link here.

* Requires a GKE cluster with multiple namespaces

Run `support-bundle` using the following minimal spec:

```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - clusterResources: {}
```

**Additional Context**

https://github.com/replicatedhq/troubleshoot/blob/6de79afc35a514cd65294403f939a814e00c065b/pkg/collect/cluster_resources.go#L1488 returns only one ResourceRules for the namespace, with the Verb "create".  The same exercise against a local k3d cluster has an extra ResourceRules entry with Verb "*" for the namespace, and Resources "*".

This causes https://github.com/replicatedhq/troubleshoot/blob/6de79afc35a514cd65294403f939a814e00c065b/pkg/collect/cluster_resources.go#L1542 to return false for every namespace, thus very little collection happens.

adamancini:

#1167 

https://github.com/kubernetes/kubernetes/issues/65744 leads some clusters using authorization webhooks to fail to report a users' authorizations correctly, so cluster-resources collector skips over some namespaces.

## Checklist

- [ ] Yes
- [x] No


xavpaice:
the approach here makes sense to me, not entirely sure why the watch script is in this PR?

If you rebase this against main, there's been a fix to the issue that was causing test failures.

adamancini:
@xavpaice I forgot to remove it when I cherry-picked to make the PR - I'll probably have to re-open

PR #1196, opened 2023-05-31
fix(collector): solving pod stuck in terminating with CNI issue

DexterYan:
## Description, Motivation and Context

- this pr is targeted on two collectors `runPod` and `copyFromHost` 
- Make sure the run pod and daemonset have been deleted before the troubleshoot.sh finished running
- Poll every second to check if the Pod has been deleted. The maximum grace period is 1 Minute.
- If it reached 1 Minute,  all the pods created by those two collectors will be force deleted
- add klog.v(2) for reporting process

Fixes: 
#1188
#1186

Previous PR:
https://github.com/replicatedhq/troubleshoot/pull/1172
https://github.com/replicatedhq/troubleshoot/pull/1194

## Checklist

- [ ] New and existing tests pass locally with introduced changes.
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] The commit message(s) are informative and highlight any breaking changes
- [ ] Any documentation required has been added/updated. For changes to https://troubleshoot.sh/ create a PR [here](https://github.com/replicatedhq/troubleshoot.sh/pulls)

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #1222, opened 2023-06-14
feat(collector): checking existing service account before create running pod

### LINKED ISSUE #1223: preflight-e2e-test fail due to default service account missing in default namespace
**Bug Description**

When running run-pod collector without a specified namespace, the pod will use `default `service account `default `namespace. However, in action-k3s, we don't have `default` service account in `default` namespace. The preflight e2e test will fail.

**Expected Behavior**

Run make preflight-e2e-test of github action task should pass without following error
```
error - failed to run collector: run-pod/static-hi: failed to run pod: failed to create pod: pods "static-hi" is forbidden: error looking up service account default/default: serviceaccount "default" not found
```

**Steps To Reproduce**

failed job: https://github.com/replicatedhq/troubleshoot/actions/runs/5205334804/jobs/9390719637

DexterYan:
- check the service account before creating the running pod
- if the service account didn't exist, create the service account first

Related to PR: https://github.com/replicatedhq/troubleshoot/pull/1196#issuecomment-1579883857
Fixes: #1223 
## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/util.go
+++ b/pkg/collect/util.go
@@ -227,3 +227,12 @@ func getTLSParamsFromSecret(ctx context.Context, client kubernetes.Interface, se
 
 	return caCert, clientCert, clientKey, nil
 }
+
+func checkForExistingServiceAccount(client kubernetes.Interface, namespace string, serviceAccountName string) error {
+	_, err := client.CoreV1().ServiceAccounts(namespace).Get(context.Background(), serviceAccountName, metav1.GetOptions{})
+
+	if err != nil {
+		return errors.Wrapf(err, "Failed to get service account %s", serviceAccountName)
+	}
+	return nil
+}
--- a/pkg/collect/run.go
+++ b/pkg/collect/run.go
@@ -43,6 +43,10 @@ func (c *CollectRun) Collect(progressChan chan<- interface{}) (CollectorResult, 
 		serviceAccountName = c.Collector.ServiceAccountName
 	}
 
+	if err := checkForExistingServiceAccount(c.Client, namespace, serviceAccountName); err != nil {
+		return nil, err
+	}
+
 	runPodSpec := &troubleshootv1beta2.RunPod{
 		CollectorMeta: troubleshootv1beta2.CollectorMeta{
 			CollectorName: c.Collector.CollectorName,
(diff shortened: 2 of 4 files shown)


xavpaice:
I have one concern, in that we should already have the default serviceaccount there, and if we do not, do we really want to create it as part of a support-bundle run?

pkg/collect/util.go
```diff
@@ -227,3 +228,21 @@ func getTLSParamsFromSecret(ctx context.Context, client kubernetes.Interface, se
...
 	return caCert, clientCert, clientKey, nil
 }
+
+func checkForExistingServiceAccount(client kubernetes.Interface, namespace string, serviceAccountName string) error {
+	_, err := client.CoreV1().ServiceAccounts(namespace).Get(context.Background(), serviceAccountName, metav1.GetOptions{})
+
+	if err != nil {
+		klog.V(2).Infof("Failed to get service account %s, creating it", serviceAccountName)
+		// Service account doesn't exist, create it
+		_, err = client.CoreV1().ServiceAccounts(namespace).Create(context.Background(), &corev1.ServiceAccount{
```
xavpaice: this will help immensely in tests with k3s, but we want to be sure that it's the behavior we want with production sites - the end user might not expect us to be creating serviceAccount resources when running Troubleshoot, and everything else gets deleted at the end.  I'd not want to go deleting the default sa in this case.

DexterYan:
> I have one concern, in that we should already have the default serviceaccount there, and if we do not, do we really want to create it as part of a support-bundle run?

In most standard k8s cluster, we should have default service account in default namespace, however, as you pointed out, I think we should give a warning not fail or create the default service account in default namespace. If the user is intended to delete default account. We should not create it again.Otherwise, we need to cleanup.

pkg/collect/util.go
DexterYan: I think this one only happens when user has specified its running service account and namespace in spec yaml,  then we just check if the service account existed. otherwise just warning it.

DexterYan: I have changed the logic that we just get if the service account existed or not. Not creating it

xavpaice (approved):
LGTM

PR #1240, opened 2023-06-23
feat: Dry run flag to print preflight specs to std out

### LINKED ISSUE #1024: Expose rendered spec after merging via an API and the cli

Troubleshoot has the ability to [merge multiple specs](https://troubleshoot.sh/docs/support-bundle/collecting/#collect-a-support-bundle-using-multiple-specs) and run all collectors/analysers/redactors in the resulting spec. These specs can be pulled from (as of today) the following sources
- http(s) urls
- oci uris
- files on disk (will include [multidocs as well](https://github.com/replicatedhq/troubleshoot/issues/1023))
- [embedded uris](https://troubleshoot.sh/docs/support-bundle/supportbundle/#uri)

Merged collectors are then [deduplicated](https://github.com/replicatedhq/troubleshoot/pull/972) and some default specs (cluster-resources, cluster-info, default redactors) added as well. It has now become hard to get a full picture of how the full support bundle spec would look like before executing it.

**Describe the feature**

- Define a public API that would be used to build a full spec from merged specs
- Using this public API, add a new subcommand/cli option to render the entire spec. Some examples of how the CLI interface would look like are
```sh
support-bundle render-spec <uris...> # same for preflight
support-bundle <uris...> --dry-run # same for preflight
```

**Tasks**
- [x] Add flag to `preflight` binary
- [ ] Add flag to `support-bundle` binary

banjoh:

Introduce `preflight --dry-run` flag used to print specs to stdout. I went for `--dry-run` flag instead of adding a new subcommand (`dump`, `print`, `view` etc) cause it felt a bit more intuitive. It also makes sense since when we pass in multiple specs, `--dry-run` would print out what would be run in spec format.

**Notes for reviewer**
- Introduce `ToYaml()` method that marshals `TroubleshootKinds` to yaml documents
- Move cmd/util to cmd/internal/util. Its private implementation
- Remove `PreflightSpecs` in favour of `TroubleshootKinds`

Partially addresses #1024. We still need to add `support-bundle --dry-run` flag

## Checklist

- [ ] Yes
- [x] No


banjoh:
Here is an example parsing specs from a helm chart (stdin), a url and a local file, then printing to stdout.

```sh
[evans] $ helm template examples/sdk/helm-template/mychart-0.1.0.tgz --set preflight=true | ./bin/preflight --dry-run https://raw.githubusercontent.com/replicatedhq/troubleshoot/main/examples/preflight/host/cpu.yaml -
kind: HostPreflight
metadata:
  creationTimestamp: null
  name: cpu
spec:
  analyzers:
  - cpu:
      outcomes:
      - fail:
          message: At least 4 physical CPU cores are required
          when: physical < 4
      - fail:
          message: At least 8 CPU cores are required
          when: logical < 8
      - warn:
          message: At least 16 CPU cores preferred
          when: count < 16
      - pass:
          message: This server has sufficient CPU cores
  collectors:
  - cpu: {}
status: {}
---
kind: Preflight
metadata:
  creationTimestamp: null
  name: sample
spec:
  analyzers:
  - postgres:
      checkName: Must be postgres 10.x or later
      collectorName: pg
      outcomes:
      - fail:
          message: Cannot connect to postgres server
          when: connected == false
      - fail:
          message: The postgres server must be at least version 10
          when: version < 10.x
      - pass:
          message: The postgres connection checks out
  collectors:
  - postgres:
      collectorName: pg
      uri: postgresql://user:password@hostname:5432/defaultdb?sslmode=require
status: {}
```

cmd/preflight/cli/root.go
```diff
@@ -64,10 +64,13 @@ that a cluster meets the requirements to run an application.`,
 
 	cobra.OnInitialize(initConfig)
 
-	cmd.AddCommand(VersionCmd())
+	cmd.AddCommand(util.VersionCmd())
 	cmd.AddCommand(OciFetchCmd())
 	preflight.AddFlags(cmd.PersistentFlags())
 
+	// Dry run flag should not persist between across subcommands. Adding here to avoid that
+	cmd.Flags().Bool("dry-run", false, "print the preflight spec without running preflight checks")
```
xavpaice, line 73: isn't this added below in pkg/preflight/flags.go?  Not sure I understand the comment.

pkg/preflight/run.go
```diff
@@ -41,14 +41,21 @@ func RunPreflights(interactive bool, output string, format string, args []string
...
+			return types.NewExitCodeError(constants.EXIT_CODE_CATCH_ALL, errors.Wrap(err, "failed to convert specs to yaml"))
+		}
+		fmt.Printf("%s", out)
+		return nil
+	}
 
+	warning := validatePreflight(specs)
 	if warning != nil {
 		fmt.Println(warning.Warning())
 		return nil
```
xavpaice: I kinda want this warning in the dry-run also

pkg/preflight/validate_specs.go
```diff
@@ -4,45 +4,37 @@ import (
...
-	if specs.HostPreflightSpec != nil {
-		warning := validateHostPreflightSpecItems(specs.HostPreflightSpec.Spec.Collectors, specs.HostPreflightSpec.Spec.Analyzers)
+	for _, spec := range kinds.HostPreflightsV1Beta2 {
+		warning := validateHostPreflightSpecItems(spec.Spec.Collectors, spec.Spec.Analyzers)
 		if warning != nil {
 			return warning
 		}
 	}
 
-	if specs.UploadResultSpecs != nil {
```
xavpaice, line 32: can we avoid losing the validation of this type?

xavpaice (commented):
like the approach, couple of minor questions.

cmd/preflight/cli/root.go
banjoh: > isn't this added below in pkg/preflight/flags.go? Not sure I understand the comment.

All flags in `pkg/preflight/flags.go` are also available to all subcommands e.g `preflight completions`.

```sh
[evans] $ preflight completion -h
Generate the autocompletion script for preflight for the specified shell.
See each sub-command's help for details on how to use the generated script.

Usage:
  preflight completion [command]

Available Commands:
  bash        Generate the autocompletion script for bash
  fish        Generate the autocompletion script for fish
  powershell  Generate the autocompletion script for powershell
  zsh         Generate the autocompletion script for zsh

Flags:
  -h, --help   help for completion

Global Flags:
      --collect-without-permissions   always run preflight checks even if some require permissions that preflight does not have (default true)
      --collector-image string        the full name of the collector image to use
      --collector-pullpolicy string   the pull policy of the collector image
      --cpuprofile string             File path to write cpu profiling data
      --debug                         enable debug logging
      --format string                 output format, one of human, json, yaml. only used when interactive is set to false (default "human")
      --interactive                   interactive preflights (default true)
      --memprofile string             File path to write memory profiling data
  -o, --output string                 specify the output file path for the preflight checks
      --selector string               selector (label query) to filter remote collection nodes on.
      --since string                  force pod logs collectors to return logs newer than a relative duration like 5s, 2m, or 3h.
      --since-time string             force pod logs collectors to return logs after a specific date (RFC3339)

Use "preflight completion [command] --help" for more information about a command.
```

Running `preflight completions --dry-run` or `preflight completions --since 3h` is redundant.

banjoh: Alternatively, we could change [this line](https://github.com/replicatedhq/troubleshoot/blob/d0d6f00c0f0a08d579fc4af498a007f049525d5a/cmd/preflight/cli/root.go#L69) to `preflight.AddFlags(cmd.Flags())`

Personally I prefer each subcommand to explicitly list all the flags it needs to expose like the `support-bundle binary` does. Thats for another PR though. I thought refactoring this part would have been scope creep.

pkg/preflight/validate_specs.go
banjoh: > can we avoid losing the validation of this type?

TL;DR - We aren't losing the validation. It takes place [here](https://github.com/banjoh/troubleshoot/blob/7def396527932d1bf377c137c20769216c2849cd/pkg/preflight/validate_specs.go#L19-L24) where that list of `troubleshootv1beta2.Preflight` objects will contain both `PreflightSpec` and `UploadResultSpecs` preflight objects which were in the original `PreflightSpecs` struct

**Longer description**
Changes in this PR have gotten rid of the [PreflightSpecs struct](https://github.com/replicatedhq/troubleshoot/blob/d0d6f00c0f0a08d579fc4af498a007f049525d5a/pkg/preflight/read_specs.go#L14) in favour of using `TroubleshootKinds`.

Here is a comparison of the structs

```go
type PreflightSpecs struct {
	PreflightSpec     *troubleshootv1beta2.Preflight
	HostPreflightSpec *troubleshootv1beta2.HostPreflight
	UploadResultSpecs []*troubleshootv1beta2.Preflight
}
```

```go
type TroubleshootKinds struct {
	// redacted ...
	HostPreflightsV1Beta2   []troubleshootv1beta2.HostPreflight
	PreflightsV1Beta2       []troubleshootv1beta2.Preflight
}
```

As you can see both `PreflightSpec` and `UploadResultSpecs` store `*troubleshootv1beta2.Preflight`. I changed the logic avoid transforming `TroubleshootKinds` to `PreflightSpecs` in favour of just using `TroubleshootKinds` in an effort of having a common type that houses troubleshoot spec objects. Now objects there were in `PreflightSpecs.PreflightSpec` and `PreflightSpecs.UploadResultSpecs` will be in `TroubleshootKinds.PreflightsV1Beta2`. The difference between objects that were in `PreflightSpec` and `UploadResultSpecs` is the presence/absence of `Spec.UploadResultsTo` which now handled [here](https://github.com/banjoh/troubleshoot/blob/7def396527932d1bf377c137c20769216c2849cd/pkg/preflight/run.go#L89C11-L89C31)

Changes can be found in
- Validation code
- Code to read spec inputs

xavpaice (approved):

PR #1252, opened 2023-06-28
feat(examples): loading troubleshoot specs from helm chart example

banjoh:

An example application showcasing how to load troubleshoot specs from a helm chart. The example is also executed by the CI to ensure that the examples are kept in sync with the public API. Running the example also acts as an additional a way to ensure that the APIs retain compatibility, though there will be tests specifically for this

## Checklist

- [ ] Yes
- [x] No


examples/sdk/helm-template/go.mod
```diff
@@ -0,0 +1,73 @@
+module helm-template
+
+go 1.20
```
DexterYan, line 3: looks quite good, just one thing that our troubleshoot go.mod is using go 1.19. Should we keep this sync with it https://github.com/replicatedhq/troubleshoot/blob/f0efbf658a02a36cbb0db3e706a09c808e7f54c8/go.mod#L3C1-L3C8

xavpaice: given that this is an example application, I'm comfortable with it running a current Go rather than limiting it to the project's version (which we should update anyway).

xavpaice (approved):
while at first glance this looks like it's missing tests, it is, in fact, a test addition in itself.  Happy for this to be merged.

Issue #1225, opened 2023-06-15
Smoke test troubleshoot windows binaries


banjoh:
**Describe the rationale for the suggested feature.**

We ship troubleshoot binaries for Windows OS but most of out testing is done on *nix based systems (Mac, Linux). We need to run some automated tests on Windows environments to have the quality control gating before making releases

**Describe the feature**

TBD

**Additional context**

Microsoft ships [Windows docker images](https://hub.docker.com/_/microsoft-windows-base-os-images) which we can use. We'd need to see whether they can be used without paid licenses.

Issue #962, opened 2023-01-13
Rework sbctl tests to avoid conflicts with names in Troubleshoot repo


xavpaice:
Per the design in #893 

We will need to have a separate directory for sbctl (ginkgo BDD) tests. The test harness used is ginkgo cli instead of go test and there is no straight forward way of separating these tests. Also, from a directory structure point of view, such integration tests usually reside in modules in a root level tests directory. Changes to the test code need to be minimal, ideally no change, cause they will be used to ensure any refactored code does not break. We could now easily introduce tests which ensure that changes to collectors don't break APIs in sbctl.

Issue #891, opened 2022-12-06
Increase e2e test coverage to allow better testing of PRs


banjoh:
**Describe the rationale for the suggested feature.**

PRs raised, especially by dependabot, will need to be tested better to increase confidence before approval & merging.

**Describe the feature**

- Add more collectors & analysers in both [support-bundle](https://github.com/replicatedhq/troubleshoot/blob/main/examples/support-bundle/e2e.yaml) and [preflight](https://github.com/replicatedhq/troubleshoot/blob/main/examples/preflight/e2e.yaml) e2e specs
- Run different versions of k8s in a [matrix](https://github.com/replicatedhq/action-k3s#example-usage) to allow executing troubleshoot applications in different k8s versions
- Run tests against different OS distros and version (kernels) that we support so that we catch regressions such as [this one](https://github.com/replicatedhq/troubleshoot/pull/1121)
- Run e2e tests against different k8s distros. Openshift is for example quite restrictive and it would be good to cover some basic use cases.
- Launch, where possible, ancillary systems such as databases, a web server, storage providers (ceph, longhorn) etc. They can be run within the [test k3s cluster](https://github.com/replicatedhq/troubleshoot/blob/main/.github/workflows/build-test-deploy.yaml#L111) or as docker-compose containers depending on which is the most appropriate approach. Host collectors will require special attention.
- Rethink how we assert test results. [Using jq](https://github.com/replicatedhq/troubleshoot/blob/main/test/validate-support-bundle-e2e.sh#L29) might be sufficient, but is there a better way?
- Add tests that target RBAC scenarios e.g one that would catch https://github.com/replicatedhq/troubleshoot/issues/1079 bug.

**Describe alternatives you've considered**

Leave things as they are and trust the review process solely.

**Additional context**

There is higher confidence of the correctness of PRs authored by contributors, but dependabot PRs may introduce bugs that are hard to spot.

PR #1265, opened 2023-07-12
feat(e2e): add e2e go test for cluster resource

DexterYan:

- use e2e-framework to package our current bash e2e tests into go test
- change github action to support e2e test
- add test for cluster resource collector and pod status

## Checklist

- [ ] Yes
- [ ] No


CpuID:
Interesting 👍 

Not sure if I'd want to completely remove the standalone binary e2e test just yet, but as an additional test there's some valid use cases for sure.

banjoh:
I like this framework and how it approaches end-to-end testing especially cause
* It exposes `testing.T` tests and hence uses `go test` harness.
* It has traditional/xUnit type `SetUp/TearDown` functions which help see the environment the tests run within the same code base

Out of curiosity, have you considered other alternatives? 

As we look to introducing a new testing framework, we need to evaluate the following backlog issues to see whether we address their goals.

* https://github.com/replicatedhq/troubleshoot/issues/962
* https://github.com/replicatedhq/troubleshoot/issues/1225
* https://github.com/replicatedhq/troubleshoot/issues/891
* ...

IMO, this should lead to an ADR given that it might lead to using different tools/frameworks and different approach to testing.

DexterYan:
@banjoh, I checked other k8s test framework, seems not good e2e test like this one.

Currently, it worked in GitHub [action](https://github.com/replicatedhq/troubleshoot/actions/runs/5946802820/job/16127956874?pr=1265). So it should be suitable for Windows testing.

CpuID (dismissed):
this looks pretty good to me 👍 adds more tests than we had before, and they pass :)

DexterYan:
Thank you @CpuID. I have added one more case to check the pending pod status. 

In go e2e test, we can directly import **github.com/replicatedhq/troubleshoot/pkg/convert** as convert.Result to read our analysis.json. It seems more efficient then current bash way.

CpuID (approved):

Issue #1157, opened 2023-05-15
Cluster resources collector ignores some resource types


tstephen-relx:
**Describe the rationale for the suggested feature.**

[Cluster resources collector](https://troubleshoot.sh/docs/collect/cluster-resources/) does not cover:
- `daemonsets`
- `leases`
- `service accounts`

**Describe the feature**

Enumerate each of these just as is done for other resource types. All three have namespaces so should be exported to a folder with each file containing just one namspace's contents.

**Describe alternatives you've considered**

We are forced to fall back to kubectl commands because of these missing items.

**Additional context**

N/A

xavpaice:
Related: this could use sbctl support too (TODO write an issue for that)

Issue #1217, opened 2023-06-12
Collect additional k8s resources using the cluster resources collector


banjoh:
**Describe the rationale for the suggested feature.**

Add implementation to collect following additional resources from a cluster

**Describe the feature**

Have the `clusterResources` collector collect
* Service accounts - stored in `serviceaccounts`
* Secrets - stored in `secrets`
* ConfigMaps - stored in `configmaps`
* Volume attachments - stored in `volumeattachments`

**Additional context**

Although `secrets` objects are highly sensitive to avoid collecting, they are crucial when it comes to piecing together an environment when troubleshooting a field problem. To mitigate accidental leakage of confidential information, I propose redacting secrets `data` and `stringData` fields by default and have an explicit field to not redact such values for cases where such information needs to be retained. Perhaps we should introducing a field like `retailSensitiveData` or something similar in the [clusterResources collector](https://troubleshoot.sh/docs/collect/cluster-resources/#parameters)

xavpaice:
Secrets: https://troubleshoot.sh/docs/collect/secret/ (note: includeValue defaults to false, i.e. don't collect the data)
ConfigMap: https://troubleshoot.sh/docs/collect/configmap/

How does this differ from the above?  Is it the location of the data?

Regards volumeattachments and serviceaccounts, yup that should be easy enough to add and worthwhile.

banjoh:
> How does this differ from the above? Is it the location of the data?

Yer, its the location of the data. Since these are k8s resources, storing them there as native objects like all other resources  collected by the `clusterResources` collector makes it possible to consume them with tools like `sbctl`.

I should have made that explicit in the original description.

xavpaice:
see #1157 - work in progress to add serviceaccounts

DexterYan:
I have added volumeAttachments and configmaps. Secrets will not be collected as default.

PR #1266, opened 2023-07-13
feat(cluster_resources): add leases and service account

DexterYan:

- add leases and service accounts for [Cluster resources collector](https://troubleshoot.sh/docs/collect/cluster-resources/)
- add configmap and volume attachments

Fixes: 
#1157
#1217

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/cluster_resources_test.go
+++ b/pkg/collect/cluster_resources_test.go
@@ -2,6 +2,7 @@ package collect
 
 import (
 	"context"
+	"encoding/json"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


xavpaice:
#1157 also requests daemonsets - is that planned (or already included)?

DexterYan:
> #1157 also requests daemonsets - is that planned (or already included)?

daemonsets has been included before this PR.  https://troubleshoot.sh/docs/collect/cluster-resources/#cluster-resourcesdaemonsetsnamespacenamejson

CpuID:
@DexterYan @xavpaice thoughts on tests around this one...?

looks like we don't have a whole lot of tests around it now, there's some around namespaces but not around specific resource types...

might be a bit of a can of worms though :) dedicated task in itself to add test coverage around cluster resources as a whole?

CpuID:
also it looks like https://troubleshoot.sh/docs/collect/cluster-resources/ lists the various resource types (by filename that the collector writes out), docs PR to append to that list?

interesting that daemonset's were already on the docs list... but this PR adds support for them? 🤷

xavpaice:
In the spirit of campsite rules, I'd like to see if we can add unit tests for the individual collectors here (e.g. `func serviceAccounts()`), we would however need to mock out the k8s client to do that.

I'd also like to see how far we can go with the e2e test, making, say, a serviceAccount and then confirming that the file appears in the support bundle.

DexterYan:
I have added e2e tests for checking those resources have been collected in our support bundle. For the unit test, I have added leases, service accounts, configmaps and volume attachments. I think I will create another PR to add more tests for existing resources

pkg/collect/cluster_resources_test.go
```diff
@@ -1,14 +1,200 @@
...
+			client := testclient.NewSimpleClientset()
+			ctx := context.Background()
+			err := createConfigMaps(client, tt.configMapName, tt.namespaces)
+			require.NoError(t, err)
+
+			configMaps, _ := configMaps(ctx, client, tt.namespaces)
+			assert.Equal(t, len(tt.namespaces), len(configMaps))
+
+			for _, ns := range tt.namespaces {
+				assert.NotEmpty(t, configMaps[ns+".json"])
```
CpuID, line 53: do we want to check that the config map name exists in the JSON...? at least checking 1 element would give confidence its valid JSON

pkg/collect/cluster_resources_test.go
```diff
@@ -1,14 +1,200 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			client := testclient.NewSimpleClientset()
+			ctx := context.Background()
+			err := createTestVolumeAttachments(client, tt.leaseName)
+			require.NoError(t, err)
+
+			volumeAttachments, _ := volumeAttachments(ctx, client)
+			assert.NotEmpty(t, volumeAttachments)
```
CpuID, line 106: do we want to check that the volume attachment name exists in the JSON...? at least checking 1 element would give confidence its valid JSON

pkg/collect/cluster_resources_test.go
```diff
@@ -1,14 +1,200 @@
...
+			client := testclient.NewSimpleClientset()
+			ctx := context.Background()
+			err := createTestLeases(client, tt.leaseName, tt.namespaces)
+			require.NoError(t, err)
+
+			leases, _ := leases(ctx, client, tt.namespaces)
+			assert.Equal(t, len(tt.namespaces), len(leases))
+
+			for _, ns := range tt.namespaces {
+				assert.NotEmpty(t, leases[ns+".json"])
```
CpuID, line 166: do we want to check that the lease name exists in the JSON...? at least checking 1 element would give confidence its valid JSON

pkg/collect/cluster_resources_test.go
```diff
@@ -1,14 +1,200 @@
...
+			client := testclient.NewSimpleClientset()
+			ctx := context.Background()
+			err := createTestServiceAccounts(client, tt.serviceAccountName, tt.namespaces)
+			require.NoError(t, err)
+
+			servicesAccounts, _ := serviceAccounts(ctx, client, tt.namespaces)
+			assert.Equal(t, len(tt.namespaces), len(servicesAccounts))
+
+			for _, ns := range tt.namespaces {
+				assert.NotEmpty(t, servicesAccounts[ns+".json"])
```
CpuID, line 229: do we want to check that the service account name exists in the JSON...? at least checking 1 element would give confidence its valid JSON

CpuID (changes requested):
looking very close. couple of comments for minor changes

also can we get a docs PR together? thanks :)

pkg/collect/cluster_resources_test.go
DexterYan: Thank you! I have added json content check to check those elements have been written,

DexterYan:
> looking very close. couple of comments for minor changes
> 
> also can we get a docs PR together? thanks :)

Documents PR is ready https://github.com/replicatedhq/troubleshoot.sh/pull/520

CpuID (approved):

PR #1279, opened 2023-07-20
feat: Add Strict flag to LoadSpecs API

### LINKED ISSUE #1272: Introduce a strict mode flag on LoadSpecs API

There are some use cases where we would like to provide spec input to the [LoadSpecs API](https://github.com/replicatedhq/troubleshoot/blob/73201c03041fbc5cc8b244ba59a654f8f6d24d5b/pkg/loader/loader.go#L49) where some of the YAML documents may not be valid e.g may still have template functions, can have non YAML docs etc. In such instances, it would be good to give users the flexibility of having the API ignore errors and get results of the specs that were successfully parsed, and ignore the anything else.

**Describe the feature**

Introduce a `Strict` flag which can be used to toggle between `true` (raising errors if a document is invalid - current mode) and `false` (ignoring invalid documents, perhaps logging a warning).

```go
type LoadOptions struct {
	RawSpecs []string
	RawSpec  string
	Strict bool
}
```

We may also take an approach where we collect all errors, and continue parsing inputs, and return them to the caller. It would be the caller's responsibility to decide what to do with the errors. The only drawback here is that other errors may have occurred which now get mixed up here. This would make it hard to pick what errors to fail on and which ones not to.

banjoh:

Strict flag which can be used to toggle between true (raising errors if a document is invalid)
and false (ignoring invalid documents, perhaps logging a warning).

Fixes: #1272

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/internal/traces/exporter.go
+++ b/internal/traces/exporter.go
@@ -22,7 +22,7 @@ var (
 	printer  = message.NewPrinter(language.English)
 )
 
-const legend = "Suceeded (S), eXcluded (X), Failed (F)\n"
+const legend = "Succeeded (S), eXcluded (X), Failed (F)\n"
 
 // FUTURE WORK: This exporter should only be used by troubleshoot CLIs
 // until the following issue is addressed:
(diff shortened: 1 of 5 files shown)


divolgin (dismissed):

adamancini (dismissed):
lgtm

divolgin (dismissed):

divolgin (approved):

PR #1281, opened 2023-07-24
Fix(support-bundle): add pod force deletion and check failedmount event for copy from host daemonset pod

### LINKED ISSUE #1269: Performance: support-bundle takes > 3 mins
**Bug Description**

When running tests in KOTS, the `support-bundle` CLI test takes > 3 mins to run, causing timeouts.  While increasing the timeout is a workaround, 3 mins is too long to expect Troubleshoot to run.

**Expected Behavior**

`support-bundle` returns in < 3 mins

**Steps To Reproduce**

kURL cluster: https://kurl.sh/8cb4d5d
Command: `kubectl support-bundle --load-cluster-specs --redactors=configmap/default/kotsadm-redact-spec/redact-spec,configmap/default/kotsadm-qakotsregression-redact-spec/redact-spec --interactive=false --allow-insecure-connections`

Test: [https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/master/test/uqPj73W2fTCb[…]-id=za47mjQ58SYjqckP&path=iu3mBMtET269b9Oc%3A7IGXsuKvhZ1myT35](https://app.testim.io/#/project/wpYAooUimFDgQxY73r17/branch/master/test/uqPj73W2fTCbqCbM/step/7IGXsuKvhZ1myT35?result-id=za47mjQ58SYjqckP&path=iu3mBMtET269b9Oc%3A7IGXsuKvhZ1myT35)

```
curl -sSL https://kurl.sh/8cb4d5d | sudo bash
```

**Additional Context**

- Operating system
- Operating system version

TODO find the troubleshoot version, content of the extra redactors, and the Troubleshoot spec stored in cluster.

### LINKED ISSUE #1282: Improve collectd collector performance by detecting hostPath mount errors

Whenever the `collectd` collector runs, it mounts `/var/lib/collectd` host path. If the path does not exist, the pod get stuck in `ContainerCreating` state until its forcefully terminated. This leads to a lot of time wasting. The collector runs for `90s` unnecessarily.

**Describe the feature**

We need to figure out how we can detect if a pod is failing due to not finding the `/var/lib/collectd` directory and stop the collector it gracefully.

```
Events:
  Type     Reason       Age                From               Message
  ----     ------       ----               ----               -------
  Normal   Scheduled    59s                default-scheduler  Successfully assigned default/troubleshoot-copyfromhost-pgw7w-t2jsc to k3d-mycluster-server-0
  Warning  FailedMount  27s (x7 over 59s)  kubelet            MountVolume.SetUp failed for volume "host" : hostPath type check failed: /var/lib/collectd is not a directory
```

**Additional context**

This collector pod gets launched using a `DaemonSet`. This means that there is a pod restart policy to consider. We do not want to have it to `Never` cause there may be legitimate intermittent conditions stopping the pod from starting.

DexterYan:

- set MAX_TIME_TO_WAIT_FOR_POD_DELETION to 0 to introduce force delete, since those pods are created by troubleshoot.sh and waste time for waiting the pod to be deleted
- add checkDaemonsePodStatus function to monitor copy_from_host pod, if there is an event of mount error, that suggest the host path does not exist. We will delete the pending pod immediately

Fixes: 
#1269
#1282

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/apis/troubleshoot/v1beta2/collector_shared.go
+++ b/pkg/apis/troubleshoot/v1beta2/collector_shared.go
@@ -135,15 +135,16 @@ type Copy struct {
 }
 
 type CopyFromHost struct {
-	CollectorMeta   `json:",inline" yaml:",inline"`
-	Name            string            `json:"name,omitempty" yaml:"name,omitempty"`
-	Namespace       string            `json:"namespace" yaml:"namespace"`
-	Image           string            `json:"image" yaml:"image"`
-	ImagePullPolicy string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
-	ImagePullSecret *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
-	Timeout         string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
-	HostPath        string            `json:"hostPath" yaml:"hostPath"`
-	ExtractArchive  bool              `json:"extractArchive,omitempty" yaml:"extractArchive,omitempty"`
+	CollectorMeta    `json:",inline" yaml:",inline"`
+	Name             string            `json:"name,omitempty" yaml:"name,omitempty"`
+	Namespace        string            `json:"namespace" yaml:"namespace"`
+	Image            string            `json:"image" yaml:"image"`
+	ImagePullPolicy  string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
+	ImagePullSecret  *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
+	Timeout          string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	HostPath         string            `json:"hostPath" yaml:"hostPath"`
+	ExtractArchive   bool              `json:"extractArchive,omitempty" yaml:"extractArchive,omitempty"`
+	RetryFailedMount bool              `json:"retryFailedMount,omitempty" yaml:"retryFailedMount,omitempty"`
 }
 
 type Sysctl struct {
@@ -205,13 +206,14 @@ type TLSSecret struct {
 }
 
 type Collectd struct {
-	CollectorMeta   `json:",inline" yaml:",inline"`
-	Namespace       string            `json:"namespace" yaml:"namespace"`
-	Image           string            `json:"image" yaml:"image"`
-	ImagePullPolicy string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
-	ImagePullSecret *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
-	Timeout         string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
-	HostPath        string            `json:"hostPath" yaml:"hostPath"`
+	CollectorMeta    `json:",inline" yaml:",inline"`
+	Namespace        string            `json:"namespace" yaml:"namespace"`
+	Image            string            `json:"image" yaml:"image"`
+	ImagePullPolicy  string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
+	ImagePullSecret  *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
+	Timeout          string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	HostPath         string            `json:"hostPath" yaml:"hostPath"`
+	RetryFailedMount bool              `json:"retryFailedMount,omitempty" yaml:"retryFailedMount,omitempty"`
 }
 
 type Ceph struct {
--- a/pkg/apis/troubleshoot/v1beta2/collector_shared.go
+++ b/pkg/apis/troubleshoot/v1beta2/collector_shared.go
@@ -135,16 +135,15 @@ type Copy struct {
 }
 
 type CopyFromHost struct {
-	CollectorMeta    `json:",inline" yaml:",inline"`
-	Name             string            `json:"name,omitempty" yaml:"name,omitempty"`
-	Namespace        string            `json:"namespace" yaml:"namespace"`
-	Image            string            `json:"image" yaml:"image"`
-	ImagePullPolicy  string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
-	ImagePullSecret  *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
-	Timeout          string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
-	HostPath         string            `json:"hostPath" yaml:"hostPath"`
-	ExtractArchive   bool              `json:"extractArchive,omitempty" yaml:"extractArchive,omitempty"`
-	RetryFailedMount bool              `json:"retryFailedMount,omitempty" yaml:"retryFailedMount,omitempty"`
+	CollectorMeta   `json:",inline" yaml:",inline"`
+	Name            string            `json:"name,omitempty" yaml:"name,omitempty"`
+	Namespace       string            `json:"namespace" yaml:"namespace"`
+	Image           string            `json:"image" yaml:"image"`
+	ImagePullPolicy string            `json:"imagePullPolicy,omitempty" yaml:"imagePullPolicy,omitempty"`
+	ImagePullSecret *ImagePullSecrets `json:"imagePullSecret,omitempty" yaml:"imagePullSecret,omitempty"`
+	Timeout         string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
+	HostPath        string            `json:"hostPath" yaml:"hostPath"`
+	ExtractArchive  bool              `json:"extractArchive,omitempty" yaml:"extractArchive,omitempty"`
 }
 
 type Sysctl struct {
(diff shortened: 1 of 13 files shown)


DexterYan:
### Test Cluster
```
curl https://kurl.sh/2d2678c | sudo bash
```
### Test Support Bundle Yaml
```
kind: Collector
metadata:
  name: sentry-enterprise
spec:
  collectors:
    - secret:
        name: myapp-postgres
        key: uri
        includeValue: false
    - logs:
        selector:
          - app=sentry
          - role=worker
        limits:
          maxAge: 720h # 30*24
          maxLines: 10000
    - run:
        collectorName: ping-google
        image: flungo/netutils
        command: ["ping"]
        args: ["www.google.com"]
        timeout: 5s
    - http:
        collectorName: echo-ip
        get:
          url: https://api.replicated.com/market/v1/echo/ip
    - copyFromHost:
        collectorName: "copy kURL logs"
        image: alpine
        hostPath: "/var/log/kurl/"
        name: "logs"
        extractArchive: true
    - collectd:
        collectorName: collectd
        hostPath: /var/lib/collectd/rrd
        image: alpine
        imagePullPolicy: IfNotPresent
        timeout: 5m
```

### Result
```
run/ping-google (F)               : 5,051ms
copy-from-host/copy kURL logs (S) : 2,112ms
http/echo-ip (S)                  : 1,126ms
collectd/collectd (F)             : 1,027ms
cluster-resources (S)             : 914ms
logs/app=sentry,role=worker (S)   : 4ms
cluster-info (S)                  : 2ms
secret (S)                        : 0ms

In-cluster collectors : 2,000ms

No analyzers executed

Duration: 12,284ms
```

pkg/collect/copy_from_host.go
```diff
@@ -424,3 +430,34 @@ func deleteDaemonSet(client kubernetes.Interface, ctx context.Context, createdDS
 		}
 	}
 }
+
+func checkDaemonsePodStatus(client kubernetes.Interface, ctx context.Context, labels map[string]string, namespace string) error {
```
banjoh: Nit: `checkDaemonPodStatus`

pkg/collect/copy_from_host.go
```diff
@@ -424,3 +430,34 @@ func deleteDaemonSet(client kubernetes.Interface, ctx context.Context, createdDS
...
+	for _, pod := range pods.Items {
+		if pod.Status.Phase != corev1.PodRunning {
+			events, _ := client.CoreV1().Events(namespace).List(ctx, metav1.ListOptions{
+				FieldSelector: fmt.Sprintf("involvedObject.uid=%s", pod.UID),
+			})
+
+			for _, event := range events.Items {
+				// If the pod has a FailedMount event, it means that the pod failed to mount the volume and the pod will be stuck in the Pending state.
+				// In this case, we return an error to the caller to indicate that path does not exist.
+				if event.Reason == "FailedMount" {
```
banjoh: I would prefer if this decision was driven by an argument passed by the caller (not spec). We can add a `RetryFailedMount` to `CollectCopyFromHost` which once set to `true` will have the collector behave the same way as it is today. I say this cause the collector can be used in other scenarios in addition to `collectd`. 

`collectd`: `RetryFailedMount=false`
Default: `RetryFailedMount=true`

banjoh (changes requested):

pkg/collect/copy_from_host.go
DexterYan: Good idea, there is a issue which I need some help. One thing is that how we can set RetryFailedMount as defaul true if the user not defined it in yaml
we are using 
```
type Collectd struct {
        .....
	RetryFailedMount      bool             `json:"retryFailedMount,omitempty" yaml:"retryFailedMount,omitempty"`
}
```
since it is bool, as default, it will be false. We don't have a clear way to identify whether the user has set it to be false, or just didn't set it. 

I can switch it to string, however, that will bring another issue that we need to use "true" instead of true.

I updated the pr with introducing a argument called retryLimit. If the event count has reached retry limit, which means that k8s retried few times but it didn't work. I think that is a good signal that we should terminate the pod instead of timeout.

In default, without defined in yaml, retryLimit will be 5.

banjoh: My suggestion was to add it in the `CollectCopyFromHost` struct, not the spec. See commit below

https://github.com/banjoh/troubleshoot/commit/4e0b359525ec37bdd8c9d7392c81f7fd4ba486d5

pkg/apis/troubleshoot/v1beta2/collector_shared.go
```diff
@@ -144,6 +144,7 @@ type CopyFromHost struct {
 	Timeout         string            `json:"timeout,omitempty" yaml:"timeout,omitempty"`
 	HostPath        string            `json:"hostPath" yaml:"hostPath"`
 	ExtractArchive  bool              `json:"extractArchive,omitempty" yaml:"extractArchive,omitempty"`
+	RetryLimit      int32             `json:"retryLimit,omitempty" yaml:"retryLimit,omitempty"`
```
banjoh: This is not what I had in mind. My intention was to gracefully exit if the collectd hostpath does not exist. We would not retry since we assume the path will never appear. This would be our opinionated way of saying `collectd` is not installed. We do not want to give users more knobs to adjust (`retryLimit` extra parameter). Users who choose to collect these same files but would like to continuously retry until the hostpath is available, or timeout can use the [copy from host](https://troubleshoot.sh/docs/collect/copy-from-host/) collector.

Think of this extra bit as an implementation detail of the `collectd` collector. We need it cause the code is shared by both collectors.

banjoh (changes requested):

pkg/collect/copy_from_host.go
DexterYan: Sorry, I missed the (not spec) part. Now the PR has been updated to add it in the CollectCopyFromHost

pkg/collect/copy_from_host.go
```diff
@@ -229,6 +230,13 @@ func copyFromHostCreateDaemonSet(ctx context.Context, client kubernetes.Interfac
 	for {
 		select {
 		case <-time.After(1 * time.Second):
+			if !retryFailedMount {
```
banjoh: Shouldn't this logic be in line `458`? When we check the status of a pod, we want to exit with error when event is `FailedMount` and `retryFailedMount == false`. The implementation you have here will never check the pod status when `retryFailedMount == false`.

```go
for _, event := range events.Items {
	// If the pod has a FailedMount event, it means that the pod failed to mount the volume and the pod will be stuck in the Pending state.
	// In this case, we return an error to the caller to indicate that path does not exist.
	if event.Reason == "FailedMount" && !retryFailedMount {
		klog.V(2).Infof("pod %s has a FailedMount event: %s", pod.Name, event.Message)
		return errors.Errorf("%q path does not exist", hostPath)
	}
}
```

pkg/collect/copy_from_host.go
```diff
@@ -125,7 +126,7 @@ func (c *CollectCopyFromHost) Collect(progressChan chan<- interface{}) (Collecto
 	}
 }
 
-func copyFromHostCreateDaemonSet(ctx context.Context, client kubernetes.Interface, collector *troubleshootv1beta2.CopyFromHost, hostPath string, namespace string, generateName string, labels map[string]string) (name string, cleanup func(), err error) {
+func copyFromHostCreateDaemonSet(ctx context.Context, client kubernetes.Interface, collector *troubleshootv1beta2.CopyFromHost, retryFailedMount bool, hostPath string, namespace string, generateName string, labels map[string]string) (name string, cleanup func(), err error) {
```
banjoh: Nit pick: Instead of creating `retryFailedMount` new parameter, just make these private functions as class methods then you can reference internal fields like `c.retryFailedMount`. You might clean up other parameters out but you don't have to in this PR

```suggestion
func (c *CollectCopyFromHost) copyFromHostCreateDaemonSet(ctx context.Context, client kubernetes.Interface, collector *troubleshootv1beta2.CopyFromHost, hostPath string, namespace string, generateName string, labels map[string]string) (name string, cleanup func(), err error) {
```

banjoh (changes requested):

pkg/collect/copy_from_host.go
DexterYan: good catch, I have moved that into function and updated the tests

pkg/collect/copy_from_host.go
DexterYan: nice, no need for new parameters for private functions. PR has been updated

pkg/constants/constants.go
```diff
@@ -14,7 +14,8 @@ const (
 	// DEFAULT_LOGS_COLLECTOR_TIMEOUT is the default timeout for logs collector.
 	DEFAULT_LOGS_COLLECTOR_TIMEOUT = 60 * time.Second
 	// MAX_TIME_TO_WAIT_FOR_POD_DELETION is the maximum time to wait for pod deletion.
-	MAX_TIME_TO_WAIT_FOR_POD_DELETION = 60 * time.Second
+	// 0 seconds for force deletion.
```
banjoh, line 17: I assume you wanted to modify the grace period, not the polling interval

Here is an example of force deletion https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/copy_from_host.go#L413

With this change `MAX_TIME_TO_WAIT_FOR_POD_DELETION` just set all polling interval timeouts to `0`.

DexterYan: Before I was thinking that MAX_TIME_TO_WAIT_FOR_POD_DELETION take too long for copy_from_host. Actually, good point, MAX_TIME_TO_WAIT_FOR_POD_DELETION is also for other collectors. So I have revert this change. And add force delete to copy_from_host instead of change polling interval

pkg/collect/copy_from_host.go
```diff
@@ -373,7 +379,14 @@ func copyFilesFromHost(ctx context.Context, dstPath string, clientConfig *restcl
 
 func deleteDaemonSet(client kubernetes.Interface, ctx context.Context, createdDS *appsv1.DaemonSet, namespace string, labels map[string]string) {
 	klog.V(2).Infof("Daemonset %s has been scheduled for deletion", createdDS.Name)
-	if err := client.AppsV1().DaemonSets(namespace).Delete(context.Background(), createdDS.Name, metav1.DeleteOptions{}); err != nil {
+	zeroGracePeriod := int64(0)
+	// Foreground is used to delete the DaemonSet pods before deleting the DaemonSet
+	deletePropagationForeground := metav1.DeletePropagationForeground
+
+	if err := client.AppsV1().DaemonSets(namespace).Delete(context.Background(), createdDS.Name, metav1.DeleteOptions{
```
banjoh: Use parent context

```suggestion
	if err := client.AppsV1().DaemonSets(namespace).Delete(ctx, createdDS.Name, metav1.DeleteOptions{
```

banjoh (dismissed):

banjoh (approved):

Issue #1078, opened 2023-03-24
Add flexibility to collectors & analysers to allow running external programs


banjoh:
**Describe the rationale for the suggested feature.**

In an effort of supporting extensibility of using support-bundle & preflight binaries, it would be good to be able to run other available programs in a system without only relying on stdout/stderr/exit codes. Some example use cases that come to mind are

* Allowing cluster administrators that have home-built tools that collect and analyse data from hosts or k8s clusters to embed these tools without the need of rewriting them as specs to fit troubleshoot's current model.
* Allow embedding in automation pipelines e.g CI which come with a plethora of tools

**Describe the feature**

The "how" part is still open, but here is a suggestion that has been touched on in a community meeting (notes can be found [here](https://hackmd.io/yZbotEHdTg6TfRZBzb8Tcg)) and other various discussions.

* Extend the [run collector](https://troubleshoot.sh/docs/host-collect-analyze/run/)

Here is my fictional collector that collects audit event logs, enriches them with user data and stores the output in `$WORKSPACE_OUTPUT`. `$WORKSPACE_OUTPUT` is a unique directory created by the framework for this collector instance. Contents are them copied over to the bundle once executing the collector completes.

```yaml
apiVersion: troubleshoot.sh/v1beta2
kind: SupportBundle
metadata:
  name: run
spec:
  hostCollectors:
    - run:
        collectorName: "enriched-audit-logs"
        command: "python3"
        args: ["--timeout", "10m", "--output-dir", "$WORKSPACE_OUTPUT"]
        # Arbitrary parameters which get stored in on disk as YAML/JSON and fed passed
        # on to the command via a $CONFIG env
        config:   # perhaps call it "params"?
          username: postgres
          password: <my-pass>
          dbHost: <hostname>
          map:
            key: value
          list:
            - val1
            - val2
```

* Create a similar analyser which takes in arbitrary parameters to operate on

**Open questions**
* How would this look like in-cluster? Extending the [run pod collector](https://troubleshoot.sh/docs/collect/run-pod/)?
* Is this the only way to allow extensibility?

**Inspirations**
* https://k9scli.io/topics/plugins/
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable - GH action runners passes environment variables to jobs for them to store information that persists across jobs.

mhrabovcin:
[...]

PR #1288, opened 2023-08-02
feat: save host run file output

cwyl02:

Address the collector part of https://github.com/replicatedhq/troubleshoot/issues/1078

- added `.env` to allow user to specify which env vars to import into this host run
- added `.input` to allow user to create input files(e.g., config file, sample data for this host run). User needs to utilize `TS_INPUT_DIR` env var in their script, with the filename they specify in `.input` , e.g., `$TS_INPUT_DIR/dummy.yaml`
- added `.outputDir` to allow user to specify a name of a directory(within the directory of the collector in bundle) to host their host run command file output. Note that command run has no access to this directory

Example SupportBundle:
```yaml
kind: SupportBundle
metadata:
  name: run
spec:
  hostCollectors:
    - run:
        collectorName: "some-magic-on-host"
        command: "/home/cwyl02/custom-host-run.sh"
        args: ["--timeout", "10m"]
        env:
            - AWS_REGION=us-west-2
        inheritParentEnvs: false  # Whether to inherit all the variable or not. If true, inheritEnvs would be made redundant
        inheritEnvs:   # list of env variables to inherit from the parent process
            - SHELL
            - LANG
            - PYTHON
        outputDir: "magic-outputs"
        input:
          dummy.yaml: |-
            username: postgres
            password: <my-pass>
            dbHost: <hostname>
            map:
              key: value
            list:
              - val1
              - val2
          config.toml: |-
             title = "TOML Example"
             [owner]
             name = "Tom Preston-Werner"
```

Result bundle:
```bash
├── analysis.json
├── execution-data
│   └── summary.txt
├── host-collectors
│   └── run-host
│       └── some-magic-on-host
│           ├── collector-info.json
│           ├── collector.txt
│           └── magic-outputs
│               ├── magic-dir-0
│               │   └── t1.txt
│               └── magic-dir-1
│                   └── t2.txt
└── version.yaml
```

## Checklist

- [x] Yes
- [ ] No

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	go.opentelemetry.io/otel v1.18.0
 	go.opentelemetry.io/otel/sdk v1.18.0
 	golang.org/x/exp v0.0.0-20230905200255-921286631fa9
-	golang.org/x/sync v0.3.0
+	golang.org/x/sync v0.4.0
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.28.2
 	k8s.io/apiextensions-apiserver v0.28.1
--- a/go.sum
+++ b/go.sum
@@ -1224,8 +1224,8 @@ golang.org/x/sync v0.0.0-20210220032951-036812b2e83c/go.mod h1:RxMgew5VJxzue5/jJ
 golang.org/x/sync v0.0.0-20220601150217-0de741cfad7f/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
 golang.org/x/sync v0.0.0-20220929204114-8fcdb60fdcc0/go.mod h1:RxMgew5VJxzue5/jJTE5uejpjVlOe/izrB70Jof72aM=
-golang.org/x/sync v0.3.0 h1:ftCYgMx6zT/asHUrPw8BLLscYtGznsLAnjq5RH9P66E=
-golang.org/x/sync v0.3.0/go.mod h1:FU7BRWz2tNW+3quACPkgCx/L+uEAv1htQ0V83Z9Rj+Y=
+golang.org/x/sync v0.4.0 h1:zxkM55ReGkDlKSM+Fu41A+zmbZuaPVbGMzvvdUPznYQ=
+golang.org/x/sync v0.4.0/go.mod h1:FU7BRWz2tNW+3quACPkgCx/L+uEAv1htQ0V83Z9Rj+Y=
 golang.org/x/sys v0.0.0-20180823144017-11551d06cbcc/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20180830151530-49385e6e1522/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
 golang.org/x/sys v0.0.0-20180905080454-ebe1bf3edb33/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=


xavpaice:
Thanks for adding this.  To pass tests, you'll want to run `make schemas` and commit the changes.

Is there any testing that can be added?

Also, it's likely this will need a docs change, particularly if this is going to change paths to output which we would usually consider a breaking change.

cwyl02:
hi @xavpaice!

> Thanks for adding this. To pass tests, you'll want to run `make schemas` and commit the changes.

 I will make sure to run that before flipping to ready for review.

> Is there any testing that can be added?

I plan to add some unit test after we agree on the approach and implementation.

> Also, it's likely this will need a docs change, particularly if this is going to change paths to output which we would usually consider a breaking change.

Agree. Definitely need to add more docs for this feature.

banjoh:
This change captures the feature well. My minor comments are
- Prefix env variables e.g `TS_WORKSPACE`
- The `config` map would lead to files being dropped on disk where the names of the files are the map keys and the contents are the map values. e.g the config below would lead to files named `dummy.yaml`, `path/to/data` and `config.toml` with the contents of their corresponding values. I would suggest we rename `config` to `input` or `files` which is generic and can hold any kind of input files
```yaml
config:
  dummy.yaml: |-
    username: postgres
    password: <my-pass>
  path/to/data: "some sample data"
  config.toml: |-
    title = "TOML Example"
    [owner]
    name = "Tom Preston-Werner"
```

pkg/collect/host_run.go
```diff
@@ -32,16 +38,57 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+		bundleOutputPath          string
+		bundleOutputRelativePath  string
+		cmdConfigFilePath         string
+		cmdConfigFileRelativePath string
+	)
 
 	cmd := exec.Command(runHostCollector.Command, runHostCollector.Args...)
 
+	cmd.Env = append(os.Environ(),
+		fmt.Sprintf("BUNDLE_PATH=%s", c.BundlePath),
```
banjoh: I'm not sure we want to expose this bundle path. We would not want misbehaving tools to tamper with other collected files.

pkg/collect/host_run.go
```diff
@@ -32,16 +38,57 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+	)
+	collectorRelativePath := filepath.Join("host-collectors/run-host", collectorName)
+
+	// if we choose to save result for the command run
+	if runHostCollector.OutputDir != "" {
+		bundleOutputRelativePath = filepath.Join(collectorRelativePath, runHostCollector.OutputDir)
+		bundleOutputPath = path.Join(c.BundlePath, bundleOutputRelativePath)
+		cmd.Env = append(cmd.Env,
+			fmt.Sprintf("WORKSPACE_DIR=%s", bundleOutputPath),
```
banjoh: Do we want to create a temp dir somewhere else then copy the data back here? It might be an overkill but again, we would want to ensure that external programs won't be able to access files in the bundle path.

pkg/collect/host_run.go
```diff
@@ -54,26 +101,20 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
 		}
 	}
 
-	collectorName := c.hostCollector.CollectorName
-	if collectorName == "" {
-		collectorName = "run-host"
-	}
-	resultInfo := filepath.Join("host-collectors/run-host", collectorName+"-info.json")
-	result := filepath.Join("host-collectors/run-host", collectorName+".txt")
+	resultInfo := filepath.Join(collectorRelativePath, "collector-info.json")
```
banjoh: You probably want to revert this file names to what they were.

pkg/collect/host_run.go
cwyl02: gonna drop this, i forgot why i even added it here 😓

pkg/collect/host_run.go
cwyl02: makes sense!

pkg/collect/host_run.go
cwyl02: Makes sense to avoid giving external program the bundle path and use temp dir.

What do you think about this: instead of let the user specify the outputDir, we only ask them whether their command generate output. 
Then in the documentation we tell the user to ensure their program to utilize the env var `TS_WORKSPACE_DIR` to control where their program writes output to. This way we can avoid the data copying, which could be somehow costly.

cwyl02:
> This change captures the feature well. My minor comments are
> 
> * Prefix env variables e.g `TS_WORKSPACE`
> * The `config` map would lead to files being dropped on disk where the names of the filed are the map keys and the contents are the map valued. e.g the config below would lead to a files named `dummy.yaml`, `path/to/data` and `config.toml` with the contents if their corresponding values. I would suggest we rename `config` to `input` or `files` which is generic and can hold any kind of input files
> 
> ```yaml
> config:
>   dummy.yaml: |-
>     username: postgres
>     password: <my-pass>
>   path/to/data: "some sample data"
>   config.toml: |-
>     title = "TOML Example"
>     [owner]
>     name = "Tom Preston-Werner"
> ```

Like this idea, although maybe we should just not allow `/` in the file name to avoid the need of a logic to handle the different OS, and the logic to conditionally create a dir?

banjoh:
> Like this idea, although maybe we should just not allow `/` in the file name to avoid the need of a logic to handle the different OS, and the logic to conditionally create a dir?

My intention with this is to reduce friction when landing files for programs/scripts to consume. This way, programs/scripts would not need to arrange files to directories before running. **But I think we can leave this out since its a bit of an edge case. If need arises, we can introduce it**

With regards to `env`s, I think the map should list additional environment variables to add to the command being launched instead of what to inherit from the parent process

Additionally, for the security focused folks, I think we need to be explicit on whether to inherit the entire environment of the parent process using a key such as `inheritParentEnvs` or something similar and another `inheritEnv` which explicitly lists what to inherit. By default, a process created using the [exec command will inherit the parent's environment](https://pkg.go.dev/os/exec#Cmd).

We will have some always present environment variables such as `PATH`, `KUBECONFIG`, `TS_WORKSPACE` etc though

Here is a complete example

```yaml
       env:
            - AWS_REGION=us-east-1
            - AWS_ACCESS_KEY_ID={{ aws.key-id.template }}
            - AWS_SECRET_ACCESS_KEY={{ aws.secret.template }}
       inheritParentEnvs: false  # Whether to inherit all the variable or not. If true, inheritEnvs would be made redundant
       inheritEnvs:   # list of env variables to inherit from the parent process
            - SHELL
            - LANG
            - PYTHON
```

For backward compatibility `inheritParentEnvs` would need to be true by default

pkg/collect/host_run.go
```diff
@@ -66,14 +119,55 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
+	for _, key := range guaranteedEnvs {
+		guaranteedEnvVal, found := os.LookupEnv(key)
+		if found {
+			cmd.Env = append(cmd.Env,
+				fmt.Sprintf("%s=%s", key, guaranteedEnvVal))
+		}
+	}
+
+	runHostCollector := c.hostCollector
+	if runHostCollector.InheritParentEnvs == true {
```
banjoh: Do you handle `InheritParentEnvs = false`? I'm expecting to see the `cmd.Env` cleared off the defaults it inherits from the parent.

banjoh (commented):
Changes look good now. Could you please add some tests as well?

pkg/collect/host_run.go
cwyl02: handled by always setting `cmd.Env`

pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
```diff
@@ -180,8 +180,13 @@ type HostServices struct {
...
 	HostCollectorMeta `json:",inline" yaml:",inline"`
-	Command           string   `json:"command"`
-	Args              []string `json:"args"`
+	Command           string            `json:"command"`
+	Args              []string          `json:"args"`
+	OutputDir         string            `json:"outputDir,omitempty" yaml:"outputDir,omitempty"`
+	Input             map[string]string `json:"input,omitempty" yaml:"input,omitempty"`
+	Env               []string          `json:"env,omitempty" yaml:"env,omitempty"`
+	InheritEnvs       []string          `json:"inheritEnvs" yaml:"inheritEnvs,omitempty"`
+	InheritParentEnvs bool              `json:"inheritParentEnvs" yaml:"inheritParentEnvs,omitempty"`
```
banjoh: Lets have this be `InheritParentEnvs *bool` so we can default it to true. We want it to be `true` to ensure we are compatible with the current behaviour

Snippet to show what I mean https://goplay.tools/snippet/qRFk96B5Wob

... or maybe change the name to `IgnoreParentEnvs` with a default of `false` :). **I like this more**

pkg/collect/host_run.go
```diff
@@ -66,14 +85,94 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
+				cmd.Env = append(cmd.Env,
+					fmt.Sprintf("%s=%s", key, envVal))
+			}
+		} else {
+			cmd.Env = append(cmd.Env, os.Environ()...)
+		}
+	}
+
+	if runHostCollector.Env != nil {
+		for i := range runHostCollector.Env {
```
banjoh: My custom envs have an extra `=` sign at the end

```yaml
        env:
          - AWS_REGION=us-east-1
          - AWS_ACCESS_KEY_ID={{ aws.key-id.template }}
          - AWS_SECRET_ACCESS_KEY={{ aws.secret.template }}
```

Output of `env | grep AWS` in my custom script
```
AWS_REGION=us-east-1=
AWS_SECRET_ACCESS_KEY={{ aws.secret.template }}=
AWS_ACCESS_KEY_ID={{ aws.key-id.template }}=
```

pkg/collect/host_run.go
```diff
@@ -66,14 +85,94 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
+		}
+	}
 
-	runHostOutput := map[string][]byte{
-		resultInfo: b,
-		result:     stdout.Bytes(),
+	// if we choose to save result for the command run
+	if runHostCollector.OutputDir != "" {
+		cmdOutputTempDir, err = os.MkdirTemp("", runHostCollector.OutputDir)
+		defer os.RemoveAll(cmdOutputTempDir)
```
banjoh: `TS_WORKSPACE_DIR` directory gets removed after `processEnvVars()` function returns. My script had this error

```sh
I1018 12:33:03.988617   17142 host_run.go:90] ls: cannot access '/var/folders/19/bp6c9chj0sgcpcxmxxl69s040000gn/T/magic-outputs1460992333': No such file or directory
```

pkg/collect/host_run.go
```diff
@@ -66,14 +85,94 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
+			return "", errors.New(fmt.Sprintf("failed to created temp dir for: %s", runHostCollector.OutputDir))
+		}
+		cmd.Env = append(cmd.Env,
+			fmt.Sprintf("TS_WORKSPACE_DIR=%s", cmdOutputTempDir),
+		)
+	}
+
+	if runHostCollector.Input != nil {
+		cmdInputTempDir, err := os.MkdirTemp("", "input")
+		defer os.RemoveAll(cmdInputTempDir)
```
banjoh: This will get deleted as well

pkg/collect/host_run.go
```diff
@@ -66,14 +85,94 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
-		resultInfo: b,
-		result:     stdout.Bytes(),
+	// if we choose to save result for the command run
+	if runHostCollector.OutputDir != "" {
+		cmdOutputTempDir, err = os.MkdirTemp("", runHostCollector.OutputDir)
+		defer os.RemoveAll(cmdOutputTempDir)
+		if err != nil {
+			return "", errors.New(fmt.Sprintf("failed to created temp dir for: %s", runHostCollector.OutputDir))
+		}
+		cmd.Env = append(cmd.Env,
```
banjoh: Should we change the `CWD` of the command to `TS_WORKSPACE_DIR`? The working dir is currently where `support-bundle` is launched from

pkg/collect/host_run.go
```diff
@@ -12,9 +14,12 @@ import (
...
 
 type HostRunInfo struct {
-	Command  string `json:"command"`
-	ExitCode string `json:"exitCode"`
-	Error    string `json:"error"`
+	Command   string   `json:"command"`
+	ExitCode  string   `json:"exitCode"`
+	Error     string   `json:"error"`
+	OutputDir string   `json:"outputDir"`
+	Input     string   `json:"input"`
```
banjoh: Should this be a `map[string]string` or `map[string]any`? I could not load this yaml. How is the `input` parameter meant to behave? Isn't the `key` meant to be the filename and the the `value` be the contents of the file, all stored as flat files in `TS_INPUT_DIR` directory?

```yaml
        input:
          dummy.yaml: My sample data REDACTED SECRET
          config:   # I cannot load this config
            dummy.yaml: |-
              username: postgres
              password: <my-pass>
            config.toml: |-
              title = "TOML Example"
              [owner]
              name = "Tom Preston-Werner"
```

banjoh (changes requested):
We are almost there. I managed to run the collector locally but could not collect anything since `TS_WORKSPACE_DIR` does not exist. I added a few more comments in the PR (see below).

Here are the files I used to test
- custom script
```sh
#!/usr/bin/env bash

pwd # Should we ensure the command's CWD is the $TS_WORKSPACE_DIR? Its currently the dir where support bundle is launched
env | grep AWS
ls -al $TS_WORKSPACE_DIR    # This fails. Dir does not exist
ls -al $TS_INPUT_DIR    # This fails. Dir does not exist
cat $TS_INPUT_DIR/dummy.yaml
echo "Drop my data: $($TS_INPUT_DIR/dummy.yaml)" >> $TS_WORKSPACE_DIR/mydata.txt # I'll have a redactor redact my secret later on
```
- Spec
```yaml
kind: SupportBundle
metadata:
  name: run
spec:
  hostCollectors:
    - run:
        collectorName: "ping-google"
        command: "custom-host-run.sh"
        env:
          - AWS_REGION=us-east-1
          - AWS_ACCESS_KEY_ID={{ aws.key-id.template }}
          - AWS_SECRET_ACCESS_KEY={{ aws.secret.template }}
        inheritParentEnvs: false
        inheritEnvs:
          - SHELL
          - LANG
          - TERM
        outputDir: "magic-outputs"
        input:
          dummy.yaml: My sample data REDACTED SECRET
          # config:   # I cannot load this config
          #   dummy.yaml: |-
          #     username: postgres
          #     password: <my-pass>
          #   config.toml: |-
          #     title = "TOML Example"
          #     [owner]
          #     name = "Tom Preston-Werner"
---
kind: Redactor
metadata:
  name: redactor-spec
spec:
  redactors:
    - name: redact-text
      removals:
        values:
        -  REDACTED SECRET
```

banjoh (changes requested):
With regards to tests, please add
- An example in [examples directory](https://github.com/replicatedhq/troubleshoot/tree/main/examples/collect/host) similar to what I posted in my previous comment. The command part can be `bash -c <some inline code e.g env, cat $TS_INPUT_DIR/dummy.yaml, ls -al...>` so as not to be forced to add an external script
- Unit tests around processEnvVars
  - Ensure directories that need to be created are created, and exist.
  - Ensure cmd has the correct parameters set i.e env vars, CWD etc
- Add at least one level 2 (`klog.V(2)`) log, more if possible, that should show what the collector is doing like [in this example](https://github.com/replicatedhq/troubleshoot/blob/42fc4f06de1094137e8510816c38726bca3b073d/pkg/collect/host_filesystem_performance.go#L409).

pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
cwyl02: `IgnoreParentEnvs` makes more sense to me too

pkg/collect/host_run.go
cwyl02: good catch!!

pkg/collect/host_run.go
cwyl02: great question. `TS_WORKSPACE_DIR` is the temporary output directory so it only exists when user specify `.outputDir`. 

> The working dir is currently where support-bundle is launched from

I think this is the existing behavior. I thought it is confusing if the `CWD` changes when the `.outputDir` is specified vs not specified?

pkg/collect/host_run.go
cwyl02: So the input files will be written into the `TS_INPUT_DIR`, which is a temp directory. It will always be deleted. I previously also choose to output the input file to somewhere but then [removed that part](https://github.com/replicatedhq/troubleshoot/pull/1288/commits/08714e06d00efdd6f72e271720c2ac2478a30c30) -- talked with Martin and we thought input might be sensitive and might be good to just skip storing it.

The user triggers this bundle will have input file on their hand anyways but anything output to that bundle might do harm when the bundle is given to someone else.

cwyl02: or btw it should be

```yaml
        input:
          dummy.yaml: My sample data REDACTED SECRET
          dummy2.yaml: |-
            username: postgres
            password: <my-pass>
          config.toml: |-
            title = "TOML Example"
            [owner]
            name = "Tom Preston-Werner"
```

that's why its `map[string]string`

pkg/collect/host_run.go
cwyl02: nvm I think the spec i was using was with absolute path so masked the issue. let me fix that

pkg/collect/host_run.go
cwyl02: moved out from the func. realized this cannot be easily moved into the function due to `defer`

pkg/collect/host_run.go
cwyl02: same as the other comment, moved this out from the func.

pkg/collect/host_run.go
cwyl02: fixed

cwyl02: fixed [here](https://github.com/replicatedhq/troubleshoot/pull/1288/commits/e561a2b86387a84c100034f064b7239223e85cee#diff-533bcc91b87f8c262408a8d74f39eecbf2bd51fd01bc56ba38bfce0582011caaR166)

cwyl02:
> ```yaml
> command: "custom-host-run.sh"
> ```

after fixing the issue this should work although you need to change command to: "./custom-host-run.sh"

cwyl02:
closed this PR accidentally 🙈  and opened a new one https://github.com/replicatedhq/troubleshoot/pull/1376

PR #1291, opened 2023-08-07
feat(redact): use a scan regex for default redact rule of lines to improve cpu usage and reduce time cost 

### LINKED ISSUE #1151: Slow performance generating a host collector support bundle
**Bug Description**

Performed the following procedure to generate a host collector support bundle:
1. Installed support bundle binary: `curl -L https://github.com/replicatedhq/troubleshoot/releases/latest/download/support-bundle_linux_amd64.tar.gz | tar xzvf -`
2. `sudo ./support-bundle --interactive=false https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/host/default.yaml`
 --- I waited at least 5 minutes for the support bundle to generate and then cancelled.

I originally thought that the support bundle was hanging because it was taking so long; however, I did not realize that the process was not hanging but just extremely slow. I found this out by just letting the process run until it completed.  The process generally took 6 mins +/- 15 sec to complete; however, I experienced wait times for up to ~11-12 minutes.

Environment:
- 3 node kurl installed k8s cluster using [linode shared cpu instances ](https://www.linode.com/products/shared/)

**Expected Behavior**
I expected the host collector support bundle generation process to be completed in <90-120 secs which is what I have always experienced when generating a host collect support bundle prior.

Note: I experienced the same behavior on all three nodes.

**Steps To Reproduce**
see bug description above for steps to reproduce.

**Additional Context**

   - Replicated Troubleshoot v0.62.1-26-g97efe83
   - Go version: go version go1.20.3 linux/amd64
- Operating system: Ubuntu
- Operating system version: 22.04.2 LTS (Jammy Jellyfish)
   -Further investigation elucidated that the redactor process appears to be the culprit in causing the long processing time.  I will articulate details in how I came to suspect the redactor process in my next comments in this issue.

DexterYan:

- add goroutine to RedactResult
- add simplified scan regex for both single line redactor and multiple lines redactor
- use scan regex to check the line which in lower case first, if it didn't match, then continue next
- if scan regex matched the line, then it will run original regex check and string replace
- the scan regex is a simple version of  original regex to reduce cpu usage and improve speed
- use bufio.NewScanner to improve read performance
- klog the size of redact file for better monitoring

Fixes: #1151
## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/result.go
+++ b/pkg/collect/result.go
@@ -153,7 +153,7 @@ func (r CollectorResult) ReplaceResult(bundlePath string, relativePath string, r
 	}
 	defer tmpFile.Close()
 
-	_, err = io.CopyBuffer(tmpFile, reader, make([]byte, 1024*1024))
+	_, err = io.Copy(tmpFile, reader)
 	if err != nil {
 		return errors.Wrap(err, "failed to write tmp file")
 	}
--- a/pkg/collect/result.go
+++ b/pkg/collect/result.go
@@ -142,18 +142,21 @@ func (r CollectorResult) ReplaceResult(bundlePath string, relativePath string, r
 		return nil
 	}
 
-	targetPath := filepath.Join(bundlePath, relativePath)
-
-	// Open the file directly in the desired location
-	outFile, err := os.Create(targetPath)
+	tmpFile, err := os.CreateTemp("", "replace-")
 	if err != nil {
-		return errors.Wrap(err, "failed to create file")
+		return errors.Wrap(err, "failed to create temp file")
 	}
-	defer outFile.Close()
+	defer tmpFile.Close()
 
-	_, err = io.CopyBuffer(outFile, reader, make([]byte, 1024*1024))
+	_, err = io.CopyBuffer(tmpFile, reader, make([]byte, 1024*1024))
 	if err != nil {
 		return errors.Wrap(err, "failed to write tmp file")
+	}
+
+	// This rename should always be in /tmp, so no cross-partition copying will happen
+	err = os.Rename(tmpFile.Name(), filepath.Join(bundlePath, relativePath))
+	if err != nil {
+		return errors.Wrap(err, "failed to rename tmp file")
 	}
 
 	return nil
--- a/pkg/collect/result.go
+++ b/pkg/collect/result.go
@@ -142,21 +142,18 @@ func (r CollectorResult) ReplaceResult(bundlePath string, relativePath string, r
 		return nil
 	}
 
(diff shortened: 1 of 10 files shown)


pkg/collect/result.go
```diff
@@ -142,21 +142,18 @@ func (r CollectorResult) ReplaceResult(bundlePath string, relativePath string, r
 		return nil
 	}
 
-	tmpFile, err := os.CreateTemp("", "replace-")
-	if err != nil {
-		return errors.Wrap(err, "failed to create temp file")
-	}
-	defer tmpFile.Close()
+	targetPath := filepath.Join(bundlePath, relativePath)
```
CpuID: curious re the old vs new approaches to handling this:

- the old creates a file in `/tmp`, then writes content to it, then renames it (in the same partition supposedly)
- the new creates the desired filename instead, and writes to that directly

is there a noticeable IO improvement to the new...? In theory both approaches write to the partition once, and the rename operation should be very fast...?

Not opposed to the change in behaviour, just curious about the before/after benchmarks :) I know there were issues around this operation in the past being slow.

CpuID (commented):
looking reasonably good so far...

would love to see some test coverage passing then I'll review/approve 👍

pkg/collect/result.go
DexterYan: Very good point, after testing, none of the files were moving between different partitions. And if we write to the same file, the previous io.read will conflict with io.Copy which causes less file size to be redact. 
For example, we have `apiserver-audit-logs/k8s-audit-2023-08-08T01-47-13.742.log` with origin size `100M`.
Without writing to a tmp file and rename, the result file size is 4K

That's why it feels like we have reduce the redact time, actually, it is a mistake.

So currently, I switch the current function to use goroutine to reduce waiting time of redacting each files.

It can reduce the time 
```
Duration: 230,272ms                 ->                188,182ms
```

DexterYan:
## Test Yaml
```
kind: SupportBundle
metadata:
  name: default
spec:
  hostCollectors:
    - copy:
        collectorName: "apiserver-audit-logs"
        path: /var/log/apiserver/k8s-audit*
```

## Troubleshoot 0.70.2
```
apiserver-audit-logs (S) : 1,134ms

Host collectors : 271,338ms

No analyzers executed

Duration: 272,473ms
```


## New
```
apiserver-audit-logs (S) : 1,043ms

Host collectors : 100,786ms

No analyzers executed

Duration: 101,830ms
```

CpuID (approved):

PR #1298, opened 2023-08-11
Feat(redact): reduce regex cpu usage and improve redact efficient  

### LINKED ISSUE #1151: Slow performance generating a host collector support bundle

DexterYan:

- if the regex is case insensitive, it is not wise to use camel case in regex. That needs be updated in docs
For example. 
```
(?i)(Database *= *)(?P<mask>[^\;]+)(;)   
-> (?i)(database *= *)(?P<mask>[^\;]+)(;)


(?i)(\\\"name\\\":\\\"[^\"]*OWNER_?ACCOUNT\\\",\\\"value\\\":\\\")(?P<mask>[^\"]*)(\\\") 
-> (?i)(\\\"name\\\":\\\"[^\"]*owner_?account\\\",\\\"value\\\":\\\")(?P<mask>[^\"]*)(\\\")
```
- if the regex is case insensitive, all the text read from files of waiting redact will be transformed to lowercase for normalize. 
```
PWD = abcdef;   ->  pwd = abcdef;
```
- case insensitive regex will be split into case-sensitive and insensitive parts. case-sensitive part will be used to check those lowercase texts which can reduce CPU usage. 
```
Original regex:  (?i)(pwd *= *)(?P<mask>[^\;]+)(;) 
For matching
use (pwd *= *)(?P<mask>[^\;]+)(;) 
agaist pwd = abcdef;
```

- if the regex is case sensitive, it will keep passing to redact directly with existing logical
- add goroutine to RedactResult
- use bufio.NewWriter and bufio.NewScanner to improve read and write performance
- klog the size of redact file for better monitoring

Fixes: #1151
## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ build:
 
 .PHONY: support-bundle
 support-bundle:
-	go build ${BUILDFLAGS} ${LDFLAGS} -o bin/support-bundle github.com/replicatedhq/troubleshoot/cmd/troubleshoot
+	 GOOS=linux GOARCH=amd64 go build ${BUILDFLAGS} ${LDFLAGS} -o bin/support-bundle-linux github.com/replicatedhq/troubleshoot/cmd/troubleshoot
 
 .PHONY: preflight
 preflight:
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ build:
 
 .PHONY: support-bundle
 support-bundle:
-	 GOOS=linux GOARCH=amd64 go build ${BUILDFLAGS} ${LDFLAGS} -o bin/support-bundle-linux github.com/replicatedhq/troubleshoot/cmd/troubleshoot
+	go build ${BUILDFLAGS} ${LDFLAGS} -o bin/support-bundle github.com/replicatedhq/troubleshoot/cmd/troubleshoot
 
 .PHONY: preflight
 preflight:
--- a/pkg/collect/redact.go
+++ b/pkg/collect/redact.go
@@ -5,9 +5,11 @@ import (
 	"bytes"
 	"compress/gzip"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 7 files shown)


xavpaice:
I am concerned at this change in that there's a lot going on, and some of the items are likely to be worrying for the user.  Let's split this up:

1. update the built in regex to remove any camel case in case-insensitive searches, and remove case-insensitive if it's not absolutely required
2. update the docs as suggested
3. review the option to convert everything to lower case and search that without the `(?i)`.  It's understood that if a search is case insensitive, the result isn't going to be affected if we convert case on the string searched, just need to be very careful that there aren't side-effects with capture groups, and to make sure that the redacted output is not converted to lower case when it's saved.

PR #1332, opened 2023-09-11
chore(redactors): memory consumption improvements

### LINKED ISSUE #1331: Redactors in v0.71.1 consume too much memory
**Bug Description**

In certain scenarios, running redactor when collector a support bundle can consume a lot of memory leading to OOM signals killing pods.

Here is are memory profiles of this happening in [KOTS ](https://github.com/replicatedhq/kots).

**With v0.71.1**
Screen Shot 2023-09-08 at 10 26 46 AM

**With v0.70.3**
Screen Shot 2023-09-08 at 10 23 20 AM

~~NOTE: This has not been reproduced when using the CLI.~~ From comments below

- Checkout `v0.71.1` of troubleshoot
- Build binaries `make build`
- Collect and redact support bundle - `./bin/support-bundle ./support-bundle.yaml --memprofile=mem.conf --redactors ./redactor.yaml` with [support-bundle.yaml](https://github.com/replicatedhq/troubleshoot/files/12563266/support-bundle.yaml.txt) and [redactor.yaml](https://github.com/replicatedhq/troubleshoot/files/12563268/redactor.yaml.txt)
- Analyse memory allocations - `go tool pprof -http :8888 -alloc_space mem.prof` will start an http server on port 8888 

**Expected Behavior**

Redaction code should not consume excess memory.

**Steps To Reproduce**

Include the commands to reproduce the issue including any output. Any information that will help us to understand the problem is useful.  Feel free to paste long output into a [Github gist](https://gist.github.com) and include the link here.

To be fleshed

**Additional Context**

- Operating system
- Operating system version

banjoh:

Redactor memory optimisations
- Reduce buffer size allocated for io reader scanner - This was the smoking gun
- Prefer working with `[]byte` over `string`s. Strings are immutable and lead to more memory allocations
- Use a regex cache to store compiled regular expressions. This led to a minimal reduction of consumption of clock cycles but a good optimisation
- Use `bufio.Scanner` where possible. Reading raw bytes until a newline is reached has the likely hood of using up too much heap memory leading to OOMs. This is possible when a large enough file, without newlines, is read. With `bufio.Scanner`, there is a `10MB` limit

Fixes: #1331

## Checklist

- [ ] Yes
- [x] No


banjoh:
### Improvements analysis

**Environment used**
* Decent Mac laptop
* Empty `k3d` cluster
* Spec collecting cluster resources only
```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - clusterResources: {}
```
**Commands used**
* `support-bundle spec.yaml --debug --memprofile=mem.prof`
* `go tool pprof -http=":8001" -alloc_space mem.prof`

**Results**

I ran these tests several times to reduce the error margin of resources being consumed in my laptop

 - Original run with code in `main` branch
orig

- Run after reducing buffer size to `4096` (4KB). This is the default size used by other std libraries such as `bufio`
with_buffer_limit

- Run after the `string` to `[]byte` data type change
final

**Notes**
- The buffer size change made the largest impact
- The `string` to `[]byte` change reduced memory consumption by about 11-14% after a number of runs. The sample above shows a 13% reduction (137MB -> 119MB of total memory)

diamonwiggins:
@banjoh Working to get myself educated on these changes, so apologies in advance if none of the following makes sense :) A few questions so far:

1. With regards to `bufio.Scanner` vs. `bufio.Reader`, could we have accomplished the same result by keeping `bufio.Reader`, but just lowering the buffer size? Is the major difference here that `bufio.Scanner` delimits at every newline and thus there's more of an implicit limit?

> Use bufio.Scanner where possible. Reading raw bytes until a newline is reached has the likely hood of using up too much heap memory leading to OOMs. This is possible when a large enough file, without newlines, is read. With bufio.Scanner, there is a 10MB limit

Just to make sure I understand, if there's a large file without newlines it'll fall back to the buffer size same as `bufio.Reader` correct? I wonder if what's most important here is having the correct buffer size set as to optimize for memory consumption in our use case instead of relying on newlines? Would that result in the most complete solution here without any edge cases? 

Also, you mentioned:

> Prefer working with []byte over strings. Strings are immutable and lead to more memory allocations

With `Reader` you wouldn't need to cast from a string into bytes to begin with right?

2. You mentioned that `bufio.Reader` was the smoking gun, but it doesn't seem like we've had changes to that code for a while? Any details on why we're just seeing this now, or has this been an unresolved problem for a while?

3. How is current performance of redaction impacted by these changes? Are we knowingly sacrificing performance here in order to lower memory consumption? Was this a case of the increased memory consumption not being worth the potential increase in speed of redaction, or were we leaking memory?

arcolife:
FWIW, I see these pprof cumulative figures on M2 pro:

**pre-fix (main)**:
- `bufio.(*Scanner).Text`: 12.01GB
- `redact.readLine`: 11.22GB
- `SingleLineRedactor`: 24.64GB
- `MultiLineRedactor`: 17.11GB

```sh
# ./bin/support-bundle test.yaml --debug --memprofile=mem.prof
 Collecting support bundle ⠇ [apiserver-audit-logs] Running host collector...
support-bundle-2023-09-14T03_21_59.tar.gz

apiserver-audit-logs (S) : 359ms

Host collectors : 17,366ms

No analyzers executed

Duration: 17,726ms

# go tool pprof -alloc_space mem.prof 
# (pprof) top
Showing nodes accounting for 41.38GB, 99.05% of 41.78GB total
Dropped 193 nodes (cum <= 0.21GB)
Showing top 10 nodes out of 19
      flat  flat%   sum%        cum   cum%
   17.31GB 41.43% 41.43%    17.31GB 41.43%  strings.(*Builder).grow
   12.01GB 28.75% 70.18%    12.01GB 28.75%  bufio.(*Scanner).Text (inline)
   11.22GB 26.86% 97.04%    11.22GB 26.86%  github.com/replicatedhq/troubleshoot/pkg/redact.readLine
    0.37GB  0.88% 97.93%     0.37GB  0.88%  fmt.(*buffer).writeString (inline)
    0.27GB  0.64% 98.56%    24.64GB 58.98%  github.com/replicatedhq/troubleshoot/pkg/redact.(*SingleLineRedactor).Redact.func1
    0.14GB  0.32% 98.88%     0.28GB  0.68%  regexp.(*Regexp).ReplaceAllString
    0.07GB  0.17% 99.05%    17.11GB 40.95%  github.com/replicatedhq/troubleshoot/pkg/redact.(*MultiLineRedactor).Redact.func1
         0     0% 99.05%     0.37GB  0.88%  fmt.(*fmt).fmtS
         0     0% 99.05%     0.37GB  0.88%  fmt.(*fmt).padString
         0     0% 99.05%     0.37GB  0.88%  fmt.(*pp).doPrintf
```

**post-fix (feature branch)**:

- `SingleLineRedactor`: 12.14GB
- `MultiLineRedactor`: 12.33GB
- `redact.readLine` / `getNextTwoLines`: 5.85GB


```sh
# ./bin/support-bundle test.yaml --debug --memprofile=mem.prof
 Collecting support bundle ⠼ [apiserver-audit-logs] Running host collector...
support-bundle-2023-09-14T03_08_42.tar.gz

apiserver-audit-logs (S) : 303ms

Host collectors : 16,986ms

No analyzers executed

Duration: 17,290ms

# go tool pprof -alloc_space mem.prof 
# (pprof) top
(pprof) top
Showing nodes accounting for 24.22GB, 98.91% of 24.49GB total
Dropped 132 nodes (cum <= 0.12GB)
Showing top 10 nodes out of 13
      flat  flat%   sum%        cum   cum%
   17.10GB 69.81% 69.81%    17.17GB 70.10%  bytes.ToLower
    5.85GB 23.91% 93.71%     5.85GB 23.91%  github.com/replicatedhq/troubleshoot/pkg/redact.readLine
    0.71GB  2.90% 96.62%    12.33GB 50.36%  github.com/replicatedhq/troubleshoot/pkg/redact.(*MultiLineRedactor).Redact.func1
    0.41GB  1.69% 98.31%     0.41GB  1.69%  regexp.(*bitState).push (inline)
    0.15GB   0.6% 98.91%     0.15GB   0.6%  regexp.(*Regexp).replaceAll
         0     0% 98.91%    12.14GB 49.58%  github.com/replicatedhq/troubleshoot/pkg/redact.(*SingleLineRedactor).Redact.func1
         0     0% 98.91%     5.85GB 23.91%  github.com/replicatedhq/troubleshoot/pkg/redact.getNextTwoLines
         0     0% 98.91%     0.54GB  2.22%  regexp.(*Regexp).Match (inline)
         0     0% 98.91%     0.15GB   0.6%  regexp.(*Regexp).ReplaceAll
         0     0% 98.91%     0.53GB  2.18%  regexp.(*Regexp).backtrack
```
---

In either cases:
- ~~Process' memory jumped only a couple hundred MBs~~
- Duration is similar pre/post fix
- On `c2-standard-8` GCP, duration is about 37,000ms

Data used:
```yaml
# cat test.yaml
kind: SupportBundle
metadata:
  name: default
spec:
  hostCollectors:
    - copy:
        collectorName: "apiserver-audit-logs"
        path: ./log/k8s-audit*
```

```sh
# ls log
-rw-------@  1 arcolife  staff   100M Sep 14 02:29 k8s-audit-2023-09-06T07-32-52.651.log
-rw-------@  1 arcolife  staff   100M Sep 14 02:29 k8s-audit-2023-09-06T10-52-44.261.log
-rw-------@  1 arcolife  staff   100M Sep 14 02:29 k8s-audit-2023-09-06T14-07-35.432.log
-rw-------@  1 arcolife  staff   100M Sep 14 02:29 k8s-audit-2023-09-06T17-28-43.581.log
-rw-------@  1 arcolife  staff   100M Sep 14 02:29 k8s-audit-2023-09-07T07-14-41.214.log
-rw-------@  1 arcolife  staff   100M Sep 14 02:29 k8s-audit-2023-09-07T13-30-36.751.log
-rw-------@  1 arcolife  staff    49M Sep 14 02:29 k8s-audit.log

```

DexterYan:
I did this one on aws ec2 machine with 4 CPU, 16G ram

with 
```
kind: SupportBundle
metadata:
  name: default
spec:
  hostCollectors:
    - copy:
        collectorName: "apiserver-audit-logs"
        path: /var/log/apiserver/k8s-audit*
```

main branch: 
Screenshot 2023-09-14 at 4 44 13 PM
Process peak usage:
Screenshot 2023-09-14 at 4 32 13 PM
mem.prof:
[main-mem.prof.txt](https://github.com/replicatedhq/troubleshoot/files/12604818/main-mem.prof.txt)


improved branch:
Screenshot 2023-09-14 at 4 45 34 PM
Process peak usage:
Screenshot 2023-09-14 at 4 39 15 PM
mem.prof:

[improved-mem.prof.txt](https://github.com/replicatedhq/troubleshoot/files/12604819/improved-mem.prof.txt)

DexterYan:
I think we need to read a line without break for regex, otherwise the regex search is not precisely. So bufio.Scanner read by line can help it. However, large file in one line that is quite difficult to handle in current way

banjoh:
@diamonwiggins 

> With regards to bufio.Scanner vs. bufio.Reader, could we have accomplished the same result by keeping bufio.Reader, but just lowering the buffer size? Is the major difference here that bufio.Scanner delimits at every newline and thus there's more of an implicit limit?

Difference between `bufio.Scanner` and `bufio.Reader` is that `bufio.Scanner` accepts a maximum size its internal buffer can grow to. This gives use the ability to have a limit of consuming a large file with no newlines breaking it. With `bufio.Reader`, we can achieve the same by effectively counting how much data we read, but that is what `bufio.Scanner` does

> Just to make sure I understand, if there's a large file without newlines it'll fall back to the buffer size same as bufio.Reader correct?

No. Reading that file will result in an error in the lines of `a line in '<somepath>' file exceeded 10485760 bytes max length`. That max number is hardcoded to 10MB. Minified JSON files for example have no newlines.

> I wonder if what's most important here is having the correct buffer size set as to optimize for memory consumption in our use case instead of relying on newlines?

Our implementation of redaction has been designed to process one line at time. This optimization is addressing huge one line inputs

> With Reader you wouldn't need to cast from a string into bytes to begin with right?

No. Previous implementation converts bytes to strings which leads to new string objects being created. I'm proposing using the bytes as they are read.

> You mentioned that bufio.Reader was the smoking gun

I meant the buffer created [here](https://github.com/replicatedhq/troubleshoot/blob/514c86d891b867fad6de24409109ea449e8b2c27/pkg/redact/single_line.go#L54). This line gets called multiple times and each time it allocates 1MB or memory. Using a bundle with 100 files, 1800MBs will be allocated "instantly" (depending on how fast goroutines start up)

> How is current performance of redaction impacted by these changes? Are we knowingly sacrificing performance here in order to lower memory consumption?

In my tests, it doesn't look like the performance is getting impacted. I can provide some measurements here.

> Was this a case of the increased memory consumption not being worth the potential increase in speed of redaction, or were we leaking memory?

We we a bit too greedy in the memory we reserved as buffers for data read in from files.

diamonwiggins:
> Our implementation of redaction has been designed to process one line at time. This optimization is addressing huge one line inputs

Thanks for the explanation on everything. This in particular was helpful to understand.

diamonwiggins (dismissed):

laverya (dismissed):

pkg/redact/literal.go
```diff
@@ -28,32 +31,36 @@ func (r literalRedactor) Redact(input io.Reader, path string) io.Reader {
...
-				return
-			}
+			line := scanner.Bytes()
 
-			clean := strings.ReplaceAll(line, r.matchString, MASK_TEXT)
+			clean := bytes.ReplaceAll(line, r.match, maskTextBytes)
 
-			// io.WriteString would be nicer, but scanner strips new lines
-			fmt.Fprintf(writer, "%s\n", clean)
+			_, err = writer.Write(append(clean, '\n')) // Append newline since scanner strips it
```
emosbaugh: i think if the line is really long and you use `append(line, '\n')` you will allocate another up to 10M. wouldnt it just be better to write twice?

```suggestion
			_, err = writer.Write(clean)
			_, err = writer.Write('\n') // Append newline since scanner strips it
```

this goes for all occurrences of append below

pkg/collect/redact.go
```diff
@@ -16,21 +16,28 @@ import (
...
+// also spawns goroutines to redact files in tar archives and
+// other goroutines for each redactor spec.
+const MAX_CONCURRENT_REDACTORS = 10
+
 func RedactResult(bundlePath string, input CollectorResult, additionalRedactors []*troubleshootv1beta2.Redact) error {
 	wg := &sync.WaitGroup{}
 
 	// Error channel to capture errors from goroutines
 	errorCh := make(chan error, len(input))
+	limitCh := make(chan struct{}, MAX_CONCURRENT_REDACTORS)
```
emosbaugh, line 30: dont you have to close this channel?

pkg/redact/multi_line.go
```diff
@@ -66,33 +71,42 @@ func (r *MultiLineRedactor) Redact(input io.Reader, path string) io.Reader {
...
-			clean := r.re2.ReplaceAllString(line2, substStr)
+			clean := r.re2.ReplaceAll(line2, substStr)
 
-			// io.WriteString would be nicer, but reader strips new lines
-			fmt.Fprintf(writer, "%s\n%s\n", line1, clean)
+			_, err = writer.Write(append(line1, '\n'))
+			if err != nil {
+				return
+			}
+			_, err = writer.Write(append(clean, '\n'))
```
DexterYan: should we also change it to write twice here?

pkg/redact/single_line.go
```diff
@@ -42,50 +43,60 @@ func (r *SingleLineRedactor) Redact(input io.Reader, path string) io.Reader {
...
 				continue
 			}
 
-			clean := r.re.ReplaceAllString(line, substStr)
-
-			// io.WriteString would be nicer, but scanner strips new lines
-			fmt.Fprintf(writer, "%s\n", clean)
+			clean := r.re.ReplaceAll(line, substStr)
 
+			_, err = writer.Write(append(clean, '\n')) // Append newline since scanner strips it
```
DexterYan: also here, maybe need change to write twice

arcolife:
Thus far, confirming [smoke tests](https://github.com/replicatedhq/kots/actions/runs/6189935773/) ✅  for https://github.com/replicatedhq/kots/pull/4031 with [latest fixes](https://github.com/replicatedhq/kots/pull/4031/commits)

pkg/collect/redact.go
banjoh: Haha. Yer, need to

emosbaugh (approved):

arcolife (approved):

diamonwiggins (approved):

PR #1333, opened 2023-09-11
Fix incorrect result URI for pass and warn outcomes in common status analyzer

diamonwiggins:

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/pkg/analyze/common_status.go
+++ b/pkg/analyze/common_status.go
@@ -65,7 +65,7 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, name string, iconKey 
 			if exists == false && outcome.Warn.When != "absent" {
 				result.IsFail = true
 				result.Message = fmt.Sprintf("The %s %q was not found", resourceType, name)
-				result.URI = outcome.Fail.URI
+				result.URI = outcome.Warn.URI
 				return result, nil
 			}
 
--- a/pkg/analyze/common_status.go
+++ b/pkg/analyze/common_status.go
@@ -2,10 +2,11 @@ package analyzer
 
 import (
 	"fmt"
-	"github.com/pkg/errors"
-	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
 	"strconv"
 	"strings"
+
+	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


pkg/analyze/common_status.go
```diff
@@ -104,7 +105,7 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, name string, iconKey
 			if exists == false && outcome.Pass.When != "absent" {
 				result.IsFail = true
 				result.Message = fmt.Sprintf("The %s %q was not found", resourceType, name)
-				result.URI = outcome.Fail.URI
```
CpuID, line 107: is it worth doing an `if outcome.Pass != nil { result.URI = outcome.Pass.URI }` or something here? just to be super safe...? since we're talking about data coming back potentially as JSON content?

in theory it should always be populated the way we'd expect, but defensive programming and all :) especially since we've got a nil pointer reference we can't repro and all, which could be environment specific somehow

pkg/analyze/common_status.go
```diff
@@ -64,7 +65,7 @@ func commonStatus(outcomes []*troubleshootv1beta2.Outcome, name string, iconKey
 			if exists == false && outcome.Warn.When != "absent" {
 				result.IsFail = true
 				result.Message = fmt.Sprintf("The %s %q was not found", resourceType, name)
-				result.URI = outcome.Fail.URI
```
CpuID, line 67: see https://github.com/replicatedhq/troubleshoot/pull/1333/files#diff-d2f5edb6b9cc4705fc281663c6648c1792d04ad3c4e7e732167a96011dbca51cL107 similar thoughts but for `outcome.Warn`

CpuID (dismissed):
one suggestion possibly, but I think you're on the right track with this...

danj-replicated (approved):
yeh this looks correct

xavpaice (approved):

pkg/analyze/common_status.go
banjoh: @CpuID isn't this condition checked in line [103 above](https://github.com/replicatedhq/troubleshoot/blob/c079d96efedbdac9033547e56aff912910167a2b/pkg/analyze/common_status.go#L103)? Unless I'm missing something here.

PR #1338, opened 2023-09-12
fix: use duration strings for http collector timeout

banjoh:

This follows the same format that all other collectors use.

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/http_test.go
+++ b/pkg/collect/http_test.go
@@ -217,7 +217,7 @@ func TestCollectHTTP_Collect(t *testing.T) {
 					CollectorName: "example-com",
 				},
 				Get: &troubleshootv1beta2.Get{
-					Timeout: 200 * time.Nanosecond,
+					Timeout: "200ms",
 				},
 			},
 			args: args{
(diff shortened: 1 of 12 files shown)


DexterYan (dismissed):

pkg/collect/http_test.go
```diff
@@ -217,7 +217,7 @@ func TestCollectHTTP_Collect(t *testing.T) {
 					CollectorName: "example-com",
 				},
 				Get: &troubleshootv1beta2.Get{
-					Timeout: 200 * time.Nanosecond,
+					Timeout: "200ms",
```
arcolife: This was `200 * time.Nanosecond` earlier, so that should be "200ns" I believe. The purpose is to have as a very quick timeout and have it fail on purpose.

arcolife (changes requested):

pkg/collect/http_test.go
arcolife: the server delay is 200ms so this would become flaky if exactly equal to delay.

banjoh: Good catch. I'll update the test case.

arcolife (approved):
LGTM

DexterYan (approved):

diamonwiggins (approved):

PR #1376, opened 2023-10-19
feat: save host run file output

cwyl02:

Address the collector part of https://github.com/replicatedhq/troubleshoot/issues/1078

- added `.env` to allow user to specify which env vars to import into this host run
- added `.input` to allow user to create input files(e.g., config file, sample data for this host run). User needs to utilize `TS_INPUT_DIR` env var in their script, with the filename they specify in `.input` , e.g., `$TS_INPUT_DIR/dummy.yaml`
- added `.outputDir` to allow user to specify a name of a directory(within the directory of the collector in bundle) to host their host run command file output. Note that command run has no access to this directory

Example SupportBundle:
```yaml
kind: SupportBundle
metadata:
  name: run
spec:
  hostCollectors:
    - run:
        collectorName: "some-magic-on-host"
        command: "/home/cwyl02/custom-host-run.sh"
        args: ["--timeout", "10m"]
        env:
            - AWS_REGION=us-west-2
        ignoreParentEnvs: false  # Whether to inherit all the variable or not. If true, inheritEnvs would be made redundant
        inheritEnvs:   # list of env variables to inherit from the parent process
            - SHELL
            - LANG
            - PYTHON
        outputDir: "magic-outputs"
        input:
          dummy.yaml: |-
            username: postgres
            password: <my-pass>
            dbHost: <hostname>
            map:
              key: value
            list:
              - val1
              - val2
          config.toml: |-
             title = "TOML Example"
             [owner]
             name = "Tom Preston-Werner"
```

Result bundle:
```bash
├── analysis.json
├── execution-data
│   └── summary.txt
├── host-collectors
│   └── run-host
│       └── some-magic-on-host
│           ├── collector-info.json
│           ├── collector.txt
│           └── magic-outputs
│               ├── magic-dir-0
│               │   └── t1.txt
│               └── magic-dir-1
│                   └── t2.txt
└── version.yaml
```

## Checklist

- [x] Yes
- [ ] No


pkg/collect/host_run.go
```diff
@@ -31,20 +37,67 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse env variable")
+	}
+
+	// if we choose to save result for the command run
+	if runHostCollector.OutputDir != "" {
+		cmdOutputTempDir, err = os.MkdirTemp("", runHostCollector.OutputDir)
+		defer os.RemoveAll(cmdOutputTempDir)
+		if err != nil {
+			return nil, errors.New(fmt.Sprintf("failed to created temp dir for: %s", runHostCollector.OutputDir))
```
banjoh:
```suggestion
			return nil, errors.New(fmt.Sprintf("failed to create temp dir for: %s", runHostCollector.OutputDir))
```

pkg/collect/host_run.go
```diff
@@ -31,20 +37,67 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+		cmd.Env = append(cmd.Env,
+			fmt.Sprintf("TS_WORKSPACE_DIR=%s", cmdOutputTempDir),
+		)
+	}
+
+	if runHostCollector.Input != nil {
+		cmdInputTempDir, err = os.MkdirTemp("", "input")
+		defer os.RemoveAll(cmdInputTempDir)
+		if err != nil {
+			return nil, errors.New("failed to created temp dir for host run input")
```
banjoh:
```suggestion
			return nil, errors.New("failed to create temp dir for host run input")
```

pkg/collect/host_run.go
```diff
@@ -66,14 +116,67 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
+}
+
+func (c *CollectHostRun) processEnvVars(cmd *exec.Cmd) error {
+	runHostCollector := c.hostCollector
+
+	if runHostCollector.IgnoreParentEnvs {
+		klog.V(2).Info("Not inheriting the environment variables!")
+		if runHostCollector.InheritEnvs != nil {
+			klog.V(2).Infof("the following environment variables will not be loaded to the command: %",
+				strings.Join(runHostCollector.InheritEnvs, " "))
```
banjoh:
```suggestion
			klog.V(2).Infof("The following environment variables will not be loaded to the command: [%s]",
				strings.Join(runHostCollector.InheritEnvs, ", "))
```

pkg/collect/host_run.go
```diff
@@ -31,20 +37,67 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+	}
+
+	if runHostCollector.Input != nil {
+		cmdInputTempDir, err = os.MkdirTemp("", "input")
+		defer os.RemoveAll(cmdInputTempDir)
+		if err != nil {
+			return nil, errors.New("failed to created temp dir for host run input")
+		}
+		for inFilename := range runHostCollector.Input {
+			if strings.Contains(inFilename, "/") {
```
banjoh, line 94: It seems like you do not do anything with the input files besides checking that they have no `/`

examples/collect/host/run-and-save-output.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: HostCollector
+metadata:
+  name: run-host-cmd-and-save-output
+spec:
+  collectors:
+    - run:
+        collectorName: "my-custom-run"
+        command: "cat $TS_INPUT_DIR/dummy.yaml"
```
banjoh: This will also not work. It needs to be something like below

```yaml
        command: "cat"
        args: [ "$TS_INPUT_DIR/dummy.yaml" ]
```

examples/collect/host/run-and-save-output.yaml
```diff
@@ -0,0 +1,34 @@
...
+        args: [ "--timeout", "10m" ]
```
banjoh: This is unnecessary. `cat` will error

cwyl02: yes, gonna change it to somelike bash

pkg/collect/host_run.go
cwyl02: correct. i used to also collect the input file into the bundle but talking with @mhrabovcin we think that the input files can be sensitive to anyone that has access to the bundle while the bundle creator has the input anyways.

banjoh: Isn't the idea to store them in `TS_INPUT_DIR` then delete them once done? If I understood correctly what you want to avoid is having them end up in the support bundle, correct?

examples/collect/host/run-and-save-output.yaml
banjoh: This is more accurate. `$TS_INPUT_DIR` will be evaluated by the shell, not `cat`

```yaml
command: "sh"
args: [ "-c", "cat $TS_INPUT_DIR/dummy.yaml" ]
```

pkg/collect/host_run.go
cwyl02: introduced a commit that is not needed -- reverted that commit now.

examples/collect/host/run-and-save-output.yaml
cwyl02: done

examples/collect/host/run-and-save-output.yaml
cwyl02: done

pkg/collect/host_run.go
cwyl02: done

pkg/collect/host_run.go
cwyl02: done

pkg/collect/host_run.go
cwyl02: done

pkg/collect/host_run.go
```diff
@@ -31,20 +37,72 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+	}
+
+	if runHostCollector.Input != nil {
+		cmdInputTempDir, err = os.MkdirTemp("", "input")
+		defer os.RemoveAll(cmdInputTempDir)
+		if err != nil {
+			return nil, errors.New("failed to create temp dir for host run input")
+		}
+		for inFilename, inFileContent := range runHostCollector.Input {
+			if strings.Contains(inFileContent, "/") {
```
banjoh:
```suggestion
			if strings.Contains(inFilename, "/") {
```

pkg/collect/host_run.go
```diff
@@ -66,14 +121,67 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
 	}
 
-	output := NewResult()
 	output.SaveResult(c.BundlePath, resultInfo, bytes.NewBuffer(b))
 	output.SaveResult(c.BundlePath, result, bytes.NewBuffer(stdout.Bytes()))
+	// walkthrough the output directory and save result for each file
+	if runHostCollector.OutputDir != "" {
+		runInfo.OutputDir = runHostCollector.OutputDir
+		bundleOutputRelativePath = filepath.Join(collectorRelativePath, runHostCollector.OutputDir)
+		klog.V(2).Infof("saving run host command output to %s", runHostCollector.OutputDir)
```
banjoh:
```suggestion
		klog.V(2).Infof("Saving command output to %q in bundle", bundleOutputRelativePath)
```

pkg/collect/host_run.go
```diff
@@ -31,20 +37,72 @@ func (c *CollectHostRun) IsExcluded() (bool, error) {
...
+	}
 
 	cmd := exec.Command(runHostCollector.Command, runHostCollector.Args...)
 
-	var stdout, stderr bytes.Buffer
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
-
-	runInfo := HostRunInfo{
+	runInfo := &HostRunInfo{
```
banjoh, line 55:
```suggestion
	klog.V(2).Infof("Run host collector command: %q", cmd.String())
	runInfo := &HostRunInfo{
```

banjoh (commented):
I've run some tests and the changes LGTM now. I added some suggestions. Once those are accepted, I'll approve the PR

banjoh (approved):
LGTM

PR #1390, opened 2023-11-14
fix(collector): Let pgx library parse TLS parameters

### LINKED ISSUE #1163: PostgreSQL collector ignoring sslmode parameter in the connection string
**Bug Description**

When the [sslmode](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION) parameter is provided in the connection string, it indicates to the client what level of SSL verification should be done. `verify-full` for example verifies that the server certificate is signed using the CA cert provided and that the hostname making the connection matches the hostnames in the server cert (`Common Name` or `Subject Alternative Name`s). `verify-ca` on the other hand just checks that the server certificate is signed by the CA cert provided. There are other modes such as `disable` that tells the client to ignore any verification.

The collector below will fail to connect to the database if the server `hostname` is not in the server's certificate. This should not be the case since the connection string has `verify-ca` which should just verify that the certificate was signed by the CA.

```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - postgres:
        collectorName: pg
        uri: postgresql://user:password@hostname:5432/defaultdb?sslmode=verify-ca
        tls:
          cacert: |
            -----BEGIN CERTIFICATE-----
            ...
            <truncated>
            ...
            -----END CERTIFICATE-----
```

**Expected Behavior**

The collector should respect the `sslmode` parameter provided in the connection string an collect a bundle containing `{"isConnected":true,"version":"X.X"}` in `<bundle>/postgres/pg.json`.

**Steps To Reproduce**

* Generate TLS parameters
```sh
CONFIG="
[req]
distinguished_name=dn
[ dn ]
[ ext ]
basicConstraints=CA:TRUE,pathlen:0
"

openssl genrsa -out ca.key 2048
openssl req -config <(echo "$CONFIG") -new -x509 -days 365 -key ca.key -subj "/CN=Root CA" -out ca.crt
openssl req -newkey rsa:2048 -nodes -keyout server.key -subj "/CN=server" -out server.csr
openssl x509 -req -extfile <(printf "subjectAltName=DNS:example.com,IP:127.0.0.1") -days 365 -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt
chmod 600 server.key
```
* Start a postgres server exposing an SSL enabled port
```sh
docker run -d --rm --name postgres -p5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust -v "$(pwd)/server.crt:/var/lib/postgresql/server.crt:ro" -v "$(pwd)/server.key:/var/lib/postgresql/server.key:ro" -v "$(pwd)/ca.crt:/var/lib/postgresql/ca.crt:ro"  postgres:13-alpine -c ssl=on -c ssl_cert_file=/var/lib/postgresql/server.crt -c ssl_key_file=/var/lib/postgresql/server.key -c ssl_ca_file=/var/lib/postgresql/ca.crt
```
* Connect to the server using `psql`. This should succeed
```sh
psql -h localhost "sslmode=verify-ca sslrootcert=ca.crt hostaddr=127.0.0.1 port=5432 user=postgres"
```
* Collect a support bundle using the below spec. This will contain connection errors in `<bundle>/postgres/pg.json`
```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  collectors:
    - postgres:
        collectorName: pg
        uri: postgresql://postgres:postgres@localhost:5432/postgres?sslmode=verify-ca
        tls:
          cacert: |
            -----BEGIN CERTIFICATE-----
            <ca.crt>
            -----END CERTIFICATE-----

```

banjoh:

This allows the collector to respect the sslmode parameters

**Reviewer notes**
For the postgres collector, we let pgx library parse TLS parameters via its [ParseConfig](https://github.com/jackc/pgx/blob/df5d00eb608ce36f269009dba61607a0507a802e/pgconn/config.go#L166) function instead of the collector's default implementation of construction a [tls.Config instance for secure connections](https://github.com/replicatedhq/troubleshoot/blob/bc4856869e5a1c70b7742c0f06011ac6c1c20dd2/pkg/collect/util.go#L148). `pgx` [respects the sslmode](https://github.com/jackc/pgx/blob/df5d00eb608ce36f269009dba61607a0507a802e/pgconn/config.go#L628) parameters when constructing a tls.Config instance. Troubleshoot does not need to do this.

We drop the TLS param files to a file and pass them to the library as environment variables, then cleanup after ourselves.

Fix: #1163

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -120,19 +120,16 @@ func (c *CollectClusterResources) Collect(progressChan chan<- interface{}) (Coll
 	var namespaceNames []string
 	if len(c.Collector.Namespaces) > 0 {
 		namespaces, namespaceErrors := getNamespaces(ctx, client, c.Collector.Namespaces)
-		klog.V(4).Infof("checking for namespaces access: %s", string(namespaces))
 		namespaceNames = c.Collector.Namespaces
 		output.SaveResult(c.BundlePath, path.Join(constants.CLUSTER_RESOURCES_DIR, fmt.Sprintf("%s.json", constants.CLUSTER_RESOURCES_NAMESPACES)), bytes.NewBuffer(namespaces))
 		output.SaveResult(c.BundlePath, path.Join(constants.CLUSTER_RESOURCES_DIR, fmt.Sprintf("%s-errors.json", constants.CLUSTER_RESOURCES_NAMESPACES)), marshalErrors(namespaceErrors))
 	} else if c.Namespace != "" {
 		namespace, namespaceErrors := getNamespace(ctx, client, c.Namespace)
-		klog.V(4).Infof("checking for namespace access: %s", string(namespace))
 		output.SaveResult(c.BundlePath, path.Join(constants.CLUSTER_RESOURCES_DIR, fmt.Sprintf("%s.json", constants.CLUSTER_RESOURCES_NAMESPACES)), bytes.NewBuffer(namespace))
 		output.SaveResult(c.BundlePath, path.Join(constants.CLUSTER_RESOURCES_DIR, fmt.Sprintf("%s-errors.json", constants.CLUSTER_RESOURCES_NAMESPACES)), marshalErrors(namespaceErrors))
 		namespaceNames = append(namespaceNames, c.Namespace)
 	} else {
 		namespaces, namespaceList, namespaceErrors := getAllNamespaces(ctx, client)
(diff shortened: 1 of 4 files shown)


CpuID (approved):
LGTM, and tests pass, approved

PR #1418, opened 2024-01-02
fix (supportbundle): Add default collectors when expected

### LINKED ISSUE #1417: Misbehaviour when merging a support bundle spec with an empty or missing collectors list
**Bug Description**

Passing in a support bundle spec that does not have collectors ought to add the `clusterInfo` and `clusterResources` collectors by default. This is not happening at the moment.

```sh
cat <<EOF | support-bundle --dry-run -
kind: SupportBundle
metadata:
  name: simplest-support-bundle
spec:
  collectors: []
EOF
``` 

Running the command above results in the output below

```sh
kind: SupportBundle
metadata:
  creationTimestamp: null
  name: merged-support-bundle-spec
spec: {}
status: {}
```

Trying to collect a support bundle with the same spec ends up failing like so

```sh
$ cat <<EOF | support-bundle -
kind: SupportBundle
metadata:
  name: simplest-support-bundle
spec:
  collectors: []
EOF

Error: failed to run collect and analyze process: failed to generate support bundle
✘-1 ~/repos/replicated/troubleshoot [main {origin/main}|…7⚑ 9]
```

**Expected Behaviour**

The expected output out of the dry run command above should be
```sh
kind: SupportBundle
metadata:
  creationTimestamp: null
  name: merged-support-bundle-spec
spec:
  collectors:
  - clusterInfo: {}
  - clusterResources: {}
status: {}
```

Include a description of the expected behaviour here.

**Additional Context**

This regression is related to https://github.com/replicatedhq/troubleshoot/issues/1393

- Troubleshoot version - `v0.79.0`
- Operating system - Darwin

banjoh:

Ensure default collectors (`clusterInfo` and `clusterResources`) are added by default whenever required when running `support-bundle`. Below are specs representing possible combinations

- Defaults added
```yaml
kind: SupportBundle
spec:
  collectors: []
```

- Defaults added
```yaml
kind: SupportBundle
spec:
  collectors:
  - logs: {}
```

- Defaults added (similar to when `spec: {}` is present. `spec` is not mandatory. This depicts the minimum spec one can create)
```yaml
kind: SupportBundle
```

- Defaults **NOT** added
```yaml
kind: SupportBundle
spec:
  hostCollectors:
  - cpu: {}
```

- Defaults **NOT** added in the resulting merged spec
```yaml
kind: HostCollector
spec:
  collectors:
  - cpu: {}
```

- Defaults added
```yaml
kind: SupportBundle
spec:
  hostCollectors:
  - cpu: {}
  collectors:
  - logs: {}
```

- Defaults **NOT** added. `collectors: []` is treated as if its absent
```yaml
kind: SupportBundle
spec:
  hostCollectors:
  - cpu: {}
  collectors: []
```

Fixes: #1417

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -315,11 +315,12 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface) 
 		mainBundle.Spec.HostCollectors = util.Append(mainBundle.Spec.HostCollectors, hc.Spec.Collectors)
 	}
 
-	if mainBundle.Spec.Collectors != nil || mainBundle.Spec.Analyzers != nil {
-		// If we have in-cluster collectors, ensure cluster info and cluster resources
-		// collectors are in the merged spec. We need to add them here so when we --dry-run,
-		// these collectors are included. supportbundle.runCollectors duplicates this bit.
-		// We'll need to refactor it out later when its clearer what other code depends on this logic e.g KOTS
+	if !(len(mainBundle.Spec.HostCollectors) > 0 && len(mainBundle.Spec.Collectors) == 0) {
+		// Always add default collectors unless we only have host collectors
+		// We need to add them here so when we --dry-run, these collectors
+		// are included. supportbundle.runCollectors duplicates this bit.
+		// We'll need to refactor it out later when its clearer what other
+		// code depends on this logic e.g KOTS
 		mainBundle.Spec.Collectors = collect.EnsureCollectorInList(
 			mainBundle.Spec.Collectors,
 			troubleshootv1beta2.Collect{ClusterInfo: &troubleshootv1beta2.ClusterInfo{}},
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -315,7 +315,7 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface) 
 		mainBundle.Spec.HostCollectors = util.Append(mainBundle.Spec.HostCollectors, hc.Spec.Collectors)
 	}
 
-	if mainBundle.Spec.Collectors != nil {
+	if mainBundle.Spec.Collectors != nil || mainBundle.Spec.Analyzers != nil {
 		// If we have in-cluster collectors, ensure cluster info and cluster resources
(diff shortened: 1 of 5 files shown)


chris-sanders:
So this I expected

Defaults added:
```
kind: SupportBundle
spec:
  collectors: []
```

But this one I'm surprised by:
Defaults NOT added:
```
kind: SupportBundle
spec:
  hostCollectors:
  - cpu: {}
  collectors: []
```

Is that latter an intentional choice or a side effect of how the object is created. I would prefer these both add defaults, it looks like they would.


Wait this doesn't have defaults either?
```
kind: SupportBundle
spec:
  hostCollectors:
  - cpu: {}
```

So is there an intentional rule, if hostCollectors are added don't add default collectors? Is that the logic?

diamonwiggins:
> So is there an intentional rule, if hostCollectors are added don't add default collectors? Is that the logic?

I think this is what we did in the past to avoid any code paths that would try contacting a Kubernetes cluster if there were only host collectors/analyzers in the spec. As in we figured it was intentional that's the behavior you wanted it you didn't at least include an empty collector/analyzer property.

DexterYan:
According the test cases, I saw
- Defaults added
```
kind: SupportBundle
spec:
  analyzers: []
```

just to confirm it will not have Defaults as follow, right?
```
kind: SupportBundle
spec:
  hostCollectors:
  - cpu: {}
  analyzers: []
```

banjoh:
> kind: SupportBundle
> spec:
>   hostCollectors:
>   - cpu: {}
>   analyzers: []

Correct. Defaults will not be added. `analyzers: []` is also treated the same as being absent

banjoh:
> But this one I'm surprised by: Defaults NOT added:
> 
> ```
> kind: SupportBundle
> spec:
>   hostCollectors:
>   - cpu: {}
>   collectors: []
> ```
> 
> Is that latter an intentional choice or a side effect of how the object is created.

Yes, its intentional. Its code that was added a while back as commented in https://github.com/replicatedhq/troubleshoot/pull/1418#issuecomment-1877840777

banjoh:
The logic in troubleshoot treats zero length lists the same as being absent for `collectors`, `analyzers`, and `redactors`

xavpaice (approved):
I read this as a bugfix, making things behave as we originally intended.

PR #1435, opened 2024-01-27
feat: add timeout to run host collector

### LINKED ISSUE #1377: Add a timeout to the run host collector

The [run host collector](https://troubleshoot.sh/docs/host-collect-analyze/run/) us used to run arbitrary command/programs which the collector has little control of with regards to how long the process takes. This can lead to the collector running for too long and in some cases, not even stopping

Collecting a support bundle using the spec below will never stop

```yaml
kind: SupportBundle
metadata:
  name: run
spec:
  hostCollectors:
    - run:
        name: "ping-google"
        command: "ping"
        args:
          - "google.com"
```

**Describe the feature**

Add a `timeout` field that can be used to limit how long the collector should run. The timeout should be
- A `string` [parsable into a time.Duration](https://pkg.go.dev/time#ParseDuration)
- Non negative. If its a negative value, default to passing no timeout

Here is an example spec

```yaml
    - run:
        name: "ping-google"
        timeout: 
        command: "ping"
        args:
          - "google.com"
```

Leave things as they are 

**Additional context**

- The run pod collector has a [timeout field](https://troubleshoot.sh/docs/collect/run-pod/#timeout-optional) which behaves as described above. This can be "borrowed"
- Eventually, [exec.CommandContext](https://pkg.go.dev/os/exec#CommandContext) would end up being called if there is a timeout

nvanthao:

#1377

## Checklist

- [ ] Yes
- [X] No

### DIFF
--- a/pkg/collect/host_run.go
+++ b/pkg/collect/host_run.go
@@ -143,7 +143,7 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
 			runInfo.ExitCode = strings.TrimPrefix(werr.Error(), "exit status ")
 			runInfo.Error = stderr.String()
 		} else if err == context.DeadlineExceeded {
-			runInfo.ExitCode = "0"
+			runInfo.ExitCode = "-1"
 			runInfo.Error = fmt.Sprintf("command timed out after %s", timeout.String())
 		} else {
 			return nil, errors.Wrap(err, "failed to run")
--- a/pkg/collect/host_run.go
+++ b/pkg/collect/host_run.go
@@ -2,12 +2,14 @@ package collect
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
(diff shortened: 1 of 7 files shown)


pkg/collect/host_run.go
```diff
@@ -49,7 +51,23 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
...
+	var (
+		timeout            time.Duration
+		cmd                *exec.Cmd
+		errInvalidDuration error
+	)
+
+	ctx := context.Background()
+	cmdPath := c.attemptToConvertCmdToAbsPath()
+
+	timeout, errInvalidDuration = time.ParseDuration(runHostCollector.Timeout)
```
banjoh: If the format is wrong, return an error like we do elsewhere

https://github.com/search?q=repo%3Areplicatedhq%2Ftroubleshoot%20%20time.ParseDuration&type=code

pkg/collect/host_run.go
```diff
@@ -118,6 +136,9 @@ func (c *CollectHostRun) Collect(progressChan chan<- interface{}) (map[string][]
 		if werr, ok := err.(*exec.ExitError); ok {
 			runInfo.ExitCode = strings.TrimPrefix(werr.Error(), "exit status ")
 			runInfo.Error = stderr.String()
+		} else if err == context.DeadlineExceeded {
+			runInfo.ExitCode = "0"
```
banjoh: Should't we use `werr.ExitCode()`? I'd prefer not "swallowing" the error code.

banjoh (commented):
Changes look good. I added some minor comments

pkg/collect/host_run.go
nvanthao: thanks @banjoh! this is done!

pkg/collect/host_run.go
nvanthao: There is no Exit Code as as the error is about context becomes done prior to the command is run. Do you think we should update the artificial Exit Code to something else? E.g. `124`?

banjoh: The value will be `-1` which the [std library uses to indicate](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/os/exec_posix.go;l=133) that the process did not complete. Any other value can be misinterpreted as a valid return code from the process.

banjoh:
When you create a [PR to update the docs](https://github.com/replicatedhq/troubleshoot.sh/pulls), could you please add a one liner to indicate the `-1` exit code as an error possibly caused by timeouts?

pkg/collect/host_run.go
nvanthao: many thanks @banjoh this is done.

banjoh (approved):

PR #1473, opened 2024-02-18
feat: add dependabot auto merge

xavpaice:
Adds auto-merge of Dependabot PRs when they pass all tests.

### DIFF
--- a/.github/workflows/automated-prs-manager.yaml
+++ b/.github/workflows/automated-prs-manager.yaml
@@ -11,7 +11,7 @@ jobs:
     outputs:
       prs: ${{ steps.list-prs.outputs.prs }}
     env:
-      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GH_TOKEN: ${{ secrets.REPLICATED_GH_PAT }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -50,7 +50,7 @@ jobs:
       fail-fast: false
       max-parallel: 1
     env:
-      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GH_TOKEN: ${{ secrets.REPLICATED_GH_PAT }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/automated-prs-manager.yaml
+++ b/.github/workflows/automated-prs-manager.yaml
@@ -93,4 +93,4 @@ jobs:
             exit 0
           fi
 
-          echo "Less than half of the validate-* jobs are successful. Skipping."
+          echo "Less than half of the jobs are successful. Skipping."
--- a/.github/workflows/automated-prs-manager.yaml
+++ b/.github/workflows/automated-prs-manager.yaml
@@ -11,7 +11,7 @@ jobs:
     outputs:
       prs: ${{ steps.list-prs.outputs.prs }}
     env:
-      GH_TOKEN: ${{ secrets.REPLICATED_GH_PAT }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -50,7 +50,7 @@ jobs:
       fail-fast: false
       max-parallel: 1
     env:
-      GH_TOKEN: ${{ secrets.REPLICATED_GH_PAT }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/automated-prs-manager.yaml
+++ b/.github/workflows/automated-prs-manager.yaml
@@ -68,26 +68,27 @@ jobs:
           fi
 
           # If all checks passed, approve and merge.
-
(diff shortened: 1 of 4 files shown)


.github/workflows/automated-prs-manager.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+          num_of_pending_jobs=$(gh run view "$run_id" --json jobs -q '.jobs[] | select(.conclusion == "") | .name' | wc -l)
+          if [ "$num_of_pending_jobs" -gt 0 ]; then
+            echo "There are still pending jobs. Skipping."
+            exit 0
+          fi
+
+          # If all tests and required checks passed, approve and merge.
+
+          if gh run view "$run_id" --json jobs -q '.jobs[] | select(.name == "validate-success") | .conclusion' | grep -q "success"; then
```
sgalsaleh, line 71: I don't see a `validate-success` job in the PR testing workflow.

.github/workflows/automated-prs-manager.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+          num_of_pending_jobs=$(gh run view "$run_id" --json jobs -q '.jobs[] | select(.conclusion == "") | .name' | wc -l)
+          if [ "$num_of_pending_jobs" -gt 0 ]; then
+            echo "There are still pending jobs. Skipping."
+            exit 0
+          fi
+
+          # If all checks passed, approve and merge.
+
+          if gh pr checks "${{ matrix.pr.url }}" ; then
```
sgalsaleh: i believe i encountered a race condition when trying to only validate the pr checks this way. had to validate a specific job (in KOTS's case it's `validate-success` which is also a branch protection rule). i'd recommend adding one.

.github/workflows/automated-prs-manager.yaml
```diff
@@ -0,0 +1,95 @@
...
+            gh pr review --approve "${{ matrix.pr.url }}" --body-file body.txt
+            gh pr merge --auto --squash "${{ matrix.pr.url }}"
+            sleep 10
+            exit 0
+          else
+            echo "Some checks did not pass. Skipping."
+            exit 0
+          fi
+
+          # If more than half of the validate-* jobs are successful, re-run the failed jobs.
```
sgalsaleh: logic below only checks jobs that start with `validate-`, in KOTS's case that's sufficient, but i've seen other jobs in troubleshoot that don't have that prefix if i'm not mistaken.

xavpaice: I'm in two minds if we even need this retry logic.  There's 5 validate-* jobs and if they're OK, then the builds have to be OK, so retries would likely only catch the labels job.

.github/workflows/automated-prs-manager.yaml
```diff
@@ -68,26 +68,27 @@ jobs:
...
-            echo -e "LGTM :thumbsup: \n\nThis PR was automatically approved and merged by the [automated-prs-manager](https://github.com/replicatedhq/troubleshoot/blob/main/.github/workflows/automated-prs-manager.yaml) GitHub action" > body.txt
-            gh pr review --approve "${{ matrix.pr.url }}" --body-file body.txt
-            gh pr merge --auto --squash "${{ matrix.pr.url }}"
-            sleep 10
-            exit 0
-          else
-            echo "Some checks did not pass. Skipping."
-            exit 0
+          if gh run view "$run_id" --json jobs -q '.jobs[] | select(.name == "validate-success") | .conclusion' | grep -q "success"; then
+            if gh pr checks "${{ matrix.pr.url }}" ; then
```
DexterYan, line 72: if gh pr checks "${{ matrix.pr.url }}" --required; then

maybe add --required to have less check, and only dependents on the GitHub PR Checks are labeled with required

xavpaice: I don't want to auto merge anything that fails a test - it'll mean all future PRs fail the same test

.github/workflows/automated-prs-manager.yaml
```diff
@@ -0,0 +1,96 @@
...
+          num_of_jobs=$(gh run view "$run_id" --json jobs -q '.jobs[].name ' | wc -l)
+          num_of_successful_jobs=$(gh run view "$run_id" --json jobs -q '.jobs[] | select (.conclusion == "success")) |.name ' | wc -l)
+
+          if [ "$num_of_successful_jobs" -gt $((num_of_jobs / 2)) ]; then
+            echo "More than half of the jobs are successful. Re-running failed jobs."
+            gh run rerun "$run_id" --failed
+            exit 0
+          fi
+
+          echo "Less than half of the validate-* jobs are successful. Skipping."
```
sgalsaleh:
```suggestion
          echo "Less than half of the jobs are successful. Skipping."
```

.github/workflows/automated-prs-manager.yaml
sgalsaleh: Yeah, you might not need a retry logic. It was necessary in KOTS because there are more than 100 tests and flakiness is common.

sgalsaleh (dismissed):

sgalsaleh (approved):

DexterYan (approved):

PR #1474, opened 2024-02-19
New Event Analyzer

### LINKED ISSUE #911: Kubernetes Events Analyzer

Often times there are key indicators of issues within Kubernetes events. These events surface issues as it relates to scheduling, lack of resources, readiness timeouts. We should have an Events analyzer where you can specify one or all namespaces, match for text and evaluate it in a Warn/Fail/Pass condition similar to `textAnalyze` - https://troubleshoot.sh/docs/analyze/regex/

**Describe the feature**
```
kind: SupportBundle
metadata:
  name: bundle
spec:
  collectors:
    - clusterResources: {}
  analyzers:
    - eventsAnalyze:
        checkName: "insufficient-cpu"
        namespace: my-app
        regex: '.*Insufficient cpu.*'
        outcomes:
          - pass:
              when: "false"
              message: "Sufficient CPU for all Pods in my-app namespace"
          - fail:
              when: "true"
              message: "Insufficient CPU for some Pods in my-app namespace"
```

**Additional context**

nvanthao:

https://github.com/replicatedhq/troubleshoot/issues/911

This analyzer will search for all events under given `namespace` (if not provided, use `default` namespace)
It then will match the event by `reason`, `kind` (all case-insensitive) and `regex` for content of event `Message`

Sample preflight spec

```
kind: Preflight
metadata:
  name: preflight-sample-event
spec:
  collectors:
    - cluster-resources: {}
  analyzers:
    - event:
        checkName: disk capacity
        namespace: default
        reason: "InvalidDiskCapacity"
        regex: "invalid capacity.*"
        outcomes:
          - fail:
              when: "true"
              message: node does not have enough storage
          - pass:
              when: "false"
              message: we good
    - event:
        checkName: OOMkilled detector
        namespace: default
        reason: "OOMKilled"
        kind: Pod
        outcomes:
          - fail:
              when: "true"
              message: OOMKilled detected 
          - pass:
              when: "false" 
              message: we good

```

## Checklist

- [ ] Yes
- [x] No


pkg/analyze/event.go
```diff
@@ -0,0 +1,196 @@
...
+	err := json.Unmarshal(data, &eventList)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert []byte to corev1.EventList: %w", err)
+	}
+	return &eventList, nil
+}
+
+func getEvent(events *corev1.EventList, filter eventFilter) *corev1.Event {
+	for _, event := range events.Items {
+		if !matchReason(event.Reason, filter.reason) {
```
DexterYan, line 108: It looks like that `Reason` is must have field. It filter reason first. 
One question, is that possible that user just need to add `regex` without reason? Like Diamon's example
https://github.com/replicatedhq/troubleshoot/issues/911#issue-1498626521

nvanthao: I am thinking of 2 approaches:

1. User must know the specific `Reason` and the `regex` is only for `Message`. We can document this https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/events/event.go
2. User only need to provide `regex` and the `regex` will be matched on both `Reason` and `Message`. I am concern this will have performance impact as the current implementation we are iterating over all Events.

Please let me know your thought and I will update the code accordingly.

nvanthao: I have updated the code and we can use `regex` for both Reason and Message. The `reason` field is also optional now. Please let me know how it goes.

pkg/analyze/event.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	return strings.EqualFold(actual, expected)
+}
+
+func matchMessage(actual, expectedRegex string) bool {
+	// message Regex is optional
+	if expectedRegex == "" {
+		return true
+	}
+	re, err := regexp.Compile(expectedRegex)
```
banjoh: Minor: I'd compile this once before iterating over the events

nvanthao: many thanks @banjoh! that's a great catch! I have updated the code.

banjoh (approved):

PR #1480, opened 2024-02-20
remove Helm SDK example

xavpaice:

Removes the Helm SDK example which is a Go application and would need ongoing maintenance.  It is not a fully supported option and we do not require the example here.

## Checklist

- [ ] Yes
- [ ] No


DexterYan:
maybe also need to remove run test in https://github.com/replicatedhq/troubleshoot/blob/main/test/run-examples.sh

xavpaice:
given this is used in testing, which is valid, I'll leave it here and update it.

PR #1513, opened 2024-03-26
feat: Prompt for privileged user if host collectors present in spec

banjoh:

If host collectors are present in a spec when running `support-bundle` or `preflight` CLIs, we want to prompt a user to run the binary as an elevated user. This is because some of the tools executed require elevated privileges. Without elevated privileges, the collected data ends up being useless.

**Prompt to run as root**
```sh
support-bundle https://raw.githubusercontent.com/replicatedhq/troubleshoot-specs/main/host/default.yaml

Some host collectors may require elevated privileges to run.
Do you want to exit and rerun the command as a privileged user? (yes/no): yes
Exiting...
```

When the user declines (no), the process continues as normal

**Additional context**

We collectively check the presence/absence of host collectors because we cannot confidently tell which collectors require elevated privileges or don't. Both [run](https://troubleshoot.sh/docs/host-collect-analyze/run/) and [copy](https://troubleshoot.sh/docs/host-collect-analyze/copy/) collectors for example can wrap any command/path that requires root access. Our [default host collector spec](https://github.com/replicatedhq/troubleshoot-specs/blob/main/host/default.yaml) uses these collectors extensively.

## Checklist

- [ ] Yes
- [x] No


diamonwiggins (approved):
lgtm

PR #1520, opened 2024-04-02
feat: node metrics analyser

### LINKED ISSUE #1496: Collector for PVC disk usage

Troubleshoot collects PVC specs but not disk usage.  

**Describe the feature**

K8s users can use a script like `kubedf` [available here](https://gist.github.com/redmcg/60cfff7bca6f32969188008ad4a44c9a) which calls the `/api/v1/nodes` API and collect capacity bytes, available bytes, and percent used.  This algorithm would port cleanly to go for implementation as a collector, maybe call it "pvcDiskUsage"?

I imagine it would take optional namespace and optional pvc name (default=all).  Note that not everyone knows all their PVC names ahead of time, sometimes they're dynamically created.

- This can be done **sometimes** using the exec collector to shell into a pod which mounts the volume and run a `df` in that pod.  However, pods which are "from scratch", et al, do not contain `df` so that's not always an option.
- We could assemble a custom image containing `kubedf`, `jq`, and `kubectl` and run that with `runPod`.  It would be better if it was builtin to troubleshoot.
- I looked for a metrics API that would let the `http` collector pull it.  That would be ideal also, but I didn't see one.
- We could scrape this ourself at the app level and log it. Again, this is something many people want, probably and would be better if not app level.

**Additional context**

Our users create PVCs dynamically and when they fill up, it's a source of errors. A support bundle containing utilization metrics would be ideal.

banjoh:

Node metrics analyser which can be used to analyse PVC usage stats like below. Analysis of other stats in the collected can be added on a need to have basis.

```yaml
kind: Analyzer
spec:
  analyzers:
    - nodeMetrics:
        checkName: Check minio pvc space usage is less than 80%
        filters:
          pvc:
            nameRegex: "minio-data-ha-minio.*"
            namespace: "minio"
        outcomes:
          - fail:
              when: "pvcUsedPercentage >= 80"
              message: "There are PVCs using more than 80% of storage: {{ .ConcatenatedPVCNames}}"
          - pass:
              message: "No PVCs are using more than 80% of storage"
```

Running example
```sh
$ analyze support-bundle-2024-04-01T11_02_48.tar.gz --analyzers spec.yaml -v=1
Fail: Check minio pvc space usage is less than 80%
 There are PVCs using more than 80% of storage: minio/minio-data-ha-minio-5, minio/minio-data-ha-minio-1
```

Fixes: https://github.com/replicatedhq/troubleshoot/issues/1496

## Checklist

- [ ] Yes
- [x] No


pkg/collect/k8s_node_metrics.go
```diff
@@ -0,0 +1,102 @@
...
+	// Find nodes by label selector
+	if c.Collector.Selector != nil {
+		nodes, err := c.Client.CoreV1().Nodes().List(c.Context, metav1.ListOptions{
+			LabelSelector: strings.Join(c.Collector.Selector, ","),
+		})
+		if err != nil {
+			klog.Errorf("failed to list nodes by label selector: %v", err)
+		}
+		for _, node := range nodes.Items {
+			nodesMap[node.Name] = fmt.Sprintf(summaryUrlTemplate, node.Name)
```
DexterYan: If we have NodeNames and Selector, they could be a potential overlapping that nodeName is same as node.Name. But it should not be a problem. Just for a notice. LGTM

DexterYan (dismissed):

DexterYan (dismissed):

DexterYan (dismissed):

pkg/collect/k8s_node_metrics.go
banjoh: Yes, this is possible, and was intentional. This is why I used a map to deduplicate. You can have a selector that includes a subset of the list of nodes in the `NodeNames` list. These lists will be merged together.

DexterYan (dismissed):

DexterYan (dismissed):

DexterYan (approved):

PR #1534, opened 2024-04-28
feat: [sc-103754] Be able to detect search domain misconfiguration #1391

### LINKED ISSUE #1391: Be able to detect search domain misconfiguration
In cases where a search domain exists and resolves to a wildcard dns record (i.e: the search domain example.com resolves to *.example.com) kubernetes will inadvertently resolve in-cluster records via the search record. i.e: test.svc.cluster.local.example.com

**Describe the feature**
to better detect this misconfiguration we can compare a manual dns query inside the cluster to what the kubernetes API asserts as the cluster IP of a service. to ensure portability we can use the kubernetes service in the default namespace, as it's guaranteed to exist.

we can compare the outputs of:
`kubectl get svc kubernetes -o jsonpath='{.spec.clusterIP}'`
and
`dig +short kubernetes.default.svc.cluster.local` run from inside a pod

there are three possible states for the output:

1) the outputs match, and it's likely that DNS is configured correctly
2) dig returns NXDOMAIN, and it's likely that coredns is down or misconfigured
3) the outputs differ, and it's likely that the search domains are interfering with cluster name resolution

It might be necessary to write a custom analyzer for this, as it would require comparing data from both the cluster-resources collector and a custom run-pod.

It might be possible to chain together existing collectors to output a file in a format that a text analyser can compare.

nvanthao:
Close https://github.com/replicatedhq/troubleshoot/issues/1391

This PR automates some of the steps in Kubernetes document on troubleshooting DNS 
https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/

I originally plan to use `busybox` as the test pod but `nslookup` does not run correctly in that image. Hence I used this image `registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3` from Kubernetes document.


danj-replicated:
I really like this, but aren't we already collecting the coredns config as a configmap?

nvanthao:
Indeed we already collecting coredns config map in support bundle. This is just for convenience.

Currently we have to do few grep and jq filter to extract `Corefile` in `cluster-resources/configmaps/kube-system.json`

Other data such as endpoints is also in support bundle but not ready to be extracted from `sbctl`. E.g.

```
kubectl describe endpoints kube-dns -n kube-system
Error from server (InternalError): an error on the server ("") has prevented the request from succeeding (get endpoints kube-dns)
```

danj-replicated:
it looks like you're still tweaking a few things. lmk when this is ready for review and I'll give it a proper look over

nvanthao:
many thanks @danj-replicated! The code is ready for review, I've just rebased to main.

danj-replicated (approved):

PR #1541, opened 2024-05-02
feat: [sc-103119] URI is not replacing the spec but merging instead

### LINKED ISSUE #1486: URI is not replacing the spec but merging instead
**Bug Description**

When I use the `spec.uri` of a support bundle the [documentation](https://troubleshoot.sh/docs/support-bundle/supportbundle/#uri) states the the version pointed by `spec.uri` replaces the spec but it is actually merging the two specs.

**Expected Behavior**

The version pointed by `spec.uri` replaces the version being applied.

**Steps To Reproduce**

- Create a support bundle spec that points to a different one through its URI property. 
- Run with the `--dry-run` and inspect the merged result.

nvanthao:

Fixes: https://github.com/replicatedhq/troubleshoot/issues/1486
Fixes: https://github.com/replicatedhq/troubleshoot/pull/1537

Demo: https://asciinema.org/a/657334

## Checklist

- [X] Yes
- [ ] No

### DIFF
--- a/cmd/troubleshoot/cli/run_test.go
+++ b/cmd/troubleshoot/cli/run_test.go
@@ -61,6 +61,70 @@ spec:
 	assert.NotNil(t, kinds.SupportBundlesV1Beta2[0].Spec.Collectors[0].ClusterInfo)
 }
 
+func Test_loadMultipleSupportBundleSpecsWithNoURIs(t *testing.T) {
+	ctx := context.Background()
+	client := testclient.NewSimpleClientset()
+	specs := []string{testutils.ServeFromFilePath(t, `
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: sb-1
+spec:
+  collectors:
+    - clusterInfo:{}`), testutils.ServeFromFilePath(t, `
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
+  name: sb-2
+  spec:
+    collectors:
+      - clusterInfo: {}`)}
+
+	sb, _, err := loadSpecs(ctx, specs, client)
+	require.NoError(t, err)
+	require.Len(t, sb.Spec.Collectors, 2)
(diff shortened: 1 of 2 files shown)


cmd/troubleshoot/cli/run_test.go
```diff
@@ -57,8 +57,72 @@ spec:
...
+	defer srv.Close()
+
+	orig := strings.ReplaceAll(orig, "$MY_URI", srv.URL)
+	specUri := testutils.ServeFromFilePath(t, orig)
+	specs := []string{specFile, specUri}
+
+	sb, _, err := loadSpecs(ctx, specs, client)
+	require.NoError(t, err)
+	assert.NotNil(t, sb.Spec.Collectors[0].Logs)
+	assert.Nil(t, sb.Spec.Collectors[1].ConfigMap)
```
banjoh: I believe this ought to be `assert.Nil(t, sb.Spec.Collectors[1].ClusterInfo)`, if replacing the uri into `orig` in line 118 worked. But it won't work because `"$MY_URI"` does not exist in your template spec (`orig`) because it was already templated out by `Test_loadSupportBundleSpecsFromURIs` test in line 48. `var orig` is a global variable :(

What you want to do is extract `var orig..` into a separate function and use it in your tests like below

```go
func templSpec() string {
	return `
kind: SupportBundle
metadata:
  name: sb-1
spec:
  uri: $MY_URI
  collectors:
    - configMap:
        name: kube-root-ca.crt
        namespace: default
`
}
```

Then use it in your tests like so

```go
orig := strings.ReplaceAll(templSpec(), "$MY_URI", srv.URL)
```

banjoh (changes requested):

cmd/troubleshoot/cli/run_test.go
nvanthao: many thanks @banjoh! I've updated the tests.

banjoh (approved):

PR #1542, opened 2024-05-03
bug: [sc-103753] Host analysers are not deduplicating during multiple spec merges #1485

### LINKED ISSUE #1485: Host analysers are not deduplicating during multiple spec merges
**Bug Description**

If I have the same host analyser defined in two different specs they end up showing twice when I run `kubectl support-bundle spec0 spec1`. The same does not happen with a regular analyser, only with host analysers. It seems like the same deduplication applied for analysers is not applied for host analysers.

**Expected Behavior**

My duplicated host analyser to show up only once.

**Steps To Reproduce**

Create two specs with the same host analyser:

```yaml
kind: SupportBundle
metadata:
  name: embedded-cluster-host-support-bundle
spec:
  hostCollectors:
  - memory: {}
  hostAnalyzers:
  - memory:
      checkName: Amount of Memory
      outcomes:
      - warn:
          when: "< 2G"
          message: At least 2G of memory is recommended
      - pass:
          message: The system has at least 2G of memory
```

Then run the support-bundle passing both specs:

```
kubectl support-bundle spec0.yaml spec1.yaml
```

With the `--dry-run` flag I can see that the same host analyser is duplicated in the merged result.

nvanthao:

Fixes: https://github.com/replicatedhq/troubleshoot/issues/1485

Given spec `sb.yaml`

```yaml
kind: SupportBundle
metadata:
  name: sb
spec:
  analyzers:
  - clusterVersion: {}
  hostCollectors:
  - cpu: {}
  hostAnalyzers:
  - cpu: {}
```

and command

```
support-bundle --interactive=false sb.yaml sb.yaml --dry-run
```

Before

```yaml
kind: SupportBundle
metadata:
  creationTimestamp: null
  name: merged-support-bundle-spec
spec:
  analyzers:
  - clusterVersion:
      outcomes: null
  - clusterVersion:
      outcomes: null
  hostAnalyzers:
  - cpu:
      outcomes: null
  - cpu:
      outcomes: null
  hostCollectors:
  - cpu: {}
  - cpu: {}
status: {}

```

After

```yaml
kind: SupportBundle
metadata:
  creationTimestamp: null
  name: merged-support-bundle-spec
spec:
  analyzers:
  - clusterVersion:
      outcomes: null
  hostAnalyzers:
  - cpu:
      outcomes: null
  hostCollectors:
  - cpu: {}
status: {}
```

## Checklist

- [X] Yes
- [ ] No

Duplicated specs with `Analyzer`, `HostCollector` and `HostAnalayzer` will now be merged by their stringified content

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -363,16 +363,10 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface) 
 	}
 
 	// dedupe specs
-	if len(mainBundle.Spec.Analyzers) > 0 {
-		mainBundle.Spec.Analyzers = analyzer.DedupAnalyzers(mainBundle.Spec.Analyzers)
-	}
-
-	if len(mainBundle.Spec.HostCollectors) > 0 {
-		mainBundle.Spec.HostCollectors = collect.DedupHostCollectors(mainBundle.Spec.HostCollectors)
-	}
-	if len(mainBundle.Spec.HostAnalyzers) > 0 {
-		mainBundle.Spec.HostAnalyzers = analyzer.DedupHostAnalyzers(mainBundle.Spec.HostAnalyzers)
-	}
+	mainBundle.Spec.Collectors = util.Dedup(mainBundle.Spec.Collectors)
+	mainBundle.Spec.Analyzers = util.Dedup(mainBundle.Spec.Analyzers)
+	mainBundle.Spec.HostCollectors = util.Dedup(mainBundle.Spec.HostCollectors)
+	mainBundle.Spec.HostAnalyzers = util.Dedup(mainBundle.Spec.HostAnalyzers)
 
 	return mainBundle, additionalRedactors, nil
 }
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -363,6 +363,10 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface) 
 	}
 
 	// dedupe specs
+	if len(mainBundle.Spec.Analyzers) > 0 {
+		mainBundle.Spec.Analyzers = analyzer.DedupAnalyzers(mainBundle.Spec.Analyzers)
+	}
+
 	if len(mainBundle.Spec.HostCollectors) > 0 {
 		mainBundle.Spec.HostCollectors = collect.DedupHostCollectors(mainBundle.Spec.HostCollectors)
 	}
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -362,6 +362,14 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface) 
 		additionalRedactors.Spec.Redactors = util.Append(additionalRedactors.Spec.Redactors, r.Spec.Redactors)
 	}
 
+	// dedupe specs
+	if len(mainBundle.Spec.HostCollectors) > 0 {
+		mainBundle.Spec.HostCollectors = collect.DedupHostCollectors(mainBundle.Spec.HostCollectors)
+	}
+	if len(mainBundle.Spec.HostAnalyzers) > 0 {
+		mainBundle.Spec.HostAnalyzers = analyzer.DedupHostAnalyzers(mainBundle.Spec.HostAnalyzers)
+	}
+
 	return mainBundle, additionalRedactors, nil
 }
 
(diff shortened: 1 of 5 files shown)


xavpaice:
quick question - how does this interact if there's a url in one of the specs?  Just want to confirm that the dedup happens _after_ the url content replaced that particular spec.

nvanthao:
the dedupe happens after loading URI spec here
https://github.com/replicatedhq/troubleshoot/blob/f7ac39f1cb8eedae250e5cf3f81bdf86c0287172/cmd/troubleshoot/cli/run.go#L297

cmd/troubleshoot/cli/run.go
```diff
@@ -362,6 +362,18 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface)
 		additionalRedactors.Spec.Redactors = util.Append(additionalRedactors.Spec.Redactors, r.Spec.Redactors)
 	}
 
+	// dedupe specs
+	if len(mainBundle.Spec.Analyzers) > 0 {
```
DexterYan: I think we also need remove duplicates in collectors

```
	if len(mainBundle.Spec.Collectors) > 0 {
		mainBundle.Spec.Collectors = collect.DedupCollectors(mainBundle.Spec.Collectors)
	}
```

this sb.yaml will have duplicated collectors
```
kind: SupportBundle
metadata:
  name: sb
spec:
  collectors:
  - clusterResources: {}
  analyzers:
  - clusterVersion: {}
  hostCollectors:
  - cpu: {}
  hostAnalyzers:
  - cpu: {}

```

DexterYan:
```
kind: SupportBundle
metadata:
  creationTimestamp: null
  name: merged-support-bundle-spec
spec:
  analyzers:
  - clusterVersion:
      outcomes: null
  collectors:
  - clusterResources: {}
  - clusterResources: {}
  - clusterInfo: {}
  hostAnalyzers:
  - cpu:
      outcomes: null
  hostCollectors:
  - cpu: {}
status: {}
```

pkg/collect/collect.go
```diff
@@ -192,3 +192,22 @@ func CollectRemote(c *troubleshootv1beta2.RemoteCollector, additionalRedactors *
 	collectResult.AllCollectedData = allCollectedData
 	return collectResult, nil
 }
+
+func DedupHostCollectors(allCollectors []*troubleshootv1beta2.HostCollect) []*troubleshootv1beta2.HostCollect {
```
banjoh: I would DRY this code. `DedupHostCollectors` and `DedupHostAnalyzers` are very similar and can be one function. Think go Generics

nvanthao: Thanks @banjoh!

The reason I don't why a generic function is that there're existing dedupe functions

https://github.com/replicatedhq/troubleshoot/blob/6b368f22213786d25bf71172d587325c19fbc7ec/pkg/collect/collector.go#L266

https://github.com/replicatedhq/troubleshoot/blob/db871e6889f69856be3084230569d1e558e36613/pkg/analyze/analyzer.go#L261

I think a separate PR is needed to have a generic function and replace others dedupe functions as well.

banjoh: I was referring to https://gobyexample.com/generics, but it can be done in another PR

banjoh (dismissed):

pkg/collect/collect.go
banjoh: Specifically, like so

```go
func Dedup[T any](objs []T) []T {
	seen := make(map[string]bool)
	out := []T{}

	for _, o := range objs {
		data, err := json.Marshal(o)
		if err != nil {
			out = append(out, o)
			continue
		}
		key := string(data)
		if _, ok := seen[key]; !ok {
			out = append(out, o)
			seen[key] = true
		}
	}
	return out
}
```

Then use it like this

```go
func DedupHostCollectors(allCollectors []*troubleshootv1beta2.HostCollect) []*troubleshootv1beta2.HostCollect {
	return utils.Dedup(allCollectors)
}

func DedupHostAnalyzers(allAnalyzers []*troubleshootv1beta2.HostAnalyze) []*troubleshootv1beta2.HostAnalyze {
	return interfaceutils.Dedup(allAnalyzers)
}
```

You probably do not even need `DedupHostAnalyzers` and `DedupHostCollectors`

DexterYan (changes requested):

pkg/collect/collect.go
nvanthao: many thanks @banjoh! with feedback from @DexterYan  as well on the collector dedup, I have followed our suggestions and updated accordingly.

cmd/troubleshoot/cli/run.go
nvanthao: thanks @DexterYan I have updated with collector dedup using generic

banjoh (approved):
Ship it

DexterYan (approved):

PR #1567, opened 2024-06-19
feat: [sc-106759] Troubleshoot: uri field only download when we're not downloading

nvanthao:
Story details: https://app.shortcut.com/replicated/story/106759
Demo: https://asciinema.org/a/6D8BHehl9We2v8IWWIr4ZNUa5

Sample spec: https://gist.githubusercontent.com/nvanthao/4b6314e23fbc9dbca5cd7c7acb9719b9/raw/06dcea2d3473435429da7981d83bcd8d534f32ad/sb.yaml


cmd/troubleshoot/cli/run.go
```diff
@@ -293,7 +293,8 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface)
 	}
 
 	// Load additional specs from support bundle URIs
-	if !viper.GetBool("no-uri") {
+	// only when no-uri flag is not set and no URLs are provided in the args
+	if !viper.GetBool("no-uri") && !hasURL(args) {
```
banjoh: Using `support-bundle --dry-run https://my.host.com/spec.yaml spec.yaml`, if `spec.yaml` has a uri, it will also be ignored.

banjoh (changes requested):

cmd/troubleshoot/cli/run.go
nvanthao: Gotcha! I've made the updates. Please let me know how it goes.

cmd/troubleshoot/cli/run_test.go
```diff
@@ -371,3 +371,49 @@ spec:
...
+	apiVersion: troubleshoot.sh/v1beta2
+	kind: SupportBundle
+	metadata:
+	  name: sb-2
+	spec:
+	  collectors:
+	    - logs:
+          name: podlogs/kotsadm
+          selector:
+            - app=kotsadm`))
```
banjoh: Not that it will change the test outcome, but indentation in this spec is not quite right. I think you have tabs instead of spaces

banjoh (dismissed):
LGTM

cmd/troubleshoot/cli/run_test.go
nvanthao: thanks @banjoh! I've updated the test.

xavpaice (approved):

PR #1571, opened 2024-07-05
fix: [sc-107456] exec collector is running in all pods matched the selector

### LINKED ISSUE #1568: exec collector is running in all pods matched the selector
**Bug Description**

Our [doc](https://troubleshoot.sh/docs/collect/exec/) states that
>When the selector refers to more than one replica of a pod, the exec collector will execute in only one of the pods

**Expected Behavior**
Only exec in 1 pod rather than all pods matched the selector

**Steps To Reproduce**
- Run exec collector to a Deployment with 2 replicas
- Notice exec are running in 2 pods

**Additional Context**
- Troubleshoot version: `0.93.2`

https://github.com/replicatedhq/troubleshoot/blob/53113c0170ce4d1d02ef7552e5cebdc5ae6d4feb/pkg/collect/exec.go#L86-L92

nvanthao:
Story details: https://app.shortcut.com/replicated/story/107456
Fixes: https://github.com/replicatedhq/troubleshoot/issues/1568
Demo: https://asciinema.org/a/bcM9XfqjtztAonQ9rD0AND7aq

### DIFF
--- a/pkg/collect/exec.go
+++ b/pkg/collect/exec.go
@@ -83,21 +83,20 @@ func execWithoutTimeout(clientConfig *rest.Config, bundlePath string, execCollec
 	}
 
 	if len(pods) > 0 {
-		for _, pod := range pods {
-			stdout, stderr, execErrors := getExecOutputs(ctx, clientConfig, client, pod, execCollector)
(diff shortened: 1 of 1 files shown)


banjoh (approved):
LGTM

PR #1572, opened 2024-07-05
feat: [sc-106927] Allow kernelConfig analyser to check kernel capability is either built in or loaded for EC host preflights

### LINKED ISSUE #1563: Allow kernelConfig analyser to check kernel capability is either built in or loaded

When we define an analyser to check for the availability of a kernel config, it would look something like this

```yaml
- kernelConfigs:
    selectedConfigs:
      - CONFIG_CGROUP_FREEZER=y
      - CONFIG_NETFILTER_XTABLES=m
    outcomes:
      - pass:
          message: "required kernel configs are available"
      - fail:
          message: "missing kernel config(s): {{ .ConfigsNotFound }}"
```

Values in the `selectedConfigs` list are checked using an AND operation i.e they all need to exist or not. In some environments, a kernel configuration may have been loaded as a module (`m`) or built into the kernel (`y`) depending of the kernel version. We are not able to have a spec like below to allow a preflight to capture both cases.

```yaml
    selectedConfigs:
      - CONFIG_CGROUP_FREEZER=y
      - CONFIG_CGROUP_FREEZER=m
```

This will limit users to just detecting a configuration is either loaded as a module or in-built into the kernel.

**Describe the feature**

Have the analyser accept a config with multiple values i.e `CONFIG_X=my`. This would mean that we would be able to define a spec like below which would check whether a kernel configuration exists irrelevant of whether it was loaded as a module or in-built into the kernel. The ability of checking one of the other (`m` or `y`) is still preserved with this approach.

```yaml
- kernelConfigs:
    selectedConfigs:
      - CONFIG_CGROUP_FREEZER=my
      - CONFIG_NETFILTER_XTABLES=my
    outcomes:
      - pass:
          message: "required kernel configs are available"
      - fail:
          message: "missing kernel config(s): {{ .ConfigsNotFound }}"
```

nvanthao:
Story details: https://app.shortcut.com/replicated/story/106927
Fixes: https://github.com/replicatedhq/troubleshoot/issues/1563

### DIFF
--- a/pkg/analyze/host_kernel_configs.go
+++ b/pkg/analyze/host_kernel_configs.go
@@ -40,7 +40,7 @@ func (a *AnalyzeHostKernelConfigs) Analyze(
 	}
 
 	var configsNotFound []string
-	kConfigRegex := regexp.MustCompile("^(CONFIG_[A-Z0-9_]+)=([ymn])$")
+	kConfigRegex := regexp.MustCompile("^(CONFIG_[A-Z0-9_]+)=([ymn]+)$")
 	for _, config := range hostAnalyzer.SelectedConfigs {
 		matches := kConfigRegex.FindStringSubmatch(config)
 		// zero tolerance for invalid kernel config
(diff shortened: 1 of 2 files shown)


banjoh (approved):
LGTM

PR #1578, opened 2024-07-16
fix: [sc-106256] Add missing uri field to troubleshoot.sh types

nvanthao:
Story details: https://app.shortcut.com/replicated/story/106256
Demo: https://asciinema.org/a/c2tGPQrGZt6CM3jJQDZqKBsfv

This PR let `preflight` CLI load spec that has `uri` field, continued from https://github.com/replicatedhq/troubleshoot/pull/1574

I used Go reflect to extract all unique `.Spec.uri` in all troubleshoot objects. Once the PR is merged, we can perhaps update  current logic in `supportbundle` CLI

https://github.com/replicatedhq/troubleshoot/blob/191ebdb598dbab4dc135b4c167be709c5aed7828/cmd/troubleshoot/cli/run.go#L295-L302

### DIFF
--- a/internal/specs/specs.go
+++ b/internal/specs/specs.go
@@ -361,3 +361,31 @@ func LoadFromCluster(ctx context.Context, client kubernetes.Interface, selectors
 		RawSpecs: rawSpecs,
 	})
 }
+
+// LoadAdditionalSpecFromURIs loads additional specs from the URIs provided in the troubleshoot kinds.
+// This function will modify kinds in place.
+func LoadAdditionalSpecFromURIs(ctx context.Context, kinds *loader.TroubleshootKinds) {
+	uris := kinds.GetURIs()
+	if len(uris) == 0 {
+		klog.Info("No additional URIs found in all specs")
+		return
+	}
+	for _, uri := range uris {
+		rawSpec, err := downloadFromHttpURL(ctx, uri, nil)
+		if err != nil {
+			klog.Warningf("failed to download spec from URI %q: %v", uri, err)
+			continue
+		}
+		k, err := loader.LoadSpecs(ctx, loader.LoadOptions{RawSpec: string(rawSpec)})
+		if err != nil {
+			klog.Warningf("failed to load spec from URI %q: %v", uri, err)
+			continue
+		}
+		kinds.Add(k)
+	}
+
+	// dedup these top level specs
+	kinds.SupportBundlesV1Beta2 = util.Dedup(kinds.SupportBundlesV1Beta2)
+	kinds.PreflightsV1Beta2 = util.Dedup(kinds.PreflightsV1Beta2)
+	kinds.HostPreflightsV1Beta2 = util.Dedup(kinds.HostPreflightsV1Beta2)
+}
--- a/internal/specs/specs.go
+++ b/internal/specs/specs.go
@@ -362,30 +362,80 @@ func LoadFromCluster(ctx context.Context, client kubernetes.Interface, selectors
 	})
 }
 
-// LoadAdditionalSpecFromURIs loads additional specs from the URIs provided in the troubleshoot kinds.
+// LoadAdditionalSpecFromURIs loads additional specs from the URIs provided in the kinds.
 // This function will modify kinds in place.
 func LoadAdditionalSpecFromURIs(ctx context.Context, kinds *loader.TroubleshootKinds) {
-	uris := kinds.GetURIs()
-	if len(uris) == 0 {
-		klog.Info("No additional URIs found in all specs")
-		return
-	}
-	for uri := range uris {
-		rawSpec, err := downloadFromHttpURL(ctx, uri, nil)
-		if err != nil {
-			klog.Warningf("failed to download spec from URI %q: %v", uri, err)
-			continue
-		}
-		k, err := loader.LoadSpecs(ctx, loader.LoadOptions{RawSpec: string(rawSpec)})
-		if err != nil {
-			klog.Warningf("failed to load spec from URI %q: %v", uri, err)
-			continue
-		}
-		kinds.Add(k)
-	}
 
-	// dedup these top level specs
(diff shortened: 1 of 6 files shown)


banjoh (changes requested):
When I tested this out, I could not replace the entire spec with what got downloaded from a `uri`. Instead, the specs get merged together

**Original spec**
```yaml
kind: HostPreflight
metadata:
  name: my-spec
spec:
  uri: https://gist.githubusercontent.com/banjoh/142fabe108e199d42d442cc0017c7064/raw/a75ceff7f1ec40b58fc38c0dde70a20e9453b7f0/filesystem-preflight.yaml
  collectors:
  - diskUsage:
      collectorName: var-lib-kubelet
      path: /var/lib/kubelet
```
**Wrongly updated spec**
```yaml
kind: HostPreflight
metadata:
  creationTimestamp: null
  name: my-spec
spec:
  analyzers:
  - filesystemPerformance:
      collectorName: Filesystem latency check
      outcomes:
      - pass:
          message: 'Write latency is ok (p99 target < 10ms, actual: {{ .P99 }})'
          when: p99 < 10ms
      - fail:
          message: Write latency is high. p99 target < 10ms, actual:{{ .String }}
  collectors:
  - diskUsage:
      collectorName: var-lib-kubelet
      path: /var/lib/kubelet
  - filesystemPerformance:
      backgroundIOPSWarmupSeconds: 10
      backgroundReadIOPS: 50
      backgroundReadIOPSJobs: 1
      backgroundWriteIOPS: 300
      backgroundWriteIOPSJobs: 6
      collectorName: Filesystem Latency Two Minute Benchmark
      datasync: true
      directory: /var/lib/etcd
      enableBackgroundIOPS: true
      fileSize: 22Mi
      timeout: 2m
  uri: https://gist.githubusercontent.com/banjoh/142fabe108e199d42d442cc0017c7064/raw/a75ceff7f1ec40b58fc38c0dde70a20e9453b7f0/filesystem-preflight.yaml
status: {}
```

**Expected spec update**
```yaml
kind: HostPreflight
metadata:
  name: fs-preflight
spec:
  collectors:
    - filesystemPerformance:
        collectorName: Filesystem Latency Two Minute Benchmark
        timeout: 2m
        directory: /var/lib/etcd
        fileSize: 22Mi
        operationSizeBytes: 2300
        datasync: true
        enableBackgroundIOPS: true
        backgroundIOPSWarmupSeconds: 10
        backgroundWriteIOPS: 300
        backgroundWriteIOPSJobs: 6
        backgroundReadIOPS: 50
        backgroundReadIOPSJobs: 1
  analyzers:
    - filesystemPerformance:
        collectorName: Filesystem latency check
        outcomes:
          - pass:
              when: "p99 < 10ms"
              message: "Write latency is ok (p99 target < 10ms, actual: {{ .P99 }})"
          - fail:
              message: "Write latency is high. p99 target < 10ms, actual:{{ .String }}"
```

internal/specs/specs.go
```diff
@@ -361,3 +361,31 @@ func LoadFromCluster(ctx context.Context, client kubernetes.Interface, selectors
 		RawSpecs: rawSpecs,
 	})
 }
+
+// LoadAdditionalSpecFromURIs loads additional specs from the URIs provided in the troubleshoot kinds.
+// This function will modify kinds in place.
+func LoadAdditionalSpecFromURIs(ctx context.Context, kinds *loader.TroubleshootKinds) {
+	uris := kinds.GetURIs()
```
banjoh: At this point, you loose context of what uri belongs to what spec

pkg/loader/loader.go
```diff
@@ -131,6 +131,36 @@ func (kinds *TroubleshootKinds) ToYaml() (string, error) {
 	return strings.Join(rawList, "---\n"), nil
 }
 
+// GetURIs dynamically extracts all the URIs from the troubleshoot kinds
+// .Spec.Uri field
+func (kinds *TroubleshootKinds) GetURIs() map[string]bool {
```
banjoh: I would keep this function internal to troubleshoot. Its an implementation utility function to assist updating specs, which consumers of the library won't necessarily need to do

pkg/loader/loader.go
```diff
@@ -131,6 +131,36 @@ func (kinds *TroubleshootKinds) ToYaml() (string, error) {
...
+			specField := val.FieldByName("Spec")
+			if !specField.IsValid() {
+				continue
+			}
+			uriField := specField.FieldByName("Uri")
+			if !uriField.IsValid() || uriField.Kind() != reflect.String {
+				continue
+			}
+
+			uris[uriField.String()] = true
```
banjoh: I think this code will include empty strings (`""`) if the spec does not have any `uri` field defined, or if its empty

internal/specs/specs.go
nvanthao: Thanks @banjoh! May I know why do we have to maintain the context of URI-spec? As the final result is a merged object from all sources (argument, secret, config map, etc.)

banjoh: A spec downloaded from a `uri` is meant to [replace the original one entirely, not merge](https://github.com/replicatedhq/troubleshoot/blob/main/docs/design/proposal-updateable-specs.md#goals). Without knowing which spec the url is linked to, it won't be possible to know which spec needs to be replaced.

nvanthao: Argh I missed this design doc, let me update the code accordingly.

nvanthao:
Hi @banjoh,

I've made changes to the code to follow our design doc. That is, to [replace the original one entirely, not merge](https://github.com/replicatedhq/troubleshoot/blob/main/docs/design/proposal-updateable-specs.md#goals)

Updated demo: https://asciinema.org/a/2n6nrYBIByC6SIuTsTSInX1sC

Please let me know how it goes.

banjoh (approved):

PR #1586, opened 2024-07-24
feat: [sc-108689] troubleshoot: journald collector

nvanthao:
Story details: https://app.shortcut.com/replicated/story/108689
Demo: https://asciinema.org/a/NQLJdF8t759caljDvi8Ci8xiC
Doc update: https://github.com/replicatedhq/troubleshoot.sh/pull/569


banjoh (approved):
LGTM

adamancini:
@nvanthao the docs over at https://github.com/replicatedhq/troubleshoot.sh/pull/569 say that the only supported options for `output` are `"short"`, `"json"`, or `"cat"`; but I don't see that validated in the code - is the intent to only support those options (meaning, the other options format are not very useful for Troubleshoot's purposes)?

nvanthao:
Hi @adamancini, I've updated the doc to remove the misleading supported options. As we are allowing all options in `journalctl` [man page](https://man7.org/linux/man-pages/man1/journalctl.1.html).

PR #1589, opened 2024-07-30
feat: etcd collector

nvanthao:

New `etcd` collector that will collect these output from a Kubernetes cluster.

Demo: https://www.loom.com/share/6f72f73d983b457eada0044bd19f9406

```
etcdctl endpoint health
etcdctl endpoint status
etcdctl member list
etcdctl alarm list
```

We only supported distributions that we have access to `etcd`:
- kURL (static pod by kubeadm)
- k0s/embedded-cluster, process managed by `k0scontroller` 

## Checklist

- [x] Any documentation required has been added/updated. https://github.com/replicatedhq/troubleshoot.sh/pull/572

- [ ] Yes
- [X] No


pkg/collect/etcd.go
```diff
@@ -0,0 +1,375 @@
...
+	if len(parts) == 0 {
+		return ""
+	}
+	return strings.Join(parts[1:], "-")
+}
+
+// getFullPath returns the full path to the file
+// e.g. "endpoint-health" -> "etcd/endpoint-health.txt"
+func getFullPath(fileName string) string {
+	return fmt.Sprintf("%s/%s.txt", etcdOutputDir, fileName)
```
banjoh: nit: Wouldn't this be better named as `<filename>.json`?

nvanthao: that makes sense @banjoh, I've updated the code.

Makefile
```diff
@@ -105,6 +105,9 @@ tidy:
 bin/support-bundle:
 	go build ${BUILDFLAGS} ${LDFLAGS} -o bin/support-bundle github.com/replicatedhq/troubleshoot/cmd/troubleshoot
 
+bin/support-bundle-linux:
```
banjoh: This is a good addition. Would the target below be better to get all binaries? Builds run in parallel, so additional compiling time should be negligible

```
build-linux: tidy
	@echo "Build cli binaries"
	GOOS=linux GOARCH=amd64 $(MAKE) -j bin/support-bundle bin/preflight bin/analyze bin/collect
```

pkg/collect/etcd.go
```diff
@@ -0,0 +1,389 @@
...
+		err := c.createEtcdPod()
+		if err != nil {
+			return errors.Wrap(err, "failed to create etcd pod")
+		}
+		return nil
+	}
+	// if not ephemeral, find the static etcd pod to exec into
+	// get the first etcd pod in the cluster with label "component=etcd" in all namespaces
+	label := "component=etcd"
+	pods, err := c.client.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{
```
banjoh:
```suggestion
	pods, err := c.client.CoreV1().Pods("").List(c.context, metav1.ListOptions{
```

pkg/collect/etcd.go
```diff
@@ -0,0 +1,389 @@
...
+			Labels:       labels,
+		},
+		Spec: corev1.PodSpec{
+			HostNetwork: true,
+			Containers: []corev1.Container{
+				{
+					Name:    "etcd-client",
+					Image:   "quay.io/coreos/etcd:latest",
+					Command: []string{"sleep"},
+					Args:    []string{"1d"},
```
banjoh: Even if we delete the pod once done, isn't sleeping `1d` too long? If deleting fails, say deletion fails or interrupted by user, the pod will be lingering there for a while. A shorter time period like `5m` would be better IMO

pkg/collect/etcd.go
```diff
@@ -0,0 +1,389 @@
...
+			GenerateName: "etcd-collector-",
+			Namespace:    namespace,
+			Labels:       labels,
+		},
+		Spec: corev1.PodSpec{
+			HostNetwork: true,
+			Containers: []corev1.Container{
+				{
+					Name:    "etcd-client",
+					Image:   "quay.io/coreos/etcd:latest",
```
banjoh: I think this ought to be passed in as input. In airgap EC environments, we would bundle an etcd image and then pass in the image name to the collector i.e the image would be something like `registry.namespace/etcd:tag`

nvanthao: that's a great idea! I've made the requested changes. please let me know how it goes/

pkg/collect/etcd.go
```diff
@@ -0,0 +1,394 @@
...
+func (c *CollectEtcd) Collect(progressChan chan<- interface{}) (CollectorResult, error) {
+	debugInstance := etcdDebug{
+		context:      c.Context,
+		clientConfig: c.ClientConfig,
+		client:       c.Client,
+		commands: []string{
+			"etcdctl endpoint health",
+			"etcdctl endpoint status",
+			"etcdctl member list",
+			"etcdctl alarm list",
```
banjoh, line 63: Can we add `etcdctl version` here?

Output looks like this
```
etcdctl version: 3.5.1
API version: 3.5
```

banjoh (dismissed):
LGTM

banjoh (approved):

PR #1606, opened 2024-09-02
feat: [sc-110727] troubleshoot: collector/analyzer for wildcard dns

nvanthao:
Story details: https://app.shortcut.com/replicated/story/110727

Demo: https://asciinema.org/a/C9TJdETq9Hn21jefJaxSnWyga

## Updates in existing DNS collector
- use `dig` instead of `nslookup`
- test DNS resolution for a non resolvable domain
- make DNS utility `image` configurable, default to `registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3`
- add `nonResolveable` config for non-resolvable domain, defaulted to `non-existent-domain`

Current JSON output that can be used with others analyzer such as JSON analyzer

```json
{
  "kubernetesClusterIP": "10.43.0.1",
  "podResolvConf": "search default.svc.cluster.local svc.cluster.local cluster.local\nnameserver 10.43.0.10\noptions ndots:5\n",
  "query": {
    "kubernetes": {
      "name": "kubernetes",
      "address": "10.43.0.1"
    },
    "nonResolvableDomain": {
      "name": "foo.bar.xyz",
      "address": ""
    }
  },
  "kubeDNSPods": [
    "coredns-77ccd57875-76dt4"
  ],
  "kubeDNSService": "10.43.0.10",
  "kubeDNSEndpoints": "10.42.0.6:53"
}
```

Sample YAML spec

```yaml
kind: SupportBundle
metadata:
  name: support-bundle
spec:
  collectors:
    - clusterResources:
        exclude: true
    - dns:
        image: registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3
        nonResolvable: foo.bar.xyz 
  analyzers:
    - jsonCompare:
        checkName: Check for wildcard DNS config
        fileName: dns/debug.json
        path: "query.nonResolvableDomain.address"
        value: |
          ""
        outcomes:
          - fail:
              when: "false"
              message: Possible wildcard DNS configured. Non-existent domain resolved to {{ .query.nonResolvableDomain.address }}.
          - pass:
              when: "true"
              message: Confirmed that non-existent domain not resolved.
```

### DIFF
--- a/pkg/collect/dns.go
+++ b/pkg/collect/dns.go
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	dnsUtilsImage       = "registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3"
+	dnsUtilsImage       = "registry.k8s.io/e2e-test-images/agnhost:2.39"
 	nonResolvableDomain = "non-existent-domain"
 )
 
--- a/pkg/collect/dns.go
+++ b/pkg/collect/dns.go
@@ -82,7 +82,13 @@ func (c *CollectDNS) Collect(progressChan chan<- interface{}) (CollectorResult, 
 	// TODO: should nonResolvableDomain be configurable?
 	nonResolvableDomain := "non-existent-domain"
 	dnsDebug.Query.NonResolvableDomain.Name = nonResolvableDomain
-	podLog, err := troubleshootDNSFromPod(c.Client, ctx, nonResolvableDomain)
+
+	image := c.Collector.Image
+	if image == "" {
+		image = dnsUtilsImage
+	}
+
+	podLog, err := troubleshootDNSFromPod(c.Client, ctx, nonResolvableDomain, image)
 	if err == nil {
 		sb.WriteString(fmt.Sprintf("=== Test DNS resolution in pod %s: \n", dnsUtilsImage))
 		sb.WriteString(podLog)
@@ -149,7 +155,7 @@ func getKubernetesClusterIP(client kubernetes.Interface, ctx context.Context) (s
 	return service.Spec.ClusterIP, nil
 }
 
-func troubleshootDNSFromPod(client kubernetes.Interface, ctx context.Context, nonResolvableDomain string) (string, error) {
+func troubleshootDNSFromPod(client kubernetes.Interface, ctx context.Context, nonResolvableDomain string, image string) (string, error) {
 	namespace := "default"
 	command := []string{"/bin/sh", "-c", fmt.Sprintf(`
 		echo "=== /etc/resolv.conf ==="
@@ -175,7 +181,7 @@ func troubleshootDNSFromPod(client kubernetes.Interface, ctx context.Context, no
 			Containers: []corev1.Container{
 				{
 					Name:    "troubleshoot-dns",
-					Image:   dnsUtilsImage,
+					Image:   image,
 					Command: command,
 				},
 			},
--- a/pkg/collect/dns.go
+++ b/pkg/collect/dns.go
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	dnsUtilsImage = "registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3"
+	dnsUtilsImage       = "registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3"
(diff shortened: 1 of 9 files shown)


pkg/collect/dns.go
```diff
@@ -271,3 +328,35 @@ func getKubeDNSEndpoints(client kubernetes.Interface, ctx context.Context) strin
 
 	return strings.Join(endpointStrings, ", ")
 }
+
+func extractDNSQueriesFromPodLog(podLog string, dnsDebug *DNSTroubleshootResult) error {
```
banjoh, line 332: A test or two around this parsing logic would be helpful

pkg/collect/dns.go
```diff
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	dnsUtilsImage = "registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3"
+	dnsUtilsImage       = "registry.k8s.io/e2e-test-images/jessie-dnsutils:1.3"
```
banjoh: [Debian jessie EOL'd in 2020](https://www.debian.org/News/2020/20200709). Why not just use something like `busybox` or `alpine`?

pkg/collect/dns.go
```diff
@@ -104,14 +158,17 @@ func getKubernetesClusterIP(client kubernetes.Interface, ctx context.Context) (s
 	return service.Spec.ClusterIP, nil
 }
 
-func troubleshootDNSFromPod(client kubernetes.Interface, ctx context.Context) (string, error) {
+func troubleshootDNSFromPod(client kubernetes.Interface, ctx context.Context, nonResolvableDomain string, image string) (string, error) {
 	namespace := "default"
-	command := []string{"/bin/sh", "-c", `
-		set -x
+	command := []string{"/bin/sh", "-c", fmt.Sprintf(`
```
banjoh, line 163: Is there a reason why you switched from `nslookup` to `dig`? People will need find an image that has `dig` installed which is an additional hurdle to jump. Images such as `busybox` and `alpine` package `nslookup`. Embedded Cluster uses [busybox](https://github.com/replicatedhq/embedded-cluster/blob/2a28dc6f0e902f2fd9ac47fa80e714f3e166a227/operator/charts/embedded-cluster-operator/values.yaml#L16) which, if not overridden, will be present in the airgap bundle

It might be slightly more code, but you can attempt `dig` and if the binary is not present, try `nslookup`. The collector will be more robust this way.

banjoh (changes requested):

pkg/collect/dns.go
nvanthao: Above image is used because we are following the troubleshooting steps in https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution

I didn't use busybox/alpine because they have problem with inconsistent DNS resolution due to musl implementation.

I noticed there's recent update from K8S https://github.com/kubernetes/kubernetes/issues/126936#issuecomment-2318611725 here, I think we can switch the image to `registry.k8s.io/e2e-test-images/agnhost:2.39`

pkg/collect/dns.go
nvanthao: The `nslookup` command in busybox [is not working correctly](https://bugs.busybox.net/show_bug.cgi?id=14671)

E.g.

```bash
k run -it --rm debug --image busybox -- /bin/sh
If you don't see a command prompt, try pressing enter.
/ #
/ # nslookup kubernetes.default
Server:         10.43.0.10
Address:        10.43.0.10:53

** server can't find kubernetes.default: NXDOMAIN
```

I've changed to `dig` simply because of the base image has `dig` installed, we only want to resolve an address, and the `dig +short` is easier to parse.

pkg/collect/dns.go
banjoh: Good call out on alpine dns inconsistencies. `agnhost` is kept up to date

pkg/collect/dns.go
banjoh: Could you please add this as a comment in the code for future context on the technical choice?

banjoh (approved):
LGTM

pkg/collect/dns.go
nvanthao: many thanks @banjoh! I'll update the docs accordingly as well.

PR #1608, opened 2024-09-09
feat(support-bundle): add runHostCollectorsInPod in spec

DexterYan:
- Add runHostCollectorsInPod for SupportBundle yaml
```
kind: SupportBundle
metadata:
  name: sb
  runHostCollectorsInPod: true  # default is false 
spec:
  hostCollectors:
  - hostOS: {}
```
- Re-use current CollectRemote function to host collect
```
type HostCollector interface {
	Title() string
	IsExcluded() (bool, error)
	Collect(progressChan chan<- interface{}) (map[string][]byte, error)
	RemoteCollect(progressChan chan<- interface{}) (map[string][]byte, error)
	HasRemoted() bool
}
```
- Copy CollectRemote into common go for avoiding import cycle
- Support multiple node HostOS collector result but save only one for now
- Enable HostOS collector for running a privileged pod

[sc-111270](https://app.shortcut.com/replicated/story/111270/update-the-hostos-host-collector-to-run-in-a-pod-on-all-nodes#activity-111971)

Test Yaml
```
kind: SupportBundle
metadata:
  name: sb
  runHostCollectorsInPod: true  # default is false 
spec:
  hostCollectors:
  - hostOS: {}
  hostAnalyzers:
  - hostOS: {}

```

Result
[support-bundle-2024-09-12T16_02_26.tar.gz](https://github.com/user-attachments/files/16974382/support-bundle-2024-09-12T16_02_26.tar.gz)

## Checklist

- [x] Yes
- [ ] No

This PR is backward compatible.


pkg/collect/host_collector.go
```diff
@@ -8,6 +8,8 @@ type HostCollector interface {
 	Title() string
 	IsExcluded() (bool, error)
 	Collect(progressChan chan<- interface{}) (map[string][]byte, error)
+	RemoteCollect(progressChan chan<- interface{}) (map[string][]byte, error)
+	HasRemoted() bool
```
diamonwiggins: What is this for? The naming has me a bit confused.

diamonwiggins: I think i see now. I wonder if this could be renamed to something that would make it clearer what this is for.

cmd/troubleshoot/cli/run.go
```diff
@@ -183,6 +183,7 @@ func runTroubleshoot(v *viper.Viper, args []string) error {
 		OutputPath:                v.GetString("output"),
 		Redact:                    v.GetBool("redact"),
 		FromCLI:                   true,
+		RunHostCollectorsInPod:    false || mainBundle.Metadata.RunHostCollectorsInPod,
```
diamonwiggins: Can this just be set to `mainBundle.Metadata.RunHostCollectorsInPod`?

cmd/troubleshoot/cli/run.go
```diff
@@ -327,14 +328,22 @@ func loadSpecs(ctx context.Context, args []string, client kubernetes.Interface)
...
+			RunHostCollectorsInPod: false,
 		},
 	}
+
+	var enableRunHostCollectorsInPod bool
+
 	for _, sb := range kinds.SupportBundlesV1Beta2 {
 		sb := sb
 		mainBundle = supportbundle.ConcatSpec(mainBundle, &sb)
+		enableRunHostCollectorsInPod = enableRunHostCollectorsInPod || sb.Metadata.RunHostCollectorsInPod
```
diamonwiggins: Will this cause a panic if `sb.Metadata` was `nil` for some reason? Should we check that it isn't nil here and then just set `enableRunHostCollecorsInPod := sb.Metadata.RunHostCollectorsInPod`?

diamonwiggins (changes requested):
Left a few comments. Also a couple questions:

1. Should we refactor things in this PR so that we move the remote collector functions to `internal` and then have both Collectors and RemoteCollectors use those? That way we're in a better place to deprecate RemoteCollectors soon? Or would you rather just do that in a follow up?

2. I don't see any of the code for actually making the pod that's used as the runner privileged. Is that going to be in a follow up as well?

cmd/troubleshoot/cli/run.go
DexterYan: good catch, it has been set to `mainBundle.Metadata.RunHostCollectorsInPod`

cmd/troubleshoot/cli/run.go
DexterYan: yes, we should check that. I have added the checking part

DexterYan:
The runner privileged has been added into the current code. For moving things to `internal` will be another PR

pkg/collect/host_collector.go
```diff
@@ -8,6 +8,8 @@ type HostCollector interface {
 	Title() string
 	IsExcluded() (bool, error)
 	Collect(progressChan chan<- interface{}) (map[string][]byte, error)
+	RemoteCollect(progressChan chan<- interface{}) (map[string][]byte, error) // RemoteCollect is used to priviledge pods to collect data from different nodes
+	HasRemoted() bool                                                         // HasRemoted returns true if the collector has a remote collector implementation
```
diamonwiggins: I wonder if `HasRemoteCollector` would be more descriptive here. Just nitpicking mostly, will leave it up to you.

diamonwiggins (dismissed):
minor comment, but otherwise lgtm

pkg/collect/host_os_info.go
```diff
@@ -18,6 +19,9 @@ type HostOSInfo struct {
 
 const HostOSInfoPath = `host-collectors/system/hostos_info.json`
 
+type NodeInfo struct {
+	HostOSInfo HostOSInfo `json:"host-collectors/system/hostos_info.json"`
```
banjoh: Where is `json:"host-collectors/system/hostos_info.json"` used? Its an interesting use of json tags where the tag is a path. I've not seen before

pkg/collect/host_os_info.go
```diff
@@ -52,3 +56,73 @@ func (c *CollectHostOS) Collect(progressChan chan<- interface{}) (map[string][]b
...
 	return output, nil
 }
+
+func (c *CollectHostOS) RemoteCollect(progressChan chan<- interface{}) (map[string][]byte, error) {
+	restConfig, err := k8sutil.GetRESTConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to convert kube flags to rest config")
+	}
+
+	progressCh := make(chan interface{})
```
banjoh: Should we be using `progressChan` channel for sending progress feedback, instead of creating a new one? The new one is also not doing anything.

pkg/collect/host_certificate.go
```diff
@@ -77,3 +79,11 @@ func isEncryptedKey(filename string) (bool, error) {
 	}
 	return bytes.Contains(data, []byte("ENCRYPTED")), nil
 }
+
+func (c *CollectHostCertificate) RemoteCollect(progressChan chan<- interface{}) (map[string][]byte, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (c *CollectHostCertificate) HasRemoted() bool {
```
banjoh: If the only purpose of `HasRemote()` is to check that a collector has implemented the `RemoteCollect()` function, I'd leave the function out in favour or checking the `not implemented` error. It would be good to keep public functions at a minimum

You could do something like
```go
var ErrNotImplemented = errors.New("not implemented")
...
result, err := collector.RemoteCollect(opts.ProgressChan)
if errors.Is(err, collect.ErrNotImplemented) {
	result, err = collector.Collect(opts.ProgressChan)
	...
}

if err != nil {
...
}
```

banjoh (changes requested):
PR looks good. I requested a few changes

In a future PR we want to start using `"replicated/troubleshoot:<version>"` instead of `"replicated/troubleshoot:latest"` so as to ensure the same version of the image is used as by the binary calling it. There will be an issue with tests passing since the image will most likely not be available.

pkg/collect/host_collector.go
DexterYan: Yes, this function is not easy to read. I have removed it. And use 
```
if err == collect.ErrRemoteCollectorNotImplemented {
```
to check

pkg/collect/host_certificate.go
DexterYan: good idea, I have changed to 
```
if errors.Is(err, collect.ErrRemoteCollectorNotImplemented) {
```

pkg/collect/host_collector.go
DexterYan: updated to use
```
if errors.Is(err, collect.ErrRemoteCollectorNotImplemented) {
```

pkg/collect/host_os_info.go
DexterYan: good catch, it has been removed.

pkg/collect/host_os_info.go
DexterYan: awesome, that has been fixed

pkg/collect/host_os_info.go
```diff
@@ -20,7 +20,7 @@ type HostOSInfo struct {
 const HostOSInfoPath = `host-collectors/system/hostos_info.json`
 
 type NodeInfo struct {
```
banjoh: I cannot see where this is used.

DexterYan: removed

pkg/collect/collect.go
```diff
@@ -28,7 +29,7 @@ type CollectorRunOpts struct {
 	PullPolicy                string
 	LabelSelector             string
 	Timeout                   time.Duration
-	ProgressChan              chan interface{}
```
banjoh, line 31: What is this other change?

banjoh (approved):

banjoh (changes requested):
Can we add an e2e test for this use case?

DexterYan:
e2e test added

test/e2e/support-bundle/hostOS_remote_collector_e2e_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+	}
+
+	feature := features.New("Host OS Remote Collector Test").
+		Assess("check support bundle catch host os remote collector", func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
+			var out bytes.Buffer
+			supportbundleName := "host-os-remote-collector"
+			tarPath := fmt.Sprintf("%s.tar.gz", supportbundleName)
+			targetFolder := fmt.Sprintf("%s/host-collectors/system/", supportbundleName)
+			cmd := exec.CommandContext(context.Background(), sbBinary(), "spec/hostOSRemoteCollector.yaml", "--interactive=false", fmt.Sprintf("-o=%s", supportbundleName))
```
banjoh: Relace `context.Background()` with `ctx` parameter

banjoh (dismissed):

test/e2e/support-bundle/hostOS_remote_collector_e2e_test.go
DexterYan: done!

banjoh (dismissed):

banjoh (approved):

PR #1617, opened 2024-09-18
feat: host collector for DNS

nvanthao:

New host collector DNS that will run DNS query to various domains. Will help with DNS issue such as detect wildcard DNS used.

Fixes: https://app.shortcut.com/replicated/story/112277/analyse-search-domains-with-wildcard-dns-records-on-hosts
Demo: https://asciinema.org/a/Ma1i4T0xp50ODNswrfeHTrOTo

Sample spec

```yaml
kind: SupportBundle
metadata:
  name: sample
spec:
  hostCollectors:
    - dns:
        collectorName: "wildcard-check"
        hostnames:
          - "*"
  analyzers:
    - jsonCompare:
        checkName: Detect wildcard DNS
        fileName: host-collectors/dns/wildcard-check/result.json
        path: "resolvedFromSearch"
        value: |
          ""
        outcomes:
          - fail:
              when: "false"
              message: "Possible wildcard DNS detected at: {{ .resolvedFromSearch }}. Please remove the search domain OR remove the wildcard DNS entry."
          - pass:
              when: "true"
              message: No wildcard DNS detected.
```

Sample output when there's wildcard DNS configured

```json
{
  "query": {
    "*": [
      {
        "server": "127.0.0.53",
        "search": ".foo.testcluster.net.",
        "name": "*.foo.testcluster.net.",
        "answer": "*.foo.testcluster.net.\t60\tIN\tA\t192.1.2.3",
        "record": "192.1.2.3"
      },
      {
        "server": "127.0.0.53",
        "search": ".c.replicated-qa.internal.",
        "name": "*.c.replicated-qa.internal.",
        "answer": "",
        "record": ""
      },
      {
        "server": "127.0.0.53",
        "search": ".google.internal.",
        "name": "*.google.internal.",
        "answer": "",
        "record": ""
      },
      {
        "server": "127.0.0.53",
        "search": ".",
        "name": "*.",
        "answer": "",
        "record": ""
      }
    ]
  },
  "resolvedFromSearch": ".foo.testcluster.net."
}
```

## Checklist

- [ ] Yes
- [x] No


pkg/collect/host_dns.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func getDNSConfig() (*dns.ClientConfig, error) {
+	file, err := os.Open(resolvConf)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	config, err := dns.ClientConfigFromFile(file.Name())
```
diamonwiggins, line 116: Should we have an option on the collector to specify ndots, and then if its set update the DNS config here to set that value?

https://github.com/miekg/dns/blob/master/clientconfig.go#L12-L19

diamonwiggins: ndots is a good example, but anywhere someone might want to override the config from the collector could be useful to expose.

diamonwiggins: Looking further, since you're iterating and appending every search domain found on the host to see if one of them resolves something we suspect it shouldn't, then the ndots stuff probably isn't needed? So whether you decide to add that or not completely up to you at this stage.

pkg/apis/troubleshoot/v1beta2/hostcollector_shared.go
```diff
@@ -218,6 +218,11 @@ type HostJournald struct {
 	Timeout           string   `json:"timeout,omitempty" yaml:"timeout,omitempty"`
 }
 
+type HostDNS struct {
+	HostCollectorMeta `json:",inline" yaml:",inline"`
+	Hostnames         []string `json:"hostnames,omitempty" yaml:"hostnames,omitempty"`
```
banjoh: We don't validate today, but I would replace `omitempty` with `required`

pkg/collect/host_dns.go
```diff
@@ -0,0 +1,187 @@
...
+	return isExcluded(c.hostCollector.Exclude)
+}
+
+func (c *CollectHostDNS) Collect(progressChan chan<- interface{}) (map[string][]byte, error) {
+
+	names := c.hostCollector.Hostnames
+	if len(names) == 0 {
+		// if no names are provided, query a wilcard to detect wildcard DNS if any
+		names = append(names, "*")
+	}
```
banjoh: We should error of not hostnames are provided

banjoh (approved):

PR #1633, opened 2024-10-02
Refactor Remote Host Collection 

diamonwiggins:

Uses a global function in the `collect` package `collect.RemoteHostCollect` instead of a method per collector. This prevents the need to make a change to each collector for now until we have a use case that justifies that level ofgranularity

### Considerations
While this PR does prevent the need of having to maintain code for each collector, everything awkward about it i'd attribute to the remote collection code not being closely coupled with each collector. Some examples include:
* The `RBAC` code added. Everywhere else in the codebase this is implemented on a specific type. This code just feels out of place in this implementation, but to be fair, because the same pod is created for each collector its not as dynamic as the other code for standard `in-cluster` collectors needs to be. We essentially always verify the same permissions(Can I create a Pod/ConfigMap?)
* Host Collector file paths are being derived somewhat arbitrarily. Because this code isn't tightly coupled with each collector, the way the file path in the bundle is being determined is by taking the filename for the collector result, splitting it into `directory` and `filename`, inserting the node name between them, and then saving that. For example, `host-collector/system/cpu.json` becomes `host-collector/system/node-name/cpu.json`. This works, _for now_, but again feels strange, and not sure we'll want to do this longterm. Especially since some collector like `run` have the ability to completely change which file path they use.

### TODO
- [ ] Needs Tests

## Checklist

- [ ] Yes
- [ ] No


pkg/supportbundle/collect.go
```diff
@@ -253,3 +217,132 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+		params := &collect.RemoteCollectParams{
+			ProgressChan:  opts.ProgressChan,
+			HostCollector: &spec,
+			BundlePath:    bundlePath,
+			ClientConfig:  opts.KubernetesRestConfig,
+			Image:         "replicated/troubleshoot:latest",
+			PullPolicy:    "IfNotPresent",
+			Timeout:       time.Duration(60 * time.Second),
+			LabelSelector: "",
+			NamePrefix:    "hostos-remote",
```
DexterYan: should it be renamed to host-remote, since it has been moved into a general function?

pkg/collect/host_collector.go
```diff
@@ -81,3 +110,130 @@ func hostCollectorTitleOrDefault(meta troubleshootv1beta2.HostCollectorMeta, def
...
+		output.SaveResult(params.BundlePath, constants.NODE_LIST_FILE, bytes.NewBuffer(nodesBytes))
+	}
+	return output, nil
+}
+
+func runRemote(ctx context.Context, runner runner, nodes []string, collector *troubleshootv1beta2.HostCollect, nameGenerator names.NameGenerator, namePrefix string, namespace string) (map[string][]byte, error) {
+	g, ctx := errgroup.WithContext(ctx)
+	results := make(chan map[string][]byte, len(nodes))
+
+	for _, node := range nodes {
```
DexterYan, line 183: I am thinking limit the collector to run 1 node per time to reduce the kube-api limit. Not 100% sure, it is worthing
```
func runRemote(ctx context.Context, runner runner, nodes []string, collector *troubleshootv1beta2.HostCollect, nameGenerator names.NameGenerator, namePrefix string, namespace string) (map[string][]byte, error) {
	g, ctx := errgroup.WithContext(ctx)
	results := make(chan map[string][]byte, len(nodes))
	wg := &sync.WaitGroup{}
	workerLimit := make(chan struct{}, 1)

	for _, node := range nodes {
		node := node
		workerLimit <- struct{}{} // limit the number of concurrent workers
		wg.Add(1)

		g.Go(func() error {
			// May need to evaluate error and log warning.  Otherwise any error
			// here will cancel the context of other goroutines and no results
			// will be returned.
			defer wg.Done()
			defer func() { <-workerLimit }()
			return runner.run(ctx, collector, namespace, nameGenerator.GenerateName(namePrefix+"-"), node, results)
		})
	}

	// Wait for all collectors to complete or return the first error.
	if err := g.Wait(); err != nil {
		return nil, errors.Wrap(err, "failed remote collection")
	}
	close(results)

	wg.Wait()

	output := make(map[string][]byte)
	for result := range results {
		r := result
		for k, v := range r {
			output[k] = v
		}
	}

	return output, nil
}

```

diamonwiggins: I wasn't using the same k8s REST config that the standard in cluster collection used. Now that I am, I'm not seeing the throttling anymore. It's possible we would still need to keep an eye on this in the future for very large clusters though

pkg/supportbundle/collect.go
```diff
@@ -253,3 +217,136 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+func filterHostCollectors(ctx context.Context, collectSpecs []*troubleshootv1beta2.HostCollect, bundlePath string, opts SupportBundleCreateOpts) ([]FilteredCollector, error) {
+	var filteredCollectors []FilteredCollector
+
+	for _, desiredCollector := range collectSpecs {
+		collector, ok := collect.GetHostCollector(desiredCollector, bundlePath)
+		if !ok {
+			return nil, collect.ErrHostCollectorNotFound
+		}
+
+		isExcluded, _ := collector.IsExcluded()
```
banjoh, line 342: We will fail to add skipped collectors in [otel tracing](https://github.com/replicatedhq/troubleshoot/blob/0240a632c91ee94acc4ba47c6ce644c2c5cd1e64/pkg/supportbundle/collect.go#L47-L48). We can either add this here or move `isExcluded, _ := collector.IsExcluded()` logic into `collectRemoteHost` and `collectHost`

pkg/collect/host_collector.go
```diff
@@ -81,3 +110,130 @@ func hostCollectorTitleOrDefault(meta troubleshootv1beta2.HostCollectorMeta, def
...
+	if params.NamePrefix == "" {
+		params.NamePrefix = remoteCollectorNamePrefix
+	}
+
+	result, err := runRemote(ctx, runner, nodes, params.HostCollector, names.SimpleNameGenerator, params.NamePrefix, params.Namespace)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to run collector remotely")
+	}
+
+	for k, v := range result {
```
banjoh: A test for this logic would be good to have

pkg/collect/host_collector.go
```diff
@@ -81,3 +110,130 @@ func hostCollectorTitleOrDefault(meta troubleshootv1beta2.HostCollectorMeta, def
 	}
 	return defaultTitle
 }
+
+func RemoteHostCollect(params RemoteCollectParams) (map[string][]byte, error) {
```
banjoh: Pass in `ctx` from caller

pkg/collect/host_collector.go
DexterYan: I am wondering, should add a e2e test for multiple node. Or just a function test for 
```
	for k, v := range result {
		if curBytes, ok := allCollectedData[k]; ok {
			var curResults map[string]string
			...
```

DexterYan: I have added a test for it.

DexterYan (approved):

PR #1634, opened 2024-10-02
chore(support-bundle): respect using load-cluster-specs=false

### LINKED ISSUE #1638: support-bundle --load-cluster-specs=false still loads specs from a cluster
**Bug Description**

`support-bundle --load-cluster-specs=false` still leads to loading specs from a cluster.

**Expected Behavior**

`support-bundle --load-cluster-specs=false` should not attempt to load specs from the cluster

banjoh:

Using `./support-bundle --load-cluster-specs=false` still leads to loading support bundle specs from the cluster.

Fixes: https://github.com/replicatedhq/troubleshoot/issues/1638

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/cmd/troubleshoot/cli/root.go
+++ b/cmd/troubleshoot/cli/root.go
@@ -44,6 +44,19 @@ If no arguments are provided, specs are automatically loaded from the cluster by
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v := viper.GetViper()
 
+			// If there are not locations to load specs passed in the cli args, we should
+			// load them from the cluster by setting "load-cluster-specs=true". If the caller
+			// provided "--load-cluster-specs" cli option, we should respect that.
+			if len(args) == 0 {
+				// Check if --load-cluster-specs was set by the cli caller by
+				// checking if the flag was not changed from the default value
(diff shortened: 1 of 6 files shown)


diamonwiggins:
@banjoh any context on this one? FYI i think the reason we decided not to just set the flag to be `true` by default is because we didn't want there to be any hidden magic taking place if a user did specify additional flags.

Today:

`./support-bundle https://some.spec.com` would just collect the spec you provided, but with this change it would also load cluster specs. I'd imagine the tests are probably failing for the same reason. Things now being collected from the cluster that we didn't expect.

edit. are you saying that `./support-bundle --load-cluster-specs=false` doesn't work today? I could see how in the current code and the conditional being based on args that that could be the case. Its sort of confusing how there are some flags we use directly, but others we convert into args. Either way, if this is what you're trying to address, i agree we should fix. However, we'll probably want to make sure that `--load-cluster-specs` isn't always `true` though 🤔

banjoh:
> any context on this one?

As you correctly figured, `./support-bundle --load-cluster-specs=false` does not work today, which is what I'm trying to address in this PR. By default, `--load-cluster-specs` will always be `true`

diamonwiggins (approved):

PR #1653, opened 2024-10-18
bug: fix nil check in host collector filter

nvanthao:
## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/pkg/supportbundle/collect.go
+++ b/pkg/supportbundle/collect.go
@@ -290,7 +290,6 @@ func collectHost(ctx context.Context, filteredCollectors []FilteredCollector, op
 		if err != nil {
 			span.SetStatus(codes.Error, err.Error())
 			opts.ProgressChan <- fmt.Sprintf("[%s] Error: %v", collector.Title(), err)
-			return errors.Wrap(err, "failed to run host collector")
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/supportbundle/collect.go
```diff
@@ -290,7 +290,6 @@ func collectHost(ctx context.Context, filteredCollectors []FilteredCollector, op
 		if err != nil {
 			span.SetStatus(codes.Error, err.Error())
 			opts.ProgressChan <- fmt.Sprintf("[%s] Error: %v", collector.Title(), err)
-			return errors.Wrap(err, "failed to run host collector")
```
DexterYan, line 293: why remove this one, since it is filteredCollectors, should be nil and if it failed, then fail

nvanthao: this is existing behavior prior to `v0.107.0`, we simply continue when there's an error during collection. E.g. we don't want to stop collecting when there's error in 1 collector.

https://github.com/replicatedhq/troubleshoot/blob/77fd7da315f886dd065ae9c1e4451075e1664850/pkg/supportbundle/collect.go#L159-L163

DexterYan (approved):

PR #1658, opened 2024-10-22
Allow `collect` to chroot itself

hedge-sparrow:

This PR allows the `collect` binary to `chroot` based on a flag, this simplifies running it inside a container where the desired root is at a path other than the containers root without the need to copy the `collect` binary itself onto the host. 

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/cmd/collect/cli/root.go
+++ b/cmd/collect/cli/root.go
@@ -58,7 +58,7 @@ func RootCmd() *cobra.Command {
 	cmd.Flags().String("selector", "", "selector (label query) to filter remote collection nodes on.")
 	cmd.Flags().Bool("collect-without-permissions", false, "always generate a support bundle, even if it some require additional permissions")
 	cmd.Flags().Bool("debug", false, "enable debug logging")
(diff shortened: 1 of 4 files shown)


cmd/collect/cli/root.go
```diff
@@ -32,6 +32,10 @@ func RootCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			v := viper.GetViper()
 
+			if err := checkAndSetChroot(v.GetString("chroot")); err != nil {
```
diamonwiggins, line 35: Should we check if we're running as root as well eg. https://github.com/replicatedhq/troubleshoot/blob/main/internal/util/util.go#L115

hedge-sparrow: oh good point. the go chroot syscall does a pretty good job of complaining if it can't chroot but friendlier messages are always better

diamonwiggins (approved):
lgtm

PR #1659, opened 2024-10-23
feat(host-collector): add progress for host collector

DexterYan:
- using `CollectorProgressCallback` to allow host remote collector to send progress back to kotsadm
- allow `collectorCB` to be used as CollectorProgressCallback in remote host collector. https://github.com/replicatedhq/kots/blob/a57ff2d1431d961808c2bbfdf0997abe25ab0e34/pkg/supportbundle/execute.go#L167 
- use `supportBundleProgressUpdate` match the callback type  to report the progress in https://github.com/replicatedhq/kots/blob/a57ff2d1431d961808c2bbfdf0997abe25ab0e34/pkg/supportbundle/execute.go#L79

kots demo
[Library | Loom - 24 October 2024 - Watch Video](https://www.loom.com/share/67fb2598c43245b2afb7bd078d5b7ac5)

[sc-114419](https://app.shortcut.com/replicated/story/114419/kots-per-collector-progress-update-is-to-be-shown-in-admin-console#activity-114596)

## Checklist

- [ ] Yes
- [x] No


pkg/supportbundle/collect.go
```diff
@@ -227,14 +227,16 @@ func collectRemoteHost(ctx context.Context, collectSpecs []*troubleshootv1beta2.
 
 		isExcluded, _ := collector.IsExcluded()
 		if isExcluded {
-			opts.ProgressChan <- fmt.Sprintf("[%s] Excluding host collector", collector.Title())
+			msg := fmt.Sprintf("[%s] Excluding host collector", collector.Title())
+			opts.CollectorProgressCallback(opts.ProgressChan, msg)
```
marccampbell, line 231: what's the reason for not just sending directly to the channel and wrapping it in a callback?

DexterYan: It is required for Kotsadm front to pick up the host remote collector progress. I have put the details in the description

xavpaice: this matches the pattern used with in-cluster collectors, so that KOTS can display the collector progress.  Previously, KOTS didn't know about host collectors so it didn't matter.

nvanthao (commented):
Just add my thoughts,

I see we are using `opts.CollectorProgressCallback` because KOTS uses that callback to touch-up the the progress update.
So this change will have minimal impact to KOTS.

A right way I think, is not to use the callback, but to send into `opts.ProgressChan` a public struct `collect.CollectProgress`, and KOTS, as a consumer, will just use it. This will decouple KOTS and troubleshoot, no callback needed.

I see that we are already doing it for `preflight`

https://github.com/replicatedhq/kots/blob/99b056171e307ae50831d9ac9ddc71b91c84c12f/pkg/preflight/execute.go#L39-L53

That being said, KOTS will have to refactor with more code changes to cater for using `collect.CollectProgress`

nvanthao (approved):
Let's go with it now and we will have a separate card to refactor the progress logic in KOTS

PR #1664, opened 2024-10-28
chore: Refactor host cpu analyzer for remote collection

hedge-sparrow:

## Checklist

- [ ] Yes
- [ ] No


pkg/analyze/host_cpu.go
```diff
@@ -33,94 +35,45 @@ func (a *AnalyzeHostCPU) IsExcluded() (bool, error) {
 	return isExcluded(a.hostAnalyzer.Exclude)
 }
 
-func (a *AnalyzeHostCPU) Analyze(
-	getCollectedFileContents func(string) ([]byte, error), findFiles getChildCollectedFileContents,
-) ([]*AnalyzeResult, error) {
-	hostAnalyzer := a.hostAnalyzer
+func (a *AnalyzeHostCPU) CheckCondition(when string, data collectorData) (bool, error) {
```
marccampbell: why not just set data to type `[]byte` and avoid the reflect package? it seems like the type is known and expected, but we are casting in case it's not that?

diamonwiggins: I agree. I think in my initial refactor this change was made while i was doing some debugging, and i never went back to remove this. @hedge-sparrow feel free to just set the type to `[]byte` and update the other analyzers as well.

hedge-sparrow: hmm good point. in this instance this is just to remain compatible with the existing code, but after speaking to @diamonwiggins I'll update the other analyzers that were implemented using this to not.

hedge-sparrow: @marccampbell just going through this, it looks like `collectorData` being an `interface{}` actually helps with testing https://github.com/replicatedhq/troubleshoot/blob/deda4ce98c38d5a6cafdbdc442493c753cfa0bc1/pkg/analyze/host_analyzer_test.go#L26

we could update `collectorData` to be `[]byte` everywhere pretty easily but I don't have an immediate suggestion for how to keep the tests nice and simple like this, any thoughts?

pkg/analyze/host_analyzer_test.go
```diff
@@ -10,25 +11,29 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func collectorToBytes(collector any) []byte {
```
banjoh: nit: You could pass in `t` and fail if err like so

```go
func collectorToBytes(t *testing.T, collector any) []byte {
	jsonData, err := json.Marshal(collector)
	require.NoError(t, err) 
	return jsonData
}
```

then call the function like so

```
Data: collectorToBytes(t, collect.HostOSInfo{...
```

hedge-sparrow: Sick, thanks

pkg/analyze/host_cpu.go
hedge-sparrow: worked out a nice way to keep test simplicity while removing this `interface{}`

nvanthao (approved):
LGTM

PR #1671, opened 2024-11-01
exec remote host collectors in a daemonset 

hedge-sparrow:
This PR changes how remote host collection is scheduled. 

it switches the runners to be exec calls to pods managed by a daemonset, rather than one runner per pod. 

## Checklist

- [x] Yes
- [ ] No


pkg/supportbundle/supportbundle.go
```diff
@@ -328,8 +308,20 @@ func getNodeList(clientset kubernetes.Interface, opts SupportBundleCreateOpts) (
 
 	nodeList := NodeList{}
 	for _, node := range nodes.Items {
+		if isMasterNode(node) {
```
DexterYan: Is that a intended feature? It seems not work for a single node cluster

DexterYan (dismissed):

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,353 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+		return nil, err
+	}
+
+	klog.V(2).Infof("All remote host collectors completed")
+
+	defer func() {
+		// TODO:
+		// delete the config map
+		// delete the remote pods
+		clientset.AppsV1().DaemonSets(ds.Namespace).Delete(ctx, ds.Name, metav1.DeleteOptions{})
```
nvanthao: Potential panic when ds is not created

nvanthao (changes requested):

pkg/supportbundle/collect.go
DexterYan: I have added

```
	_, err := clientset.AppsV1().DaemonSets(ds.Namespace).Get(ctx, ds.Name, metav1.GetOptions{})
		if err != nil {
			klog.Errorf("Failed to verify remote host collector daemonset %s still exists: %v", ds.Name, err)
			return
		}

		if err := clientset.AppsV1().DaemonSets(ds.Namespace).Delete(ctx, ds.Name, metav1.DeleteOptions{}); err != nil {
			klog.Errorf("Failed to delete remote host collector daemonset %s: %v", ds.Name, err)
		}
```

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+
+	return createdDS, nil
 }
 
-// getGlobalRedactors returns the global redactors from the support bundle spec
-func getGlobalRedactors(additionalRedactors *troubleshootv1beta2.Redactor) []*troubleshootv1beta2.Redact {
-	if additionalRedactors != nil {
-		return additionalRedactors.Spec.Redactors
+func waitForPodRunning(ctx context.Context, clientset kubernetes.Interface, pod *corev1.Pod) error {
+	watcher, err := clientset.CoreV1().Pods(pod.Namespace).Watch(ctx, metav1.ListOptions{
```
banjoh: To improve reliability, should we poll if the user does not have the necessary `"watch"` permissions?

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+					return nil
+				}
+			}
+		}
+	}
+	return fmt.Errorf("pod %s did not complete", pod.Name)
+}
+
+func waitForDS(ctx context.Context, clientset kubernetes.Interface, ds *appsv1.DaemonSet) error {
+	watcher, err := clientset.AppsV1().DaemonSets(ds.Namespace).Watch(ctx, metav1.ListOptions{
```
banjoh: Same here. Poll if no `"watch"` permissions.

We might just default to polling if it does not introduce a significant delay.

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+	// wait for at least one pod to be scheduled
+	err = waitForDS(ctx, clientset, ds)
+	if err != nil {
+		return nil, err
+	}
+
+	klog.V(2).Infof("Created Remote Host Collector Daemonset %s", ds.Name)
+	pods, err := clientset.CoreV1().Pods(ds.Namespace).List(ctx, metav1.ListOptions{
+		LabelSelector:  selectorLabelKey + "=" + selectorLabelValue,
+		TimeoutSeconds: new(int64),
```
banjoh: Is this 0 timeout intentional?

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+		Limit:          0,
```
banjoh, line 347: Wouldn't limit=0 mean that no pods should be returned? I might be missing something here though

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+	req := client.CoreV1().RESTClient().Post().Resource("pods").Name(pod.Name).Namespace(pod.Namespace).SubResource("exec")
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		return nil, nil, err
 	}
-	return nil
+
+	parameterCodec := runtime.NewParameterCodec(scheme)
+	req.VersionedParams(&corev1.PodExecOptions{
+		Command:   []string{"/troubleshoot/collect", "-", "--chroot", "/host", "--format", "raw"},
```
banjoh, line 284: Using stdin instead of relying on configmaps is pretty neat!

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+
+				// convert host collectors into a HostCollector spec
+				spec := createHostCollectorsSpec([]*troubleshootv1beta2.HostCollect{collectorSpec})
+				specJSON, err := json.Marshal(spec)
+				if err != nil {
+					return err
+				}
+				klog.V(2).Infof("HostCollector spec: %s", specJSON)
+
+				stdout, _, err := getExecOutputs(ctx, opts.KubernetesRestConfig, clientset, pod, specJSON)
```
banjoh, line 396: This comment is for future consideration. I suggest we store stderr in the bundle somewhere. We would want to catch errors from `collect` binary if any

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+	}
+
+	klog.V(2).Infof("All remote host collectors completed")
+
+	defer func() {
+		// TODO:
+		// delete the config map
+		// delete the remote pods
+		// check if the daemonset still exists
+		_, err := clientset.AppsV1().DaemonSets(ds.Namespace).Get(ctx, ds.Name, metav1.GetOptions{})
```
banjoh: Wouldn't it be sufficient to just delete the DaemonSet and log the error if it wasn't present?

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+			Name: "remoteHostCollector",
+		},
+		Spec: troubleshootv1beta2.HostCollectorSpec{
+			Collectors: hostCollectors,
+		},
+	}
+}
+
+func createHostCollectorDS(ctx context.Context, clientset kubernetes.Interface, labels map[string]string) (*appsv1.DaemonSet, error) {
+	ns := "default"
```
banjoh: Should we allow passing in the namespace? In embedded cluster for example, `kotsadm` & `embedded-cluster` namespaces exist. KOTS may have permissions on one of these but not to the `default` namespace

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+	imageName := "replicated/troubleshoot:latest"
```
banjoh, line 486: Should we use a versioned tag instead? You can get the version string from https://github.com/replicatedhq/troubleshoot/blob/main/pkg/version/version.go

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+	imagePullPolicy := corev1.PullAlways
```
banjoh: If we use a tag, perhaps this should be `IfNotPresent`

pkg/supportbundle/collect.go
```diff
@@ -207,136 +222,394 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
+		HostNetwork: true,
+		HostPID:     true,
+		HostIPC:     true,
+		Containers: []corev1.Container{
+			{
+				Image:           imageName,
+				ImagePullPolicy: imagePullPolicy,
+				Name:            "remote-collector",
+				Command:         []string{"/bin/bash", "-c"},
+				Args:            []string{"while true; do sleep 30; done;"},
```
banjoh: Collection shouldn't take longer than 30s but I'm thinking we should have a better way of handling this. We can for example implement a `collect pause` subcommand to pause and wait for a termination signal then exit gracefully.

banjoh (changes requested):

DexterYan:
@diamonwiggins 

Copy from another PR
> Is the reporting and the progress updates accurate if the flow is:
> 
> ```
> for pod in pods:
>     run each collector
> ```
> 
> or should it be:
> 
> ```
> for collector in collectors:
>     run in each pod
> ```

pkg/supportbundle/collect.go
DexterYan: Yes, it should be sufficient to just delete the DaemonSet and log the not found error. I have modified the code.

pkg/supportbundle/collect.go
DexterYan: I think if the ds is not created, it will log the not found error.

pkg/supportbundle/collect.go
DexterYan: 0 means that there is no limit on the number of pods returned in this list operation.  It will attempt to return all matching pods without restricting the count.

pkg/supportbundle/collect.go
```diff
@@ -207,136 +223,390 @@ func getAnalysisFile(analyzeResults []*analyze.AnalyzeResult) (io.Reader, error)
...
-	return nil
+
+	klog.V(2).Infof("All remote host collectors completed")
+
+	defer func() {
+		// TODO:
+		// delete the config map
+		// delete the remote pods
+		// check if the daemonset still exists
+		if err := clientset.AppsV1().DaemonSets(ds.Namespace).Delete(ctx, ds.Name, metav1.DeleteOptions{}); err != nil {
```
nvanthao, line 444: `ds` could be nil here if there is error before `ds` is is created, we have to do a nil check for `ds` here if not the code may panic.

DexterYan: I see. I have added ds nil check and ds name check

pkg/supportbundle/collect.go
DexterYan: changed to IfNotPresent

pkg/supportbundle/collect.go
DexterYan: I found the challenging part is that if we are using version string in local development, I found the image was not existed. We have to build it in local first. Do you have any idea about that?

pkg/supportbundle/collect.go
DexterYan: I have changed to `TimeoutSeconds: ptr.To(int64(defaultTimeout)),`

pkg/supportbundle/collect.go
DexterYan: Thank you! I have checked to polling method

pkg/supportbundle/collect.go
DexterYan: changed to polling

pkg/supportbundle/collect.go
DexterYan: I have changed to use 
```
Command:         []string{"tail", "-f", "/dev/null"},
```
 It effectively blocks forever without doing anything.

pkg/supportbundle/collect.go
DexterYan: yes, that is very efficient!

pkg/supportbundle/collect.go
DexterYan: good catch. Namespace has been added

pkg/supportbundle/collect.go
DexterYan: I found one to use version tag. It has to be semantic version

DexterYan:
The remote collectors has been using 
```
for collector in collectors:
   tracing start
    run in each pod
   tracing end
```

pkg/supportbundle/collect.go
DexterYan: It seems cannot pass the tests. I reverts  docker image versioned tag.

DexterYan: I suggest that we can switch to version tag in next PR

nvanthao (approved):
LGTM

banjoh (approved):
LGTM

PR #1672, opened 2024-11-01
Fix remote host collection RBAC checks

diamonwiggins:

* Move save node list code to run after we check for RBAC
* Update RBAC checks to cover everything needed for remote host collectors

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/pkg/supportbundle/rbac.go
+++ b/pkg/supportbundle/rbac.go
@@ -37,6 +37,11 @@ func checkRemoteCollectorRBAC(ctx context.Context, clientConfig *rest.Config, ti
 	resourceAttributesList := []authorizationv1.ResourceAttributes{
 		{
 			Namespace: namespace,
+			Verb:      "get",
+			Resource:  "pods",
+		},
+		{
+			Namespace: namespace,
 			Verb:      "create",
 			Resource:  "pods",
 		},
@@ -46,13 +51,33 @@ func checkRemoteCollectorRBAC(ctx context.Context, clientConfig *rest.Config, ti
 			Resource:  "pods",
 		},
 		{
+			Namespace: namespace,
+			Verb:      "get",
+			Resource:  "pods/log",
+		},
+		{
 			Verb:     "list",
 			Resource: "nodes",
 		},
 		{
 			Namespace: namespace,
+			Verb:      "get",
+			Resource:  "configmaps",
+		},
+		{
+			Namespace: namespace,
 			Verb:      "create",
 			Resource:  "configmaps",
+		},
+		{
+			Namespace: namespace,
+			Verb:      "delete",
+			Resource:  "configmaps",
+		},
+		{
+			Namespace: namespace,
+			Verb:      "get",
+			Resource:  "serviceaccounts",
 		},
 	}
 
--- a/pkg/supportbundle/rbac.go
+++ b/pkg/supportbundle/rbac.go
@@ -34,34 +34,50 @@ func checkRemoteCollectorRBAC(ctx context.Context, clientConfig *rest.Config, ti
 
 	var forbidden []error
 
-	spec := authorizationv1.SelfSubjectAccessReviewSpec{
-		ResourceAttributes: &authorizationv1.ResourceAttributes{
-			Namespace:   namespace,
-			Verb:        "create,delete",
-			Group:       "",
-			Version:     "",
-			Resource:    "pods,configmap",
-			Subresource: "",
-			Name:        "",
+	resourceAttributesList := []authorizationv1.ResourceAttributes{
+		{
+			Namespace: namespace,
+			Verb:      "create",
+			Resource:  "pods",
 		},
-		NonResourceAttributes: nil,
+		{
+			Namespace: namespace,
+			Verb:      "delete",
+			Resource:  "pods",
+		},
+		{
+			Verb:     "list",
+			Resource: "nodes",
+		},
+		{
+			Namespace: namespace,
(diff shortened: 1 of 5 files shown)


pkg/supportbundle/rbac.go
```diff
@@ -34,34 +34,75 @@ func checkRemoteCollectorRBAC(ctx context.Context, clientConfig *rest.Config, ti
...
-			Namespace:   namespace,
-			Verb:        "create,delete",
-			Group:       "",
-			Version:     "",
-			Resource:    "pods,configmap",
-			Subresource: "",
-			Name:        "",
+	resourceAttributesList := []authorizationv1.ResourceAttributes{
+		{
+			Namespace: namespace,
```
nvanthao, line 39: should we check on empty `namespace`?

pkg/supportbundle/rbac.go
```diff
@@ -34,34 +34,75 @@ func checkRemoteCollectorRBAC(ctx context.Context, clientConfig *rest.Config, ti
...
+	for _, resourceAttributes := range resourceAttributesList {
+		spec := authorizationv1.SelfSubjectAccessReviewSpec{
+			ResourceAttributes: &resourceAttributes,
+		}
+
+		sar := &authorizationv1.SelfSubjectAccessReview{
+			Spec: spec,
+		}
+
+		resp, err := client.AuthorizationV1().SelfSubjectAccessReviews().Create(ctx, sar, metav1.CreateOptions{})
```
nvanthao, line 93: what if the caller does not have `selfsubjectaccessreviews/create` RBAC permission? is it considered an RBACPermissionError?

pkg/supportbundle/rbac.go
```diff
@@ -34,34 +34,75 @@ func checkRemoteCollectorRBAC(ctx context.Context, clientConfig *rest.Config, ti
...
-	spec := authorizationv1.SelfSubjectAccessReviewSpec{
-		ResourceAttributes: &authorizationv1.ResourceAttributes{
-			Namespace:   namespace,
-			Verb:        "create,delete",
-			Group:       "",
-			Version:     "",
-			Resource:    "pods,configmap",
-			Subresource: "",
-			Name:        "",
+	resourceAttributesList := []authorizationv1.ResourceAttributes{
```
nvanthao, line 37: this could be a constant declared outside so that we will remember to update it in the future.

nvanthao (commented):
many thanks Diamon! LGTM, just a few comments to make sure I understand the changes.

banjoh (approved):
LGTM

pkg/supportbundle/rbac.go
diamonwiggins: Is there a use case where you think we might need to? Remote host collection can execute within a single namespace right?

pkg/supportbundle/rbac.go
diamonwiggins: Good question and i'm not sure. I never ran into an issue testing locally with a service account that only fulfilled what is listed above, so i'm not sure if `selfsubjectaccessreviews/create` is implicitly granted by default so that `kubectl auth can-i` can work or what. 

Either way, if you think this is something we need to consider, we'd need to consider it everywhere else we're checking for RBAC as well, so I think its safe to follow that up in a separate issue if you'd like.

pkg/supportbundle/rbac.go
diamonwiggins: Can you clarify? I'm not sure I follow why moving it from here would make it more visible?

pkg/supportbundle/rbac.go
nvanthao: I just think passing empty `namespace` should not allowed as we are checking RBAC in both cluster-scope and namespace-scoped such as `get/pods`

pkg/supportbundle/rbac.go
nvanthao: I mean setting `resourceAttributesList` as a package constant would make it more discoverable as a configuration rather than hiding it inside `checkRemoteCollectorRBAC`. E.g. when we have new collector with additional permission.

On this, I also think we are missing permission `create pods/exec` for executing `collect` binary.

diamonwiggins: > On this, I also think we are missing permission create pods/exec for executing collect binary.

This PR is for the existing code which I don't think uses `exec` at the moment. 

> I mean setting resourceAttributesList as a package constant would make it more discoverable as a configuration rather than hiding it inside checkRemoteCollectorRBAC. E.g. when we have new collector with additional permission.

I don't have a super strong opinion here, but it doesn't feel like this list is going to be shared across multiple functions, so I'm not sure it needs to come out of this function.

pkg/supportbundle/rbac.go
diamonwiggins: Are you saying we should default to `default` for the namespace or fail if no namespace is provided? Right now, if the user doesn't provide a namespace when executing remote host collectors via the CLI, then we pretty much assume `""` for the namespace. 

This may be a larger thing we want to address in the project as I've noticed that the namespace in the kubeconfig context isn't used as the default, and it's not clear if we intend to do so or if we default to assuming cluster scoped access if you don't specify "-n" when running troubleshoot binaries. This would be another one of those issues that's a larger thing we need to revisit in the project as the behavior is the same for in-cluster collectors today. 

Let me know if that clears things up, or maybe i'm still not quite following what you're suggesting.

PR #1677, opened 2024-11-07
remove redundant tests

### LINKED ISSUE #1663: remove golangci-lint from CI
from Marc: 

> org wide, we are not recommending or using golangci-lint.  we’ve removed it from CMX and Vandoor

> golangci-lint does add additional checks fwiw, but we seem to be disabling and ignoring most of them (and rightly so IMO, they aren’t providing a TON of value.  for example, we have 100s of errors that we are ignoring a return value from a call.  after looking at all of them (literally all of them in vendor-api), they all would be replaced with _ =   While that’s more exact and specific, it not really making the product or code better.
And, i’d argue, that the Go language was pretty intentionally written to produce predictable code at Google. You can’t include an import that’s not used, you can’t have unused variables.  Why did they make it ok to ignore errors when you choose?  (I think it’s because it’s normal, and you should make that decision)

xavpaice:

Removes some test redundancy, and golangci-lint

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/build-test-deploy.yaml
+++ b/.github/workflows/build-test-deploy.yaml
@@ -27,25 +27,6 @@ jobs:
       - name: Fails in order to indicate that pull request needs to be marked as ready to review and unit tests workflow needs to pass.
         run: exit 1
 
-  test:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - uses: actions/setup-go@v5
-        with:
-          go-version: "1.22"
-      - name: setup env
-        run: |
-          echo "GOPATH=$(go env GOPATH)" >> $GITHUB_ENV
-          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
-        shell: bash
-      - name: Run linters
-        run: make install-golangci-lint lint
-      - name: Run tests
-        run: make test
-
   test-integration:
     runs-on: ubuntu-latest
     steps:
@@ -63,8 +44,8 @@ jobs:
       - uses: replicatedhq/action-k3s@main
         id: k3s
         with:
-          version: v1.24.1-k3s1
-
+          version: v1.31.2+k3s1
+      # test-integration includes unit tests
       - run: make test-integration
 
   ensure-schemas-are-generated:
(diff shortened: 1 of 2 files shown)


DexterYan (dismissed):

xavpaice:
The 3 tests waiting to be reported are the ones I've removed in this PR.

Makefile
```diff
@@ -237,18 +237,6 @@ scan:
 		--ignore-unfixed \
 		./
 
-.PHONY: lint
```
banjoh, line 240: Wasn't the aim not to run golangci-lint in CI? I think we want to keep this, for manually running it.

.github/workflows/build-test-deploy.yaml
```diff
@@ -27,25 +27,6 @@ jobs:
       - name: Fails in order to indicate that pull request needs to be marked as ready to review and unit tests workflow needs to pass.
         run: exit 1
 
-  test:
```
banjoh, line 30: We ought to just remove `Run linters`. Removing this action completely means not unit tests are run since we added `--tags=integration` to the `test-integration` make target.

banjoh (changes requested):

.github/workflows/build-test-deploy.yaml
banjoh, line 30: I think we ought to leave at least `make vet`. we could add it to an existing action such as `test-integration`

xavpaice: `go test` runs vet as step 1 regardless, iiuc.

Makefile
xavpaice: the message we got was that "Org wide, we are not recommending or using golangci-lint".  As per this, I've removed it entirely.  The results aren't that useful to us.

.github/workflows/build-test-deploy.yaml
xavpaice: https://pkg.go.dev/cmd/go/internal/test "As part of building a test binary, go test runs go vet on the package
and its test source files" - though it does run a subset. 

The Makefile that is called with `make test` includes the `make vet` as a dependency, I'll add the same to `make test-integration`.

banjoh (approved):
LGTM

banjoh:
I needed to update Branch protection rules -> Require status checks to pass before merging to reflect this PR changes

banjoh:
> The 3 tests waiting to be reported are the ones I've removed in this PR.

I needed to update `Branch protection rules` -> `Require status checks to pass before merging` to reflect this PR changes

PR #1687, opened 2024-11-12
fix(analyzer): add missing warning in outcome

### LINKED ISSUE #1652: Warn isn't working with regex analyzer
**Bug Description**

I'm trying to use **Warn** as an outcome for regex analyzer but it doesn't work, it shows fail instead

**Expected Behavior**

Just like the examples here https://troubleshoot.sh/docs/analyze/ , i expected that warn is supported as an option.

**Steps To Reproduce**

```yaml
kind: SupportBundle
metadata:
  name: test
spec:
  collectors:
    - clusterResources:
        exclude: true
    - clusterInfo:
        exclude: true
  hostCollectors:
    - run:
        collectorName: test
        command: "sh"
        args:
          - -c
          - |
            #!/bin/bash
            echo "testing"

  analyzers:
    - textAnalyze:
        checkName:  "test"
        fileName: host-collectors/run-host/test.txt
        regex: 'Check passed'
        outcomes:
          - pass:
              when: "true"
              message: "Check passed."
          - warn:
              when: "false"
              message: "There seems to be an issue with the test."
```

**Additional Context**

- Replicated Troubleshoot 0.85.0
- MacOS
- Sequoia 15.0.1

DexterYan:

- added missing outcome condition for warning
- added tests for both warning and fail

[sc-115535](https://app.shortcut.com/replicated/story/115535/regex-analyzer-warn-doesn-t-work)

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/analyze/text_analyze.go
+++ b/pkg/analyze/text_analyze.go
@@ -121,9 +121,11 @@ func analyzeRegexPattern(pattern string, collected []byte, outcomes []*troublesh
 		IconURI: "https://troubleshoot.sh/images/analyzer-icons/text-analyze.svg",
 	}
 
+	isMatch := re.MatchString(string(collected))
+
 	for _, outcome := range outcomes {
 		if outcome.Fail != nil {
-			isMatch := re.MatchString(string(collected))
+
 			// if the outcome.Fail.When is not set, default to false
 			if outcome.Fail.When == "" {
 				outcome.Fail.When = "false"
@@ -141,7 +143,6 @@ func analyzeRegexPattern(pattern string, collected []byte, outcomes []*troublesh
 				result.URI = outcome.Fail.URI
 			}
 		} else if outcome.Warn != nil {
-			isMatch := re.MatchString(string(collected))
 			// if the outcome.Warn.When is not set, default to false
 			if outcome.Warn.When == "" {
 				outcome.Warn.When = "false"
@@ -158,7 +159,6 @@ func analyzeRegexPattern(pattern string, collected []byte, outcomes []*troublesh
 				result.URI = outcome.Warn.URI
 			}
 		} else if outcome.Pass != nil {
-			isMatch := re.MatchString(string(collected))
 			// if the outcome.Pass.When is not set, default to true
 			if outcome.Pass.When == "" {
 				outcome.Pass.When = "true"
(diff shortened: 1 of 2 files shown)


pkg/analyze/text_analyze.go
```diff
@@ -115,54 +115,66 @@ func analyzeRegexPattern(pattern string, collected []byte, outcomes []*troublesh
...
-	passWhen := true
-	if passOutcome != nil && passOutcome.When != "" {
-		passWhen, err = strconv.ParseBool(passOutcome.When)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to process when statement: %s", passOutcome.When)
-		}
-	}
+	for _, outcome := range outcomes {
+		if outcome.Fail != nil {
+			isMatch := re.MatchString(string(collected))
```
nvanthao: this can be called only once before the loop

pkg/analyze/text_analyze.go
```diff
@@ -115,54 +115,66 @@ func analyzeRegexPattern(pattern string, collected []byte, outcomes []*troublesh
...
-		return &result, nil
-	}
+			if isMatch == failWhen {
+				result.IsFail = true
+				result.IsWarn = false
+				result.Message = outcome.Fail.Message
+				result.URI = outcome.Fail.URI
+			}
+		} else if outcome.Warn != nil {
+			isMatch := re.MatchString(string(collected))
```
nvanthao: this can be called only once before the loop

pkg/analyze/text_analyze.go
```diff
@@ -115,54 +115,66 @@ func analyzeRegexPattern(pattern string, collected []byte, outcomes []*troublesh
...
+				return nil, errors.Wrapf(err, "failed to process when statement: %s", outcome.Warn.When)
+			}
+
+			if isMatch == warnWhen {
+				result.IsWarn = true
+				result.Message = outcome.Warn.Message
+				result.URI = outcome.Warn.URI
+			}
+		} else if outcome.Pass != nil {
+			isMatch := re.MatchString(string(collected))
```
nvanthao: this can be called only once before the loop

nvanthao (changes requested):
just a nit otherwise LGTM

pkg/analyze/text_analyze.go
DexterYan: good catch! Fixed, Thx!

nvanthao (approved):
LGTM 👍

PR #1688, opened 2024-11-13
feat: ensure Copy collector run last

nvanthao:

Story: https://app.shortcut.com/replicated/story/115284/support-bundle-collector-to-support-order-of-execution
Demo: https://asciinema.org/a/okCHqJIFtb3KZQZqU7OT4KMdt

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/collector.go
+++ b/pkg/collect/collector.go
@@ -286,3 +286,20 @@ func DedupCollectors(allCollectors []*troubleshootv1beta2.Collect) []*troublesho
 	}
 	return finalCollectors
 }
+
+// Ensure Copy collectors are last in the list
+// This is because copy collectors are expected to copy files from other collectors such as Exec, RunPod, RunDaemonSet
+func EnsureCopyLast(allCollectors []Collector) []Collector {
+	otherCollectors := make([]Collector, 0)
+	copyCollectors := make([]Collector, 0)
+
+	for _, collector := range allCollectors {
+		if _, ok := collector.(*CollectCopy); ok {
+			copyCollectors = append(copyCollectors, collector)
+		} else {
+			otherCollectors = append(otherCollectors, collector)
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/collect/collector.go
```diff
@@ -286,3 +286,20 @@ func DedupCollectors(allCollectors []*troubleshootv1beta2.Collect) []*troublesho
 	}
 	return finalCollectors
 }
+
```
DexterYan, line 289: Looks good to me. However, I am thinking to use pointer approach to improve
```
// EnsureCopyLast ensures Copy collectors are last in the list
// This is because copy collectors are expected to copy files from other collectors such as Exec, RunPod, RunDaemonSet.
func EnsureCopyLast(allCollectors []Collector) []Collector {
	if len(allCollectors) <= 1 {
		return allCollectors // No need to rearrange if 0 or 1 collector
	}

	// Rearrange in-place using two pointers
	lastIndex := len(allCollectors) - 1
	for i := 0; i <= lastIndex; {
		if _, ok := allCollectors[i].(*CollectCopy); ok {
			// Move the Copy collector to the end
			allCollectors[i], allCollectors[lastIndex] = allCollectors[lastIndex], allCollectors[i]
			lastIndex-- // Shrink the range for non-Copy collectors
		} else {
			i++ // Proceed with non-Copy collectors
		}
	}

	return allCollectors
}
```

DexterYan (approved):

pkg/collect/collector.go
nvanthao: I was following other implementations for consistency for now, but definitely we can refactor it in the future if performance is a concern.

https://github.com/replicatedhq/troubleshoot/blob/67af4f0afa015451e34021d27d6acf59ae3d580c/pkg/collect/collector.go#L253-L265

PR #1697, opened 2024-11-28
feat: add support for matchExpressions when filtering for nodes

miguelvr:

Includes the ability to use a Kubernetes `matchExpressions` label selector for the `nodeResources` analyzer.

I wanted to write pre-flight checks to analyze nodes that contained a specific label and another that negated it. Currently, this is not possible with the `MatchLabel` selector. It would be possible with a `MatchExpressions` label selector by using the negation operators (e.g., `NotIn` / `DoesNotExist`)

## Checklist

- [x] Any documentation required has been added/updated. For changes to https://troubleshoot.sh/ create a PR [here](https://github.com/replicatedhq/troubleshoot.sh/pulls) - https://github.com/replicatedhq/troubleshoot.sh/pull/595

- [ ] Yes
- [x] No

### DIFF
--- a/config/crds/troubleshoot.sh_analyzers.yaml
+++ b/config/crds/troubleshoot.sh_analyzers.yaml
@@ -1186,6 +1186,36 @@ spec:
                               type: string
                             selector:
                               properties:
+                                matchExpressions:
+                                  items:
+                                    description: |-
+                                      A label selector requirement is a selector that contains values, a key, and an operator that
+                                      relates the key and values.
+                                    properties:
+                                      key:
+                                        description: key is the label key that the
+                                          selector applies to.
+                                        type: string
+                                      operator:
+                                        description: |-
+                                          operator represents a key's relationship to a set of values.
+                                          Valid operators are In, NotIn, Exists and DoesNotExist.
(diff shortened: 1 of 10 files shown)


nvanthao (changes requested):
Many thanks for the contribution @miguelvr!

The changes look good, I will approve it after we fix the failed Github Action.

When we introduce new field in Troubleshoot spec, we have to ensure the schemas are updated accordingly. Please run below command from the project root directory and commit the changes accordingly.

```bash
make schemas
make check-schemas
```

miguelvr:
@nvanthao `make generate` fails locally because it assumes a specific relative path directory structure (that I think it's used in CI, based on the git history)

it seems like the files are not generated to the repo anyway, so I skipped that command and committed everything else.

Let me know if there's something else I should fix on my end

nvanthao (approved):
Thank you for running the commands and for pointing out the issue with the relative paths. We'll address that separately to ensure smoother local workflows.

Since the schemas were successfully generated by other targets and everything else looks good, I'll go ahead and approve this PR. Many thanks for your contribution!

PR #1698, opened 2024-11-29
feat: container statuses analyzer

nvanthao:

This PR introduces a new container statuses analyzer that complements our existing [pod statuses analyzer](https://troubleshoot.sh/docs/analyze/cluster-pod-statuses/) functionality. While the pod statuses analyzer provides a high-level view of pod health, this enhancement allows deeper investigation into individual container states within pods, e.g. `OOMKilled`

Spec

```yaml
kind: Preflight
metadata:
  name: container-statuses
spec:
  analyzers:
    - clusterContainerStatuses:
        name: oom-detector
        namespaces:
          - default
        restartCount: 1
        outcomes:
          - fail:
              when: "== OOMKilled"
              message: "Container {{ .ContainerName }} from pod {{ .Namespace }}/{{ .PodName }} has OOMKilled"
              

```

## Checklist

- [ ] Yes
- [x] No


pkg/analyze/cluster_container_statuses.go
```diff
@@ -0,0 +1,252 @@
...
+			}
+			podsMatchedNamespace = append(podsMatchedNamespace, pods...)
+		} else {
+			podsMatchedNamespace = append(podsMatchedNamespace, podList.Items...)
+		}
+
+		// filter pods by container restart count
+		for _, pod := range podsMatchedNamespace {
+			for _, containerStatus := range pod.Status.ContainerStatuses {
+				if containerStatus.RestartCount < int32(a.analyzer.RestartCount) {
```
DexterYan: just wondering should we set a default RestartCount, like 5. Some customer may not set RestartCount, since they don't know and until saw it. I would suggest to have a default one if not specify it. Or set 0 to scan all of pods

pkg/analyze/cluster_container_statuses.go
```diff
@@ -0,0 +1,252 @@
...
+			if err := json.Unmarshal(fileContent, &pods); err != nil {
+				return nil, errors.Wrapf(err, "failed to unmarshal pods list for namespace %s", currentNamespace)
+			}
+			podsMatchedNamespace = append(podsMatchedNamespace, pods...)
+		} else {
+			podsMatchedNamespace = append(podsMatchedNamespace, podList.Items...)
+		}
+
+		// filter pods by container restart count
+		for _, pod := range podsMatchedNamespace {
```
DexterYan: I am thinking to split the loop inside of loop to two independent loops.
The first loop is finding all the `podsMatchedNamespace`,
the second loop is looping `podsMatchedNamespace` and check restart count
it may be more easy for reading

DexterYan (dismissed):

pkg/analyze/cluster_container_statuses.go
nvanthao: Currently, we have `RestartCount` default set to `0`, which ensures that all containers are monitored without missing any data. Setting a default value of 5 could inadvertently exclude containers that restart fewer times, potentially missing important information for users who rely on monitoring those instances. I would say customer can refer to document to set desired number of `RestartCount` based on their requirements. E.g. OOMKills detect

pkg/analyze/cluster_container_statuses.go
nvanthao: Many thanks! That is great idea, I have implemented the changes.

DexterYan (approved):

Issue #1162, opened 2023-05-17
Feature: add GPU capabilities to `nodeResources` analyzer


adamancini:
**Describe the rationale for the suggested feature.**

It would be good to be able to support preflights that want to check for GPU scheduling capability.  Off-hand, I don't know if this is visible in node metadata, but maybe could be detected from `containerd` configuration?  This might require a new collector or modifications to the nodeResources collector to detect if a node is capable of scheduling GPUS, and provide capacity/allocation similar to CPU, Memory, Disk.

**Describe the feature**

Not sure exactly which fields would be required, if `Allocatable` makes sense, but at a minimum something like:

`gpuCapacity` - # of GPUs available to a node

so you can write expressions like

```
- nodeResources:
        checkName: Total GPU Cores in the cluster is 4 or greater
        outcomes:
          - fail:
              when: "sum(gpuCapacity) < 4"
              message: The cluster must contain at least 4 GPUs
          - pass:
              message: There are at least 4 GPUs
```

adamancini:
Thinking through this a little bit there are a few places we can try to detect for GPU support

1. containerd configuration
2. `nvidia-smi` output
3. node metadata
4. run a no-op pod requesting GPUs and wait for successful exit

 
2: This can at least tell us if a GPU is installed, but not if Kubernetes is configured 
3: I don't know if the information we need is exposed in node metadata, requires research
1,4: I think are the best options since they are the closes to a functional test confirming that GPU workloads can be scheduled

diamonwiggins:
Adding some thoughts from a discussion in Slack, on the node metadata angle, we may be able to determine from `containerRuntimeVersion` at least when the `nvidia-container-runtime` for `containerd` is being used. Not sure if that'll be robust enough though. Imagine it could work for most cases.

from my local env:
```
    nodeInfo:
      architecture: amd64
      bootID: 81e20091-22da-4866-bfe4-a980057a1adf
      containerRuntimeVersion: containerd://1.5.9-k3s1
      kernelVersion: 5.15.49-linuxkit
      .....
```

chris-sanders:
Just chiming in on the number of GPU's question. I think this is going to be implementation specific. I don't know if we can measure it. I know for the Intel Gpu Plugin it can be configured to allow sharing gpu's or not. So the question isn't just how many gpu's are present but are they all fully scheduled.

[...]

PR #1701, opened 2024-12-02
feat: add support for GPU resources

miguelvr:

`nvidia.com/gpu` is a common resource for many AI applications, but currently, Troubleshoot does not allow users to check the availability of GPUs in their cluster. 

I've considered taking a more generic approach with a function `resourceName(<resource>)`, but decided against it because it would either require taking in the units as a parameter `resourceName(<resource>, <units>)` or default to something like `DecimalSI`, which could lead to errors for certain resources.

For simplicity, I decided to go with the more explicit approach. 

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/config/crds/troubleshoot.sh_analyzers.yaml
+++ b/config/crds/troubleshoot.sh_analyzers.yaml
@@ -1176,6 +1176,10 @@ spec:
                               type: string
                             ephemeralStorageCapacity:
                               type: string
+                            gpuAllocatable:
+                              type: string
+                            gpuCapacity:
+                              type: string
                             memoryAllocatable:
                               type: string
                             memoryCapacity:
--- a/config/crds/troubleshoot.sh_preflights.yaml
+++ b/config/crds/troubleshoot.sh_preflights.yaml
@@ -1176,6 +1176,10 @@ spec:
                               type: string
                             ephemeralStorageCapacity:
                               type: string
+                            gpuAllocatable:
+                              type: string
+                            gpuCapacity:
+                              type: string
                             memoryAllocatable:
                               type: string
                             memoryCapacity:
--- a/config/crds/troubleshoot.sh_supportbundles.yaml
+++ b/config/crds/troubleshoot.sh_supportbundles.yaml
@@ -1207,6 +1207,10 @@ spec:
                               type: string
                             ephemeralStorageCapacity:
                               type: string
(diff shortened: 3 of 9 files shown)


banjoh:
Thanks for your contribution. We'll review the PR

banjoh (changes requested):
You took the right direction, but I'm concerned that a GPU device only considers Nvidia hardware. Users who have `gpu.intel.com/i915` or `gpu.intel.com/xe` for example, cannot use this analyser.

banjoh:
> You took the right direction, but I'm concerned that a GPU device only considers Nvidia hardware. Users who have `gpu.intel.com/i915` or `gpu.intel.com/xe` for example, cannot use this analyser.

I don't yet have a clean suggestion that works, but I'd explore combining filters and outcomes.

xavpaice:
Internal tracking: https://app.shortcut.com/replicated/story/106618/in-cluster-collector-gpu-inventory

xavpaice:
> You took the right direction, but I'm concerned that a GPU device only considers Nvidia hardware. Users who have `gpu.intel.com/i915` or `gpu.intel.com/xe` for example, cannot use this analyser.

I think at this stage we can document the limitation and work with just Nvidia, expanding the scope in a future PR if there's demand.

miguelvr:
The more generic approach would be fetching by resource name. However it would have to be a function `resourceCapacity(nvidia.com/gpu, DecimalSI)`

This would cover any custom resource added by any device plugin.

I started on this, but I was having a hard time getting the regex to work as I wanted it to

xavpaice:
#1162 is related

chris-sanders:
@miguelvr we were reviewing this and issue #1162 from above and wanted to get your input on this approach.

https://docs.google.com/document/d/1thoEmuHpzf961cHHRreNjVB1rC_u8CVZYBCfxYNg8mY/edit?tab=t.0#heading=h.1q5hq5a3s115

I think that would be a bit more flexible to address the other open issue but want to confirm it would resolve your issue as well. I think it's what you're doing, but just a bit more flexible for the future. I'd like you to review that though and let us know what you think :)

banjoh:
@miguelvr I'll close this PR since this feature was implemented by https://github.com/replicatedhq/troubleshoot/pull/1708.

PR #1716, opened 2025-01-05
fix: [sc-116508] [Troubleshoot] support-bundle load-spec from URI field logic is fragile

nvanthao:
Story details: https://app.shortcut.com/replicated/story/116508
Demo: https://asciinema.org/a/jHxsEzwu3wnQ3xccJtkSp5JOb

This PR ensure that we fallback to original spec if the provided URI is malformed.


DexterYan (approved):

banjoh (approved):
LGTM

PR #1741, opened 2025-02-18
fix(host-preflights): builtin kernel modules file from wrong path

emosbaugh:

Path is wrong for some systems.

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/pkg/collect/host_kernel_modules.go
+++ b/pkg/collect/host_kernel_modules.go
@@ -128,6 +128,7 @@ func (l kernelModulesLoadable) collect(kernelRelease string) (map[string]KernelM
 	if _, err := os.Stat(kernelPath); os.IsNotExist(err) {
 		kernelPath = filepath.Join("/usr/lib/modules", kernelRelease)
 		if _, err := os.Stat(kernelPath); os.IsNotExist(err) {
+			kernelPath = filepath.Join("/lib/modules", kernelRelease)
 			klog.V(2).Infof("kernel modules are not loadable because path %q does not exist, assuming we are in a container", kernelPath)
 			return modules, nil
 		}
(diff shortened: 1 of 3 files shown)


pkg/collect/host_kernel_modules.go
```diff
@@ -228,7 +228,7 @@ func (l kernelModulesLoaded) collectBuiltin() (map[string]KernelModuleInfo, erro
 	}
 	kernel := strings.TrimSpace(string(out))
 
-	file, err := os.Open(fmt.Sprintf("/usr/lib/modules/%s/modules.builtin", kernel))
+	file, err := os.Open(fmt.Sprintf("/lib/modules/%s/modules.builtin", kernel))
```
squizzi: is it always going to consistently be this dir, no matter the distro?  Maybe we should have a list of possible locations and iterate through them `if os.IsNotExist`

squizzi: I think it _should be fine_ though.  It seems the convention is `/lib/modules`

emosbaugh: updated with a test

pkg/collect/host_kernel_modules.go
```diff
@@ -114,20 +121,16 @@ func (c *CollectHostKernelModules) Collect(progressChan chan<- interface{}) (map
...
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to determine kernel release")
-	}
-	kernel := strings.TrimSpace(string(out))
-
-	kernelPath := "/lib/modules/" + kernel
-
+	kernelPath := filepath.Join("/lib/modules", kernelRelease)
 	if _, err := os.Stat(kernelPath); os.IsNotExist(err) {
-		klog.V(2).Infof("modules are not loadable because kernel path %q does not exist, assuming we are in a container", kernelPath)
```
squizzi, line 129: nit: It might be worth logging in this instance that `/lib/modules` does not exist, checking `/usr/lib/modules` before the `assuming we are in container` log fires, because that makes it sound like we only ever checked `/usr/lib/modules`.

Or maybe update the `kernelPath` in `Infof` to be:

```go
klog.V(2).Infof("modules are not loadable because kernel path %q or %q do not exist, assuming we are in a container", kernelPathLib, kernelPathUsr)
```

squizzi (dismissed):
lgtm, just a nit on logging the directories checked.

pkg/collect/host_kernel_modules.go
emosbaugh: updated to print the expected path

laverya (approved):

PR #1751, opened 2025-03-05
feat: add CPU architecture check and template support

### LINKED ISSUE #1584: Add support for CPU architecture to cpu host collector/analyser

In some environments, someone may want to check what CPU architecture the system is running on so as to be able to perform preflight checks before deployments e.g an application that's not built for ARM may have a preflight failure for `arm64` architecture

**Describe the feature**

Extent https://troubleshoot.sh/docs/host-collect-analyze/cpu/ host collector/analyser
- Collect `KernelArch` value from https://pkg.go.dev/github.com/shirou/gopsutil/v3@v3.24.5/host#InfoStat and store in a new `machineArch` field. This value will be the output of `uname -m`
- Extend the analyser to allow checking `machineArch` values like so
```yaml
  hostAnalyzers:
    - cpu:
        checkName: "Check machine architecture"
        outcomes:
          - fail:
              when: "machineArch != x86_64"
              message: This server is not x86_64
          - fail:
              when: "machineArch == amd64"
              message: This server is amd64
```
- Add `Info.MachineArch` template to allow crafting outcome messages such as `{{ Info.MachineArch }} machine architecture is not supported`

One can use a combination of [run](https://troubleshoot.sh/docs/host-collect-analyze/run/) collector and [regex](https://troubleshoot.sh/docs/host-collect-analyze/regex/) analyser like so to achieve the same result. The message template bit would be missing though
```yaml
kind: SupportBundle
metadata:
  name: run
spec:
  hostCollectors:
    - run:
        collectorName: "uname-m"
        command: "sh"
        args: ["-c", "uname -m"]
  hostAnalyzers:
    - textAnalyze:
        checkName: "uname-m"
        file: "host-collectors/run-host/uname-m.txt"
        matchRegex: "x86_64|aarch64|arm64"
        outcomes:
          - pass:
              when: "pass"
              message: "Architecture is supported"
          - fail:
              message: "Architecture is not supported"
```

edgarlanting:

Adds template support for CPU architecture in analyzer messages, enabling dynamic insertion of the machine architecture value using `{{ Info.MachineArch }}`. This extends the existing CPU analyzer which already handles architecture checks through the `machineArch` comparison, making the output messages more informative and customizable.

The CPU analyzer continues to handle all existing functionality:
- CPU count checks (logical/physical)
- CPU flags verification
- Machine architecture validation
- Microarchitecture support checks

Fixes: #1584

## Checklist
- [x] New and existing tests pass locally with introduced changes
- [x] The commit message(s) follow conventional commits format: "feat: add CPU architecture template support"
- [x] Documentation updates not required as this extends existing functionality

- [ ] Yes
- [x] No


pkg/analyze/host_cpu.go
```diff
@@ -63,11 +63,21 @@ func (a *AnalyzeHostCPU) Analyze(
...
+	// Add template support
+	content := collectedContents[0].Data
+	cpuInfo := collect.CPUInfo{}
+	if err := json.Unmarshal(content, &cpuInfo); err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal cpu info")
+	}
 
+	for _, r := range results {
+		r.Message = strings.ReplaceAll(r.Message, "{{ Info.MachineArch }}", cpuInfo.MachineArch)
+	}
```
laverya: Would it be possible to actually template this instead of running a strings.ReplaceAll? I really hope that isn't a pattern in the rest of the project but if it is it's not one I'd really prefer to continue...

examples/test-cpu-arch.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: troubleshoot.sh/v1beta2
+kind: SupportBundle
+metadata:
```
laverya: instead of adding this to the root of the `examples` folder, could we put it in `examples/support-bundle`?

edgarlanting: @laverya yes, that is a valid point. I've moved this to `examples/support-bundle` in my next change.

edgarlanting:
@laverya I have made the suggested changes, however I will add the example file at a later moment

pkg/analyze/host_cpu.go
```diff
@@ -16,9 +17,8 @@ import (
 // ref: https://gitlab.com/x86-psABIs/x86-64-ABI
 // ref: https://developers.redhat.com/blog/2021/01/05/building-red-hat-enterprise-linux-9-for-the-x86-64-v2-microarchitecture-level
 var microarchs = map[string][]string{
-	"x86-64":    {"cmov", "cx8", "fpu", "fxsr", "mmx", "syscall", "sse", "sse2"},
```
laverya, line 19: any reason in particular this one was deleted?

pkg/analyze/host_cpu.go
```diff
@@ -27,47 +27,73 @@ type AnalyzeHostCPU struct {
...
+	findFiles getChildCollectedFileContents,
 ) ([]*AnalyzeResult, error) {
 	result := AnalyzeResult{Title: a.Title()}
 
-	// Use the generic function to collect both local and remote data
 	collectedContents, err := retrieveCollectedContents(
 		getCollectedFileContents,
-		collect.HostCPUPath,     // Local path
-		collect.NodeInfoBaseDir, // Remote base directory
-		collect.HostCPUFileName, // Remote file name
```
laverya: And why were these comments deleted?

pkg/analyze/host_cpu.go
```diff
@@ -61,13 +62,39 @@ func (a *AnalyzeHostCPU) Analyze(
 		return []*AnalyzeResult{&result}, err
 	}
 
+	content := collectedContents[0].Data
```
sgalsaleh, line 65: Why only check the first element? Isn't this supposed to be results per node? We should check all entries like we do in other places.

laverya: this is just for templating the output message, it isn't used for doing the check itself I think

laverya: that being said, let me take a look

laverya: ugh that does make the output message much less potentially useful

edgarlanting:
Closing this Pr in favor of a newer/better version of this change

PR #1833, opened 2025-08-31
bug: Respect provided kubeconfig in helm collector

Danil-Grigorev:

This change updates the Helm collector to properly use provided client configuration.

Previously, the collector used `actionConfig.Init(nil, ...)`, which relied on implicit defaults and did not respect the configured cluster context. This resulted in using server `https://localhost:8080` while performing operations on a host without `KUBECONFIG` or `~/.kube/config`. In a scenario, when the `support-bundle` was used as a library, this resulted in a failed collection for `helm`.

To fix this, a new `configGetter` type was introduced that implements `genericclioptions.RESTClientGetter`. Helm collector now passes `c.ClientConfig` through.

## Checklist

- [ ] Tests for the changes have been added (for bug fixes / features).
- [ ] The commit message(s) are informative and highlight any breaking changes.
- [ ] Any documentation required has been added/updated. For changes to https://troubleshoot.sh/ create a PR [here](https://github.com/replicatedhq/troubleshoot.sh/pulls).

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/helm.go
+++ b/pkg/collect/helm.go
@@ -10,9 +10,16 @@ import (
 
 	"github.com/pkg/errors"
 	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	"github.com/replicatedhq/troubleshoot/pkg/k8sutil"
 	"helm.sh/helm/v3/pkg/action"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/collect/helm.go
```diff
@@ -45,6 +49,32 @@ type VersionInfo struct {
...
 }
 
+type configGetter struct {
+	restConfig *rest.Config
+}
+
+// ToDiscoveryClient implements genericclioptions.RESTClientGetter.
+func (c configGetter) ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error) {
+	return nil, nil
+}
```
emosbaugh, line 67: I am unclear if this function is used by the helm client. were you able to confirm either way? it may be possible to implement it like so.

```suggestion
func (c configGetter) ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error) {
	discoveryClient, err := discovery.NewDiscoveryClientForConfig(c.restConfig)
	if err != nil {
		return nil, err
	}
	cached := memory.NewMemCacheClient(discoveryClient)
	return cached, nil
}
```

pkg/collect/helm.go
```diff
@@ -45,6 +49,32 @@ type VersionInfo struct {
...
+
+// ToRESTConfig implements genericclioptions.RESTClientGetter.
+func (c configGetter) ToRESTConfig() (*rest.Config, error) {
+	return c.restConfig, nil
+}
+
+// ToRESTMapper implements genericclioptions.RESTClientGetter.
+func (c configGetter) ToRESTMapper() (meta.RESTMapper, error) {
+	return nil, nil
+}
```
emosbaugh, line 82: same here

```suggestion
func (c configGetter) ToRESTMapper() (meta.RESTMapper, error) {
	discoveryClient, err := c.ToDiscoveryClient()
	if err != nil {
		return nil, err
	}
	mapper := restmapper.NewDeferredDiscoveryRESTMapper(discoveryClient)
	return mapper, nil
}
```

pkg/collect/helm.go
```diff
@@ -45,6 +49,32 @@ type VersionInfo struct {
...
+
+// ToRawKubeConfigLoader implements genericclioptions.RESTClientGetter.
+func (c configGetter) ToRawKubeConfigLoader() clientcmd.ClientConfig {
+	return nil
+}
```
emosbaugh, line 87: I see this is possible to get the clientcmd.ClientConfig with k8sutil. It may be best to pass this via args similar to ClientConfig via the opts structs ([here](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/collect/cli/run.go#L160), [here](https://github.com/replicatedhq/troubleshoot/blob/main/cmd/troubleshoot/cli/run.go#L180), [here](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/preflight/run.go#L355), and [here](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/preflight/run.go#L410)) although I do see collectors using the function [directly](https://github.com/replicatedhq/troubleshoot/blob/main/pkg/collect/copy_from_host.go#L73).

```suggestion
func (c configGetter) ToRawKubeConfigLoader() clientcmd.ClientConfig {
	return k8sutil.GetKubeconfig()
}
```

AmberAlston:
@Danil-Grigorev Making sure you saw the PR feedback above from @emosbaugh

pkg/collect/helm.go
Danil-Grigorev: I’ll check tomorrow, but from what I observed, only the implemented method was used. Helm operations there simply listed known resources, so it didn’t require discovery. Committed anyway, thanks for pointing that out.

emosbaugh (changes requested):
Looks like this is now failing to compile

emosbaugh (approved):

PR #1889, opened 2025-10-08
Remove duplicate goreleaser

marccampbell:
there are two goreleaser jobs running, which is... well... not the right way to hold the hammer


NoaheCampbell (approved):
looks good to me

PR #1935, opened 2025-11-26
Fix collector ordering: preserve order when grouping by type

xavpaice:
- Fix issue where EnsureClusterResourcesFirst ordering was lost when collectors were grouped by type into a map (Go maps have random iteration order)
- Preserve collector type order by tracking collectorTypeOrder slice as collectors are added to the map
- Apply fix to both pkg/preflight/collect.go and pkg/supportbundle/collect.go
- Add comprehensive tests to verify clusterResources runs first and relative order of other collectors is preserved
- Enhance EnsureClusterResourcesFirst tests with additional edge cases

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/collect/collect_test.go
+++ b/pkg/collect/collect_test.go
@@ -15,7 +15,7 @@ func Test_ensureClusterResourcesFirst(t *testing.T) {
 		list []*troubleshootv1beta2.Collect
 	}{
 		{
-			name: "Reorg OK",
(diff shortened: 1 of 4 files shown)


xavpaice:
note, this does not address the deduplication for clusterResources, we didn't do that originally because we're likely to have parameters.  That's a separate change if we want it.

emosbaugh (approved):

PR #1944, opened 2025-12-08
fix(discovery): handle partial results from ServerGroupsAndResources

adamancini:
## Summary

Fixes a bug in the `HasResource()` function where it returns immediately when `ServerGroupsAndResources()` returns an error, without checking if partial results are available. This violates the Kubernetes client-go contract.

## Problem

The cluster-resources collector fails to retrieve resources like PodDisruptionBudgets when unrelated API groups (e.g., custom.metrics.k8s.io, metrics.k8s.io) have discovery failures, resulting in "stale GroupVersion discovery" errors.

Example error:
```
"": "unable to retrieve the complete list of server APIs: custom.metrics.k8s.io/v1beta1: stale GroupVersion discovery: custom.metrics.k8s.io/v1beta1, external.metrics.k8s.io/v1beta1: stale GroupVersion discovery: external.metrics.k8s.io/v1beta1, metrics.k8s.io/v1beta1: stale GroupVersion discovery: metrics.k8s.io/v1beta1"
```

## Root Cause

The `HasResource()` function in `pkg/k8sutil/discovery/discovery.go` was returning immediately on any error without checking partial results:

```go
_, apiLists, err := dc.ServerGroupsAndResources()
if err != nil {
    return false, err  // ❌ BUG: Returns without checking apiLists
}
```

According to the [Kubernetes client-go documentation](https://github.com/kubernetes/client-go/blob/master/discovery/discovery_client.go):

> "The returned group and resource lists might be non-nil with partial results even in the case of non-nil error."

## Solution

Modified `HasResource()` to:
1. Check if `apiLists` is nil (total failure) - return error
2. If `apiLists` is non-nil (partial results available) - search for the requested resource
3. Return `true` if found, `false` with error if not found

This allows the troubleshoot collector to continue functioning even when some Kubernetes API groups have discovery failures.

### `pkg/k8sutil/discovery/discovery.go`
- Added null check for `apiLists` before processing
- Process partial results even when errors occur
- Return error along with false when resource not found
- Updated documentation

### `pkg/k8sutil/discovery/discovery_test.go`
- Added `TestHasResourceWithPartialDiscoveryFailure` with 8 test cases
- Created mock to simulate partial discovery failures
- Verified old implementation would fail 2/8 tests
- Verified new implementation passes all 11/11 tests

### Old Implementation (Buggy)
```
--- FAIL: TestHasResourceWithPartialDiscoveryFailure/resource_found_in_partial_results_with_discovery_error
    expected no error but got: unable to retrieve the complete list of server APIs
```
**Result**: 2 out of 8 tests FAILED ❌

### New Implementation (Fixed)
```
--- PASS: TestHasResourceWithPartialDiscoveryFailure (0.00s)
    --- PASS: TestHasResourceWithPartialDiscoveryFailure/resource_found_in_partial_results_with_discovery_error
    [... all 8 tests pass ...]
```
**Result**: All 11/11 tests PASSED ✅

## Impact

This fix allows collection of:
- PodDisruptionBudgets (policy/v1)
- CronJobs (batch/v1)
- Ingress (networking.k8s.io/v1)
- StorageClasses (storage.k8s.io/v1)
- PriorityClasses (scheduling.k8s.io/v1)
- CustomResourceDefinitions (apiextensions.k8s.io/v1)

...even when unrelated API groups like metrics-server have discovery issues.

## Related Issues

Similar issues in other projects:
- [linkerd/linkerd2#2780](https://github.com/linkerd/linkerd2/issues/2780)
- [kubernetes/client-go#1071](https://github.com/kubernetes/client-go/issues/1071)
- [kubevirt/kubevirt#9905](https://github.com/kubevirt/kubevirt/issues/9905)

### DIFF
--- a/pkg/k8sutil/discovery/discovery.go
+++ b/pkg/k8sutil/discovery/discovery.go
@@ -6,11 +6,15 @@ import (
 
 // HasResource takes an api version and a kind of a resource and checks if the resource
 // is supported by the k8s api server.
+// This function handles partial results from ServerGroupsAndResources(): "The returned group and resource lists might be non-nil with partial
+// results even in the case of non-nil error."
 func HasResource(dc discovery.DiscoveryInterface, apiVersion, kind string) (bool, error) {
 	_, apiLists, err := dc.ServerGroupsAndResources()
-	if err != nil {
+
+	if apiLists == nil {
 		return false, err
 	}
+
 	// Compare the resource api version and kind and find the resource.
 	for _, apiList := range apiLists {
 		if apiList.GroupVersion == apiVersion {
@@ -21,5 +25,6 @@ func HasResource(dc discovery.DiscoveryInterface, apiVersion, kind string) (bool
 			}
 		}
 	}
-	return false, nil
+
+	return false, err
 }
--- a/pkg/k8sutil/discovery/discovery_test.go
+++ b/pkg/k8sutil/discovery/discovery_test.go
@@ -159,7 +159,7 @@ func TestHasResourceWithPartialDiscoveryFailure(t *testing.T) {
 			description: "Should return (false, error) when apiLists is nil and discovery error exists",
 		},
 		{
-			name: "empty api resource list with discovery error",
+			name:            "empty api resource list with discovery error",
 			apiResourceList: []*metav1.APIResourceList{},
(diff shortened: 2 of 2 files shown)


bennyyang11 (approved):
looks good to me

scottrigby (approved):
nice

PR #2011, opened 2026-04-10
fix(supportbundle): ExtractLicenseFromBundle prefers license.json regardless of tar order

emosbaugh:
## Summary

- `ExtractLicenseFromBundle` was returning the wrong app slug (e.g. `kotsadm`) when `cluster-resources/license.json` existed in the bundle but was positioned later in the tar than a configmap that happened to contain a license ID pattern.
- Root cause: the tar is traversed sequentially; the old code returned immediately on the first license match, so whichever source appeared first in the tar won.
- Fix: collect both candidate results in a single pass, then prefer `license.json` over any configmap match at the end.

## Test plan

- [ ] `TestExtractLicenseFromBundle_PrefersLicenseJSONOverConfigmap` — synthetic bundle with configmap first, `license.json` second; asserts `license.json` values are returned
- [ ] `TestExtractLicenseFromBundle_FallsBackToConfigmap` — no `license.json`; asserts configmap fallback still works
- [ ] `TestExtractLicenseFromBundle_ReturnsEmptyWhenNotFound` — neither source has a license; asserts empty return
- [ ] `TestExtractLicenseFromBundle_RealBundle` — runs against the actual `support-bundle-2026-04-10T17_13_13.tar.gz` that triggered the bug; skipped in CI where bundle is absent

### DIFF
--- a/pkg/supportbundle/extract_license.go
+++ b/pkg/supportbundle/extract_license.go
@@ -34,6 +34,11 @@ func ExtractLicenseFromBundle(bundlePath string) (string, string, error) {
 
 	tarReader := tar.NewReader(gzReader)
 
+	// Collect results from both sources in a single pass; license.json takes priority
+	// regardless of its position in the tar, since configmaps often appear earlier.
+	var licenseJSONID, licenseJSONSlug string
+	var configmapID, configmapSlug string
+
 	for {
 		header, err := tarReader.Next()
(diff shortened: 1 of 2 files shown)


emosbaugh:
[sc-136293]

sgalsaleh (approved):

PR #2019, opened 2026-04-14
fix: record skipped collectors in bundle and improve CLI warnings

banjoh:

- Write `skipped-collectors.json` to the support bundle and preflight bundle recording which collectors were skipped (RBAC permissions or spec exclusion), with reason, errors, and timestamp
- Surface RBAC errors and skipped-collector messages as `klog.Warningf` in non-interactive mode so they're visible without `-v` flags
- Exec collector file naming now falls back to the resolved container name (from the pod spec) when `collectorName` is not set, instead of only checking `containerName` from the spec

Fixes: [sc-136406](https://app.shortcut.com/replicated/story/136406)

## Checklist

- [ ] Yes
- [x] No

### DIFF
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -172,9 +172,12 @@ func runTroubleshoot(v *viper.Viper, args []string) error {
 				case string:
 					if strings.Contains(msg, "skipping collector") {
 						klog.Warningf("Collecting support bundle: %s", msg)
+					} else {
+						klog.Infof("Collecting support bundle: %s", msg)
 					}
+				default:
+					klog.Infof("Collecting support bundle: %v", msg)
 				}
-				klog.Infof("Collecting support bundle: %v", msg)
 			}
 		}()
 	} else {
--- a/cmd/troubleshoot/cli/run.go
+++ b/cmd/troubleshoot/cli/run.go
@@ -166,6 +166,14 @@ func runTroubleshoot(v *viper.Viper, args []string) error {
 		go func() {
 			defer wg.Done()
 			for msg := range progressChan {
+				switch msg := msg.(type) {
+				case error:
+					klog.Warningf("Collecting support bundle: %v", msg)
+				case string:
+					if strings.Contains(msg, "skipping collector") {
+						klog.Warningf("Collecting support bundle: %s", msg)
+					}
+				}
 				klog.Infof("Collecting support bundle: %v", msg)
 			}
 		}()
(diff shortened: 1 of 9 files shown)


cmd/troubleshoot/cli/run.go
```diff
@@ -166,6 +166,14 @@ func runTroubleshoot(v *viper.Viper, args []string) error {
 		go func() {
 			defer wg.Done()
 			for msg := range progressChan {
+				switch msg := msg.(type) {
+				case error:
+					klog.Warningf("Collecting support bundle: %v", msg)
```
emosbaugh, line 171: will this now duplicate the message or does warning go to a different target that info? or is that the intention?

emosbaugh (dismissed):

cmd/troubleshoot/cli/run.go
banjoh: It will duplicate. I'll fix

emosbaugh (dismissed):

pkg/collect/collector.go
```diff
@@ -296,6 +298,32 @@ func DedupCollectors(allCollectors []*troubleshootv1beta2.Collect) []*troublesho
...
+	skippedJSON, err := json.Marshal(skipped)
+	if err != nil {
+		return
+	}
+	allCollectedData["skipped-collectors.json"] = skippedJSON
+	if bundlePath != "" {
+		if writeErr := os.MkdirAll(bundlePath, 0755); writeErr == nil {
+			_ = os.WriteFile(filepath.Join(bundlePath, "skipped-collectors.json"), skippedJSON, 0644)
+		}
+	}
```
laverya: This doesn't feel right
do we really write everything to disk and then compress it + cleanup temp afterwards?

banjoh: Good catch. I'll correct it

pkg/collect/collector.go
laverya: this needs an update too

pkg/collect/collector.go
```diff
@@ -296,6 +299,35 @@ func DedupCollectors(allCollectors []*troubleshootv1beta2.Collect) []*troublesho
...
+
+	// Either write to bundle path or memory
+	c := CollectorResult{}
+	if err := c.SaveResult(bundlePath, "skipped-collectors.json", bytes.NewReader(skippedJSON)); err != nil {
+		klog.Errorf("Failed to save skipped collectors: %v", err)
+	}
+
+	// Write to collected data to return downstream
+	maps.Copy(allCollectedData, c)
```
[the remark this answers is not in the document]
banjoh: We'll map copy when `c.SaveResult` succeeds

laverya (approved):

PR #2031, opened 2026-04-28
refactor: replace SPDY executor with WebSocket-first fallback executor

emosbaugh:
## Summary

- Replaces all \`remotecommand.NewSPDYExecutor\` call sites with a new \`k8sutil.NewFallbackExecutor\` helper that tries WebSocket first and falls back to SPDY when the server doesn't support it
- Deletes \`pkg/k8sutil/portforward.go\` — the \`PortForward()\` function was dead code (never called anywhere in the codebase)
- Fixes \`PodExecOptions.Stdout: false\` in exec, copy, and copy-from-host collectors — with the WebSocket executor the API server strictly respects this field and silently discards stdout when it's \`false\`; with SPDY the mismatch was harmless

## How it works

\`k8sutil.NewFallbackExecutor\` constructs a WebSocket primary executor (using \`"GET"\` per RFC 6455) and an SPDY secondary executor (using \`"POST"\`), wired together with \`remotecommand.NewFallbackExecutor\`. The fallback predicate covers both \`httpstream.IsUpgradeFailure\` and \`httpstream.IsHTTPSProxyError\`, matching kubectl's own pattern.

The method is not exposed as a parameter to prevent callers from accidentally passing \`"POST"\` (SPDY's method) to the WebSocket executor, which would silently fail.

## Test plan

- [x] \`make test\` passes
- [x] Manual E2E on k3s: exec collector captures stdout and stderr to distinct files
- [x] Support bundle collection completes successfully against live cluster


emosbaugh:
[sc-136858]

banjoh (approved):
LGTM

laverya (approved):

PR #2051, opened 2026-05-29
feat(ci): add Slack release notification via reusable workflow

kriscoleman:
## Summary

Adds a `notify` job to the release workflow that posts to Slack when a release tag is pushed, using the `replicatedhq/reusable-workflows/notify-release` reusable workflow.

- Runs in parallel with `goreleaser` (no `needs:`) so the team gets notified immediately when a release starts, not after it finishes
- Skipped for pre-release tags (e.g., `v1.0.0-rc1`)
- Posts a message with a compare URL between previous and current release
- Reusable workflow pinned to commit SHA for supply chain safety
- Requires `RELEASE_SLACK_WEBHOOK` secret

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -50,7 +50,7 @@ jobs:
   notify:
     needs: goreleaser
     if: ${{ !contains(github.ref_name, '-') }}
-    uses: replicatedhq/reusable-workflows/.github/workflows/notify-release.yml@main
+    uses: replicatedhq/reusable-workflows/.github/workflows/notify-release.yml@0432fb838fc83852b78be97a150b35991a85d56f # main
     with:
       tag: ${{ github.ref_name }}
     secrets:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,3 +46,12 @@ jobs:
         uses: rajatjindal/krew-release-bot@v0.0.51
         with:
           krew_template_file: deploy/krew/support-bundle.yaml
+
+  notify:
+    needs: goreleaser
+    if: ${{ !contains(github.ref_name, '-') }}
+    uses: replicatedhq/reusable-workflows/.github/workflows/notify-release.yml@main
+    with:
+      tag: ${{ github.ref_name }}
+    secrets:
+      slack_webhook: ${{ secrets.RELEASE_SLACK_WEBHOOK }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -46,3 +46,12 @@ jobs:
         uses: rajatjindal/krew-release-bot@v0.0.51
         with:
           krew_template_file: deploy/krew/support-bundle.yaml
+
+  notify:
+    needs: goreleaser
+    if: ${{ !contains(github.ref_name, '-') }}
+    uses: replicatedhq/reusable-workflows/.github/workflows/notify-release.yml@0432fb838fc83852b78be97a150b35991a85d56f # main
```
divolgin: `@main` if it's meant to be "main"

kriscoleman: this change was in response to https://github.com/replicatedhq/troubleshoot/pull/2051#discussion_r3326027540 

I can revert it back to `@main` if you want, but is it worth pinning?

.github/workflows/release.yaml
```diff
@@ -46,3 +46,11 @@ jobs:
...
         with:
           krew_template_file: deploy/krew/support-bundle.yaml
+
+  notify:
+    if: ${{ !contains(github.ref_name, '-') }}
+    uses: replicatedhq/reusable-workflows/.github/workflows/notify-release.yml@0432fb838fc83852b78be97a150b35991a85d56f # main
+    with:
+      tag: ${{ github.ref_name }}
+    secrets:
+      slack_webhook: ${{ secrets.RELEASE_SLACK_WEBHOOK }}
```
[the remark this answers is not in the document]
kriscoleman: I decided to notify when the release kicks off, not after.

.github/workflows/release.yaml
divolgin: > if it's meant to be "main"

The comment says "main". It's misleading because it will no longer be main after next commit.

divolgin: I'm in favor of versioning the workflows repo with tags. The last change, for example, moved a file. So pinning is needed if we can't guarantee backwards compatibility.

kriscoleman: makes sense to me! 
I updated to remove the `# main` comment and I support releasing tags for reusable-workflows next 👍

divolgin (approved):

PR #2054, opened 2026-06-03
fix(analyze/secret): guard nil deref when spec omits fail: outcome (#2053)

### LINKED ISSUE #2053: fix(analyze/secret): AnalyzeSecret panics with nil-pointer when spec has no fail: outcome and Secret is missing
## Summary

Hit this during a vendor click-install in CMX yesterday. The `secret:` analyzer in `pkg/analyze/secret.go` will SIGSEGV kotsadm if you write a Preflight spec with only `warn:` and `pass:` outcomes (no `fail:`) and the Secret being analyzed is absent. The panic happens at `secret.go:72` when `failOutcome.Fail.Message` is dereferenced.

## Affected versions

Confirmed at `v0.123.17`. The relevant function `analyzeSecret` is unchanged at HEAD as of 2026-06-03, so this still reproduces.

## Reproduction

Minimal Preflight spec:

```yaml
kind: Preflight
metadata:
  name: repro
spec:
  collectors:
    - secret:
        name: does-not-exist
        namespace: default
  analyzers:
    - secret:
        checkName: Optional Secret
        secretName: does-not-exist
        namespace: default
        outcomes:
          - warn:
              when: "notFound"
              message: "secret missing (warn)"
          - pass:
              message: "secret found"
```

Install it via a release on a channel and run through the KOTS UI click-install path (or the `preflight ...` CLI). The `does-not-exist` Secret is, well, not there. kotsadm crashes during the analyze phase, the UI hangs at "Gathering details about the cluster" near 100%, and the pod restarts losing preflight state.

Stack trace from a real install:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x39c5b8e]

goroutine 1875 [running]:
github.com/replicatedhq/troubleshoot/pkg/analyze.(*AnalyzeSecret).analyzeSecret(0xc002b2e0a0, 0xc002257030, 0xc000f7ca50)
	pkg/analyze/secret.go:72 +0x32e
github.com/replicatedhq/troubleshoot/pkg/analyze.(*AnalyzeSecret).Analyze(...)
	pkg/analyze/secret.go:29 +0x25
github.com/replicatedhq/troubleshoot/pkg/analyze.Analyze(...)
	pkg/analyze/analyzer.go:146 +0x629
github.com/replicatedhq/troubleshoot/pkg/analyze.AnalyzeLocal(...)
	pkg/analyze/download.go:43 +0x1bc
github.com/replicatedhq/kots/pkg/preflight.Execute(...)
	pkg/preflight/execute.go:169 +0x6be
```

## Code analysis

The bug lives in `analyzeSecret` around lines 63 to 72:

```go
var failOutcome *troubleshootv1beta2.Outcome
for _, outcome := range analyzer.Outcomes {
    if outcome.Fail != nil {
        failOutcome = outcome
    }
}

if !foundSecret.SecretExists {
    result.IsFail = true
    result.Message = failOutcome.Fail.Message   // nil deref
    result.URI = failOutcome.Fail.URI
}
```

The loop only captures `Fail` outcomes. If the spec has none, `failOutcome` stays nil and the deref at line 72 explodes. There is no guard.

There is also a quieter second issue. The function never reads `outcome.Warn` anywhere. So even after the panic is fixed, the warn-on-notFound idiom you would reach for does nothing. The analyzer is effectively `pass | fail`, not `pass | warn | fail`, and the docs on `Outcome` suggest otherwise.

## Prior art

This is the same defect shape as the `imagePullSecret` analyzer in #263. That one was filed in 2020 and fixed; the sibling bug in `AnalyzeSecret` slipped through.

## Suggested fix

Two pieces:

1. Guard the deref. Either synthesize a generic "Secret not found in namespace X" result when no Fail outcome is provided, or surface an analyzer-config error pointing at the missing `fail:`. The former matches how `AnalyzeImagePullSecret` was fixed.
2. Wire `outcome.Warn` into the analyzer. The warn case currently has no code path, so `warn: when: "notFound"` is impossible to express today.

Happy to send a PR if that helps. The fix is small.

## Impact

Caught during a CMX click-install for an active Replicated vendor during a CRE advisory engagement (Salt AI; the chart-side MR is `plai-app/ai-studio/salt-replicated!103`). Vendors writing what looks like a perfectly valid analyzer spec will see their preflight UI hang at 96 to 100 percent with no obvious cause, and their kotsadm pod will crash-loop until the spec is fixed or preflights are skipped. Worth a fix.

kriscoleman:
## Summary

Fixes #2053.

`pkg/analyze/secret.go` previously panicked with a SIGSEGV at `secret.go:72` when a Preflight spec defined no `fail:` outcome and the analyzed Secret was missing — the analyzer dereferenced a nil `failOutcome` to read `.Fail.Message`.

This PR restructures `analyzeSecret` to mirror the existing pattern in `pkg/analyze/image_pull_secret.go`:

- Collect `failOutcome` and `passOutcome` separately with non-nil checks (the secret analyzer only supports those two outcome types per <https://troubleshoot.sh/docs/analyze/secrets>).
- If the spec contains neither, return `nil, nil` and let the framework surface the missing-outcome error rather than fabricating a result.
- Assign each outcome's `Message`/`URI` inside the matching `IsPass` / `IsFail` branch so a passing result can never carry the fail outcome's message.
- Fill default messages at the end only when no configured outcome was used.
- `Analyze()` wrapper forwards a nil result through cleanly.

## Reproduction

Minimal Preflight spec from #2053:

```yaml
kind: Preflight
metadata:
  name: repro
spec:
  collectors:
    - secret:
        name: does-not-exist
        namespace: default
  analyzers:
    - secret:
        checkName: Optional Secret
        secretName: does-not-exist
        namespace: default
        outcomes:
          - pass:
              message: "secret found"
```

Before this PR: kotsadm panics during analyze, UI hangs at "Gathering details about the cluster", pod restarts and loses preflight state.
After: returns `IsFail` with the default message `Secret does-not-exist was not found in namespace default` (or the configured `fail:` outcome if present). No panic.

## Tests

New table-driven cases in `pkg/analyze/secret_test.go` covering:

- Not-found with no `fail:` outcome → falls back to default message (regression test for the original panic).
- Key-not-found with no `fail:` outcome → same.
- Spec with neither `fail:` nor `pass:` → `nil, nil`, framework surfaces the missing-outcome error.
- Secret found when only `fail:` was configured → uses the default pass message, **not** the fail outcome's message (regression test for a Cursor Bugbot finding mid-review).

All 9 secret-analyzer tests pass locally; existing happy paths unchanged.

## CI fix included

`validate-supportbundle-e2e-go` was racing `compile-preflight` because it downloaded the `preflight` artifact without declaring `compile-preflight` in `needs:`. Recent main runs happened to win the race; this PR's rebase hit the loss. Added the missing dependency.

## Files changed

- `pkg/analyze/secret.go` — restructure per analyzer contract; nil-guard on the dereference.
- `pkg/analyze/secret_test.go` — regression tests for the panic, the missing-outcome path, and the fail-message-leaking-into-pass case.
- `.github/workflows/build-test-deploy.yaml` — add `compile-preflight` to `validate-supportbundle-e2e-go`'s `needs:`.

### DIFF
--- a/pkg/analyze/secret.go
+++ b/pkg/analyze/secret.go
@@ -60,47 +60,34 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
 		IconURI: "https://troubleshoot.sh/images/analyzer-icons/secret.svg?w=13&h=16",
 	}
 
-	var failOutcome, warnOutcome, notFoundWarnOutcome *troubleshootv1beta2.SingleOutcome
+	var failOutcome *troubleshootv1beta2.SingleOutcome
 	for _, outcome := range analyzer.Outcomes {
 		if outcome.Fail != nil {
 			failOutcome = outcome.Fail
 		}
-		if outcome.Warn != nil {
-			warnOutcome = outcome.Warn
-			if outcome.Warn.When == "notFound" {
-				notFoundWarnOutcome = outcome.Warn
-			}
-		}
 	}
 
-	applyNotFound := func(defaultMessage string) {
-		switch {
-		case notFoundWarnOutcome != nil:
-			result.IsWarn = true
-			result.Message = notFoundWarnOutcome.Message
-			result.URI = notFoundWarnOutcome.URI
-		case failOutcome != nil:
-			result.IsFail = true
+	// The secret analyzer only supports fail (not found) and pass (found) outcomes
+	// per https://troubleshoot.sh/docs/analyze/secrets. When the spec omits fail,
+	// we still report not-found as IsFail with a default message rather than panic.
+	applyFail := func(defaultMessage string) {
+		result.IsFail = true
+		if failOutcome != nil {
 			result.Message = failOutcome.Message
 			result.URI = failOutcome.URI
-		case warnOutcome != nil:
-			result.IsWarn = true
-			result.Message = warnOutcome.Message
-			result.URI = warnOutcome.URI
-		default:
-			result.IsWarn = true
+		} else {
 			result.Message = defaultMessage
 		}
 	}
 
 	if !foundSecret.SecretExists {
-		applyNotFound(fmt.Sprintf("Secret %s was not found in namespace %s", analyzer.SecretName, analyzer.Namespace))
+		applyFail(fmt.Sprintf("Secret %s was not found in namespace %s", analyzer.SecretName, analyzer.Namespace))
 		return &result, nil
 	}
 
 	if analyzer.Key != "" {
 		if foundSecret.Key != analyzer.Key || !foundSecret.KeyExists {
-			applyNotFound(fmt.Sprintf("Key %s was not found in secret %s/%s", analyzer.Key, analyzer.Namespace, analyzer.SecretName))
+			applyFail(fmt.Sprintf("Key %s was not found in secret %s/%s", analyzer.Key, analyzer.Namespace, analyzer.SecretName))
 			return &result, nil
 		}
 	}
--- a/pkg/analyze/secret.go
+++ b/pkg/analyze/secret.go
@@ -60,27 +60,47 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
 		IconURI: "https://troubleshoot.sh/images/analyzer-icons/secret.svg?w=13&h=16",
 	}
 
-	var failOutcome *troubleshootv1beta2.Outcome
+	var failOutcome, warnOutcome, notFoundWarnOutcome *troubleshootv1beta2.SingleOutcome
 	for _, outcome := range analyzer.Outcomes {
 		if outcome.Fail != nil {
-			failOutcome = outcome
+			failOutcome = outcome.Fail
+		}
+		if outcome.Warn != nil {
+			warnOutcome = outcome.Warn
+			if outcome.Warn.When == "notFound" {
+				notFoundWarnOutcome = outcome.Warn
+			}
+		}
+	}
+
+	applyNotFound := func(defaultMessage string) {
+		switch {
+		case notFoundWarnOutcome != nil:
+			result.IsWarn = true
+			result.Message = notFoundWarnOutcome.Message
+			result.URI = notFoundWarnOutcome.URI
+		case failOutcome != nil:
+			result.IsFail = true
+			result.Message = failOutcome.Message
+			result.URI = failOutcome.URI
+		case warnOutcome != nil:
+			result.IsWarn = true
+			result.Message = warnOutcome.Message
+			result.URI = warnOutcome.URI
+		default:
+			result.IsWarn = true
+			result.Message = defaultMessage
 		}
 	}
 
 	if !foundSecret.SecretExists {
-		result.IsFail = true
-		result.Message = failOutcome.Fail.Message
-		result.URI = failOutcome.Fail.URI
-
+		applyNotFound(fmt.Sprintf("Secret %s was not found in namespace %s", analyzer.SecretName, analyzer.Namespace))
 		return &result, nil
 	}
 
 	if analyzer.Key != "" {
 		if foundSecret.Key != analyzer.Key || !foundSecret.KeyExists {
-			result.IsFail = true
-			result.Message = failOutcome.Fail.Message
-			result.URI = failOutcome.Fail.URI
-
+			applyNotFound(fmt.Sprintf("Key %s was not found in secret %s/%s", analyzer.Key, analyzer.Namespace, analyzer.SecretName))
 			return &result, nil
 		}
 	}
--- a/pkg/analyze/secret.go
+++ b/pkg/analyze/secret.go
@@ -30,6 +30,9 @@ func (a *AnalyzeSecret) Analyze(getFile getCollectedFileContents, findFiles getC
 	if err != nil {
 		return nil, err
 	}
+	if result == nil {
+		return nil, nil
+	}
 	result.Strict = a.analyzer.Strict.BoolOrDefaultFalse()
 	return []*AnalyzeResult{result}, nil
 }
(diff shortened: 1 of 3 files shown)


xavpaice:
The when field in the secret analyzer does nothing. It's parsed from YAML into the SingleOutcome struct, but the analyzer code in pkg/analyze/secret.go never reads or evaluates it. The logic is purely binary:
- If SecretExists is false → fail
- If a key is specified and KeyExists is false → fail
- Otherwise → pass

The string notFound is not a valid option for this analyzer. There is no conditional evaluation in the secret analyzer at all. Other analyzers (like nodeResources, textAnalyze) parse and evaluate when expressions, but the secret analyzer just looks up the collected JSON and returns the fail/pass message verbatim based on existence checks.
If you need conditional behavior (e.g., pass when secret is missing, or evaluate the actual value), you'd need to extend the analyzer or use a different analyzer type like textAnalyze or jsonCompare against the collected secret file.

The docs provide a clear example of how to use this analyzer, which only offers a pass/fail, no warn.  Is there a specific need for 'warn' as an outcome when a required secret or key is missing?

pkg/analyze/secret.go
```diff
@@ -60,27 +60,47 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
...
-	var failOutcome *troubleshootv1beta2.Outcome
+	var failOutcome, warnOutcome, notFoundWarnOutcome *troubleshootv1beta2.SingleOutcome
 	for _, outcome := range analyzer.Outcomes {
 		if outcome.Fail != nil {
-			failOutcome = outcome
+			failOutcome = outcome.Fail
+		}
+		if outcome.Warn != nil {
+			warnOutcome = outcome.Warn
+			if outcome.Warn.When == "notFound" {
```
banjoh: > The when attribute is not supported in the outcomes of this analyzer.

From https://troubleshoot.sh/docs/analyze/secrets

pkg/analyze/secret.go
```diff
@@ -60,27 +60,47 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
...
 	}
 
-	var failOutcome *troubleshootv1beta2.Outcome
+	var failOutcome, warnOutcome, notFoundWarnOutcome *troubleshootv1beta2.SingleOutcome
 	for _, outcome := range analyzer.Outcomes {
 		if outcome.Fail != nil {
-			failOutcome = outcome
+			failOutcome = outcome.Fail
+		}
+		if outcome.Warn != nil {
```
banjoh: I do not think this analyser supports "Warn". Its either a fail (not found) or a pass (found). We need to add this to the docs

banjoh (commented):
The change you want here is to check if a fail or pass outcome exists (non-nil checks). If the spec does not contain either do nothing. The framework will error if any of the analysers does not return an outcome.

See how its done https://github.com/replicatedhq/troubleshoot/blob/aad5f6ef0638c502c38ede690b7f41010b6d796b/pkg/analyze/image_pull_secret.go#L61

pkg/analyze/secret.go
kriscoleman: Good catch — you're right, the analyzer only supports `fail` and `pass`. I had the polecat introduce `warn` handling on the (mistaken) assumption it was a valid fallback. Dropped all the warn / `when: notFound` logic in 1304ae6e and collapsed the nil-fail case onto `IsFail` with a default message. Core nil-deref fix is preserved.

pkg/analyze/secret.go
kriscoleman: Right — `when:` is gone too. The refactor in 1304ae6e removes the entire warn/notFound branch since the analyzer's contract is fail-or-pass only.

kriscoleman:
Restructured per your review (94bc8682):

- Mirrors the `image_pull_secret.go` pattern — collect `failOutcome` + `passOutcome` with non-nil checks, default to `IsFail`, flip to `IsPass` on success, fill default messages at the end only if none were configured.
- Returns `nil, nil` when the spec contains neither `fail` nor `pass`, so the framework surfaces the missing-outcome error rather than the analyzer fabricating a result.
- `Analyze()` wrapper forwards a nil result through cleanly (previously would have nil-derefed when setting `.Strict`).
- Tests updated to match, plus a new case for the neither-outcome path. All 8 secret tests pass locally.

The original nil-deref repro (warn+pass only, missing secret) now lands in the IsFail branch with the spec's fail message if present, or the default "Secret X was not found in namespace Y" otherwise — no panic, no fabricated warn outcome.

pkg/analyze/secret.go
```diff
@@ -54,42 +57,53 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
...
-			result.Message = failOutcome.Fail.Message
-			result.URI = failOutcome.Fail.URI
-
-			return &result, nil
+	if secretFound {
+		result.IsFail = false
+		result.IsPass = true
+		if passOutcome != nil {
+			result.Message = passOutcome.Message
+			result.URI = passOutcome.URI
```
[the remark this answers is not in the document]
kriscoleman: Good catch — fixed in 97704de4. Moved both message assignments inside their respective IsPass/IsFail branches so a pass result can't carry the fail outcome's message. Added a regression test covering the exact scenario (fail-only spec + secret found → uses default pass message, not the fail message).

kriscoleman:
> The when field in the secret analyzer does nothing. It's parsed from YAML into the SingleOutcome struct, but the analyzer code in pkg/analyze/secret.go never reads or evaluates it. The logic is purely binary:
> 
>     * If SecretExists is false → fail
> 
>     * If a key is specified and KeyExists is false → fail
> 
>     * Otherwise → pass
> 
> 
> The string notFound is not a valid option for this analyzer. There is no conditional evaluation in the secret analyzer at all. Other analyzers (like nodeResources, textAnalyze) parse and evaluate when expressions, but the secret analyzer just looks up the collected JSON and returns the fail/pass message verbatim based on existence checks. If you need conditional behavior (e.g., pass when secret is missing, or evaluate the actual value), you'd need to extend the analyzer or use a different analyzer type like textAnalyze or jsonCompare against the collected secret file.
> 
> The docs provide a clear example of how to use this analyzer, which only offers a pass/fail, no warn. Is there a specific need for 'warn' as an outcome when a required secret or key is missing?

Yea, all good points. Supporting when here doesn't make much sense, if the secret is required. That's an oxymoron. I have updated the fix

pkg/analyze/secret.go
```diff
@@ -54,42 +57,53 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
...
+			result.URI = failOutcome.URI
 		}
 	}
 
-	result.IsPass = true
-	for _, outcome := range analyzer.Outcomes {
-		if outcome.Pass != nil {
-			result.Message = outcome.Pass.Message
-			result.URI = outcome.Pass.URI
+	if result.Message == "" {
```
banjoh: Most, if not all, analysers do not attempt to add a default message. If the message is empty, we assume its intentional.

kriscoleman: Good call, thanks @banjoh. Pulled the default-message fallback entirely in 48b7b88c. A configured outcome is now used verbatim (empty stays empty, URI-only preserved), and when the matched branch has no configured outcome we leave the message empty rather than fabricating one. Treating empty as intentional, consistent with the other analysers.

pkg/analyze/secret.go
```diff
@@ -54,42 +57,62 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
...
+		if outcome.Fail != nil {
+			failOutcome = outcome.Fail
+		}
+		if outcome.Pass != nil {
+			passOutcome = outcome.Pass
+		}
+	}
+	if failOutcome == nil && passOutcome == nil {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
kriscoleman: Addressed in 48b7b88c. When a spec defines neither a pass nor a fail outcome, `analyzeSecret` now returns an explicit error instead of `(nil, nil)`, so the `Analyze` wrapper surfaces the misconfiguration to the user instead of swallowing it into an empty result slice.

pkg/analyze/secret.go
```diff
@@ -54,42 +57,51 @@ func (a *AnalyzeSecret) analyzeSecret(analyzer *troubleshootv1beta2.AnalyzeSecre
...
+		if passOutcome != nil {
+			result.Message = passOutcome.Message
+			result.URI = passOutcome.URI
+		}
+	} else {
+		result.IsFail = true
+		if failOutcome != nil {
+			result.Message = failOutcome.Message
+			result.URI = failOutcome.URI
 		}
```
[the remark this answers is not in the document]
banjoh: This looks valid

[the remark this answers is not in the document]
kriscoleman: Good catch, fixed in a3a4262f. Restored the default diagnostic, but scoped to exactly the case you flagged: when the matched branch has no configured outcome pointer (a pass-only spec that hits the fail path, or a fail-only spec that passes). A configured outcome with an intentionally empty message (URI-only) is still preserved verbatim, so the two cases are no longer conflated.

pkg/analyze/secret.go
kriscoleman: @banjoh circling back here now that the follow-up thread landed. I split the difference in a3a4262f: a configured outcome with an empty message stays empty (your original point, so URI-only outcomes remain intentional), but a branch with no configured outcome at all now falls back to a default diagnostic instead of an empty message. So we distinguish "intentionally empty" from "no outcome for this branch."

banjoh (approved):

PR #2057, opened 2026-06-09
feat(redact): make MAX_CONCURRENT_REDACTORS configurable via env var (sc-138321)

kriscoleman:
## Summary

Make the redactor concurrency cap configurable via an environment variable. The hardcoded `MAX_CONCURRENT_REDACTORS = 10` in `pkg/collect/redact.go` becomes the *default*; operators running the standalone `support-bundle` binary on very large bundles can opt into a higher ceiling without a code change.

This is **additive and default-preserving** — no behavior change unless `TROUBLESHOOT_MAX_CONCURRENT_REDACTORS` is set.

## Why

- Shortcut: [sc-138321](https://app.shortcut.com/replicated/story/138321)

## Design decision

From team discussion:

- Env var with default of **10**. 
- **Not** a CLI flag in this PR — kotsadm consumers cannot easily pass CLI flags.
- **Not** GOMAXPROCS-based sizing in this PR — separate concern, and GOMAXPROCS often underestimates in containerized environments.

## Implementation

`pkg/collect/redact.go`:

- `DefaultMaxConcurrentRedactors = 10` (exported; replaces the old constant)
- `MaxConcurrentRedactorsEnvVar = \"TROUBLESHOOT_MAX_CONCURRENT_REDACTORS\"` (exported)
- `maxConcurrentRedactors() int` helper:
  - Reads the env var
  - Returns the default if unset / empty / non-numeric / `<= 0`
  - Emits a `klog.Warningf` for non-empty invalid input (don't be silent)
  - Emits a `klog.Infof` when the resolved value differs from the default (so operators see the override took effect)
- `RedactResult` now calls `maxConcurrentRedactors()` when sizing `limitCh`

## Tests

Table-driven cases in `pkg/collect/redact_test.go`:

- env unset → default
- env empty → default
- env `\"50\"` → 50 (override path)
- env `\"10\"` → default (no-op override)
- env `\"0\"` → default (warning)
- env `\"-3\"` → default (warning)
- env `\"potato\"` → default (warning)
- env `\"  4  \"` → default (warning — `strconv.Atoi` rejects padding)

```
go test ./pkg/collect/... -count=1   # PASS
go vet ./...                          # clean
go build -buildvcs=false ./...        # clean
```

## Out of scope (follow-ups)

- CLI flag wiring
- GOMAXPROCS-based dynamic sizing
- kotsadm env-var plumbing
- Changing the default of 10

### DIFF
--- a/pkg/collect/redact.go
+++ b/pkg/collect/redact.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


kriscoleman:
## Review round 1 — looks good, ready for review

Reviewed at `ea31f937`. Pulled the branch, ran the tests locally, walked the diff against the spec in tr-oaf.

**Verified against bead acceptance criteria:**

| Criterion | Status |
|---|---|
| Env var name (`TROUBLESHOOT_MAX_CONCURRENT_REDACTORS`) | ✓ Sensible Troubleshoot prefix |
| `strconv.Atoi` parsing | ✓ |
| Default (10) on unset / empty / non-numeric / zero / negative | ✓ Each path verified |
| `klog.Warningf` on non-empty invalid input | ✓ Fires on `0`, `-3`, `potato`, `"  4  "` |
| `klog.Infof` when override is active | ✓ Fires on `=50`; correctly suppressed when value resolves to default |
| Helper wired into `RedactResult` sizing | ✓ `limitCh` uses `maxConcurrentRedactors()` |
| Old constant replaced (no dangling identifier) | ✓ Only ref in repo was `redact.go` itself |
| Concurrency-safe (no shared mutable state) | ✓ Env read fresh per call |
| Tests cover all 6 failure modes + bonus | ✓ unset / empty / 50 / 10 / 0 / -3 / potato / `"  4  "` |
| Commit + PR body reference sc-138321 | ✓ |
| GoDoc on helper + constants | ✓ |
| Scope tight (only `redact.go` + `redact_test.go`) | ✓ |
| No CLI flag, no GOMAXPROCS sizing, no kotsadm changes | ✓ Held the line |

**Local verification:**
```
go test ./pkg/collect/ -run Test_maxConcurrentRedactors -count=1 -v
PASS (8/8 subtests, log output matches design)
```

**Non-blocking notes** (do not need to address before review):
- Test doesn't assert on klog output. Bead language was "*ideally* … if there's a clean way to capture klog output without making the test brittle" — klog→stderr capture is non-trivial and would add fragility for marginal value. Reasonable to defer.
- Test uses manual snapshot/restore of the env instead of `t.Setenv`. Inline comment justifies it (`t.Setenv` has no `Unsetenv` counterpart, would mask the "env unset" case). Defensible.

No blocking findings. Ready for human review and to flip out of draft when you are.

divolgin (approved):

PR #2058, opened 2026-06-11
fix(analyze): include matched file path in textAnalyze result titles for multi-file globs

kriscoleman:
### Bug report

**Symptom.** When a `textAnalyze` analyzer uses a glob `fileName` to analyze the per-node output of a [`runDaemonSet`](https://troubleshoot.sh/docs/collect/run-daemonset/) collector, the analyzer fires once per matched file as designed, but every result carries an identical title and outcome message. On a 3-node cluster, a per-node check renders as three indistinguishable rows in `preflight` output and in any downstream UI that displays analyzer results. Operators can't tell which node a failing verdict belongs to without opening the bundle and inspecting file names.

Reproduction (any multi-node cluster):

```yaml
kind: Preflight
metadata:
  name: per-node-check
spec:
  collectors:
    - runDaemonSet:
        name: swap-check
        namespace: default
        podSpec:
          containers:
            - name: swap-check
              image: busybox:1.36
              command: ["sh", "-c"]
              args: ["awk 'NR>1 {sum+=$3} END {print \"total_swap_kb=\"sum+0}' /proc/swaps"]
  analyzers:
    - textAnalyze:
        checkName: "Swap disabled on every node"
        fileName: swap-check/*.log
        regex: 'total_swap_kb=0'
        outcomes:
          - pass:
              when: "true"
              message: "Swap is disabled on this node"
          - fail:
              when: "false"
              message: "Swap is enabled on this node"
```

Expected: each result identifies the file (node) it came from.
Actual: N identical results.

**Root cause.** `analyzeTextAnalyze` iterates the glob-matched `map[string][]byte` by value and discards the keys, so the matched file path never reaches the `AnalyzeResult`, and `AnalyzeResult` has no other field that could carry it. Nothing downstream (CLI human/JSON/YAML output, or UIs that render analyzer results) can display what it never receives.

### Fix

When a glob matches more than one file:

- Append the distinguishing part of the matched path to each result title: `Swap disabled on every node (node-a.log)`. The directory prefix shared by all matches is trimmed, because at preflight runtime the map keys can be absolute paths into a temp dir, and only the differing tail identifies the file. For nested globs (e.g. exec collector output at `collector/namespace/pod/file.txt`), the preserved tail keeps the distinguishing segments (`namespace/pod/file.txt`).
- Iterate matches in sorted order so multi-file results render deterministically (map iteration order made result ordering random before).

Single-file matches keep their exact current title, so existing specs that match one file see no change.

This pairs well with `runDaemonSet`: per-node preflights were already easy to collect, and now each node's verdict is identifiable at a glance.

### Verification

Before/after on a 3-node k3s cluster, same spec as above:

```
# released preflight
PASS: Swap disabled on every node
PASS: Swap disabled on every node
PASS: Swap disabled on every node

# with this change
PASS: Swap disabled on every node (3dcb85a11.log)
PASS: Swap disabled on every node (3dcb85a12.log)
PASS: Swap disabled on every node (3dcb85a13.log)
```

(`runDaemonSet` names each log after its node, so the node name lands in the title.)

Unit tests: added a multi-file glob case (`regex`) asserting one titled result per file, and a multi-file `regexGroups` case asserting titles plus templated messages. Existing multi-file test expectations were updated to the new titles; their diff doubles as a record of the behavior change.

## Checklist

Happy to follow up with a troubleshoot.sh docs PR noting the title behavior for glob matches if you'd like one.

- [ ] Yes
- [x] No

Titles change only when a glob matches multiple files, where the previous titles were identical duplicates that couldn't be referenced individually anyway. Single-file results are byte-for-byte unchanged.

### DIFF
--- a/pkg/analyze/text_analyze.go
+++ b/pkg/analyze/text_analyze.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -79,9 +80,31 @@ func analyzeTextAnalyze(
 
 	results := []*AnalyzeResult{}
 
+	// Iterate matched files in sorted order so multi-file results are
+	// deterministic, and include the matched file path in each result title
+	// when more than one file matched. Without this, a glob fileName (e.g.
+	// the per-node output of a runDaemonSet collector) produces N identical
+	// results that cannot be told apart in any rendered report. The shared
+	// directory prefix is trimmed because the matched keys can be absolute
+	// paths into a temp dir at runtime; only the distinguishing part of the
+	// path belongs in the title.
+	fileNames := make([]string, 0, len(collected))
+	for fileName := range collected {
+		fileNames = append(fileNames, fileName)
+	}
+	sort.Strings(fileNames)
+
+	sharedPrefix := commonDirPrefix(fileNames)
+	resultTitle := func(fileName string) string {
+		if len(collected) > 1 {
+			return fmt.Sprintf("%s (%s)", title, strings.TrimPrefix(fileName, sharedPrefix))
+		}
+		return title
+	}
+
 	if analyzer.RegexPattern != "" {
-		for _, fileContents := range collected {
-			result, err := analyzeRegexPattern(analyzer.RegexPattern, fileContents, analyzer.Outcomes, title)
+		for _, fileName := range fileNames {
+			result, err := analyzeRegexPattern(analyzer.RegexPattern, collected[fileName], analyzer.Outcomes, resultTitle(fileName))
 			if err != nil {
 				return nil, err
 			}
@@ -92,8 +115,8 @@ func analyzeTextAnalyze(
 	}
 
 	if analyzer.RegexGroups != "" {
-		for _, fileContents := range collected {
-			result, err := analyzeRegexGroups(analyzer.RegexGroups, fileContents, analyzer.Outcomes, title)
+		for _, fileName := range fileNames {
+			result, err := analyzeRegexGroups(analyzer.RegexGroups, collected[fileName], analyzer.Outcomes, resultTitle(fileName))
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


banjoh (changes requested):
I'm not sure this is the direction we should take to address this issue. Wouldn't it be better to use templating capability available in `textAnalyzer`?

Using the spec below you will get this results

```yaml
kind: HostPreflight
metadata:
  name: per-node-check
spec:
  collectors:
    - run:
        collectorName: swap-check
        command: sh
        args:
          - -c
          - 'echo hostname=$(hostname); awk ''NR>1 {sum+=$3} END {print "total_swap_kb="sum+0}'' /proc/swaps'
  analyzers:
    - textAnalyze:
        checkName: "Swap disabled on this node"
        fileName: host-collectors/run-host/swap-check.txt
        regexGroups: '(?s)hostname=(?P<Hostname>\S+).*total_swap_kb=(?P<Swap>\d+)'
        outcomes:
          - pass:
              when: "Swap == 0"
              message: "Swap is disabled on node {{ .Hostname }}"
          - fail:
              when: "Swap > 0"
              message: "Swap is enabled on node {{ .Hostname }} (total_swap_kb={{ .Swap }})"
```

```sh
preflight --interactive=false spec.yaml
[swap-check] Running collector...

   --- PASS Swap disabled on this node
      --- Swap is disabled on node 303decca
--- PASS   per-node-check
PASS

Saving preflight bundle to preflightbundle-2026-06-26T16_06_04.tar.gz
```

You should get the same result using this in-cluster spec

```yaml
kind: Preflight
metadata:
  name: per-node-check
spec:
  collectors:
    - runDaemonSet:
        name: swap-check
        namespace: default
        podSpec:
          containers:
            - name: swap-check
              image: busybox:1.36
              command: ["sh", "-c"]
              args: ["echo hostname=\$(hostname); awk 'NR>1 {sum+=\$3} END {print \"total_swap_kb=\"sum+0}' /proc/swaps"]
  analyzers:
    - textAnalyze:
        checkName: "Swap disabled on every node"
        fileName: swap-check/*.log
        regexGroups: '(?s)hostname=(?P<Hostname>\S+).*total_swap_kb=(?P<Swap>\d+)'
        outcomes:
          - pass:
              when: "Swap == 0"
              message: "Swap is disabled on node {{ .Hostname }}"
          - fail:
              when: "Swap > 0"
              message: "Swap is enabled on node {{ .Hostname }} (total_swap_kb={{ .Swap }})"
```

kriscoleman:
closing as Evans has a good point about templating, and also suggested we could even template the title

PR #2066, opened 2026-07-07
add factory workflow

xavpaice:

Add automation for Elasticclaw to run dependency updates.

## Checklist

- [ ] Yes
- [ ] No

### DIFF
--- a/.factory-workflows/dependency-update.yaml
+++ b/.factory-workflows/dependency-update.yaml
@@ -1,0 +1,202 @@
+schema_version: v1
+name: dependency-update-troubleshoot
+enabled: true
+
+trigger:
+  cron:
+    schedule: "0 21 * * *"
+    timezone: "America/Chicago"
+    overlap_policy: skip
+    timeout: 2h
+
+enable_manual_trigger: true
+
+tags: ["cron", "dependencies"]
+color: cyan
+
+stages:
+  - id: start
+    label: Start
+    entry: true
+    on_enter:
+      inject: |
+        Workflow task: run the Go dependency update maintenance workflow for Troubleshoot.
+
+        This is an unattended dependency maintenance workflow. Do not ask for
+        permission to start. Immediately say exactly [PROCEED] so the workflow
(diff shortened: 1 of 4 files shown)


.factory-workflows/dependency-update.yaml
```diff
@@ -0,0 +1,202 @@
...
+    overlap_policy: skip
+    timeout: 2h
+
+enable_manual_trigger: true
+
+tags: ["cron", "dependencies"]
+color: cyan
+
+stages:
+  - id: start
```
nvanthao, line 18: I'm curious why do we need both `start` and `working` stage, they seems overlap to me.

.factory-workflows/dependency-update.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+            if ! command -v go >/dev/null 2>&1; then
+              nix profile install nixpkgs#go_1_26 || nix-env -iA nixpkgs.go_1_26
+            fi
+
+            go_path="$(command -v go)"
+            go_version="$(go version)"
+            printf "{\"status\":\"ready\",\"go_path\":\"%s\",\"go_version\":\"%s\"}\n" "$go_path" "$go_version"
+          '
+        output: go_toolchain
```
nvanthao, line 52: is this output being used in any where?

nvanthao (commented):
I have some comments after going through the docs https://www.elasticclaw.ai/docs/workflows but not sure how to test this workflow to see if it is deterministic. Also, we should update why this is preferred over Dependabot.

.factory-workflows/dependency-update.yaml
xavpaice: the agent receives it, it's there for context

.factory-workflows/dependency-update.yaml
xavpaice: I've updated the overlap a bit, the split probably isn't necessary but it does make it clear what each step is for

nvanthao (approved):

PR #2071, opened 2026-07-09
Add per-collector subcommands to the collect binary

sgalsaleh:
## What
Adds `collect http|postgres|mysql|mssql|redis` subcommands — each runs a single
collector and prints its native result JSON to stdout.

## Why
Lets a single connectivity collector run inside a Pod via the troubleshoot image
(e.g. as a `runPod` collector), so the check executes from within the cluster
instead of wherever the CLI runs. Currently these collectors can only run as part
of a full spec.

## Notes
- One cobra subcommand per collector, each with its own flags/help mapping to that
  collector's spec fields (`mysql` adds `--parameters`; `mssql` is `--uri`-only as
  it has no TLS field; the others share `--uri` + TLS flags).
- Runs the existing `Collect()` in memory and prints the JSON; the k8s client
  (only used for TLS-from-Secret) matches the preflight/support-bundle path
  (QPS/Burst/UserAgent).
- Verified each subcommand against live http/postgres/mysql/redis targets.

## Example
Running the `postgres` check from inside the cluster via a `runPod` collector:

```yaml
collectors:
  - runPod:
      name: postgres-check
      namespace: default
      podSpec:
        restartPolicy: Never
        containers:
          - name: check
            image: proxy.replicated.com/library/troubleshoot:0.130.0
            command: ["collect", "postgres", "--uri", "postgres://user:pass@pg:5432/app?sslmode=disable"]
analyzers:
  - textAnalyze:
      checkName: Postgres reachable from cluster
      collectorName: postgres-check
      fileName: "*.log"
      regex: '"isConnected":true'
      outcomes:
        - pass: { when: "true", message: "connected to postgres" }
        - fail: { when: "false", message: "cannot connect to postgres" }
```

### DIFF
--- a/cmd/collect/cli/http.go
+++ b/cmd/collect/cli/http.go
@@ -1,0 +1,79 @@
+package cli
+
+import (
+	"fmt"
+	"strings"
+
+	troubleshootv1beta2 "github.com/replicatedhq/troubleshoot/pkg/apis/troubleshoot/v1beta2"
+	"github.com/replicatedhq/troubleshoot/pkg/collect"
+	"github.com/spf13/cobra"
+)
+
+// HTTPCmd runs the http collector against a single endpoint and prints the
+// native result JSON ({"response":{"status":...}} or {"error":{...}}).
+func HTTPCmd() *cobra.Command {
+	var (
+		method   string
+		url      string
+		headers  map[string]string
+		body     string
+		timeout  string
+		proxy    string
+		insecure bool
+		caCert   string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "http",
+		Short: "Run the http collector against an endpoint",
+		Long: `Run the http collector: issue an HTTP request and print the collector result as JSON.
+
+The result contains the response status, body and headers, or an error object.
+
+Examples:
+  collect http --url http://myapp.default.svc:8080/healthz
+  collect http --method POST --url https://api.internal/ping \
+      --header 'Content-Type=application/json' --body '{"ping":true}'`,
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var tls *troubleshootv1beta2.TLSParams
+			if caCert != "" {
+				tls = &troubleshootv1beta2.TLSParams{CACert: caCert}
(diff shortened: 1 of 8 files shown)


banjoh (dismissed):
LGTM

cmd/collect/cli/postgres.go
```diff
@@ -0,0 +1,88 @@
...
+				URI:           uri,
+			}
+
+			var (
+				client kubernetes.Interface
+				cfg    *rest.Config
+			)
+			if skipVerify || caCert != "" || clientCert != "" || clientKey != "" || secretName != "" {
+				tls := &troubleshootv1beta2.TLSParams{
+					SkipVerify: skipVerify,
```
divolgin, line 52: Doesn't look like CollectPostgres uses this

cmd/collect/cli/http.go
```diff
@@ -0,0 +1,79 @@
...
+
+	f := cmd.Flags()
+	f.StringVar(&url, "url", "", "request URL (required)")
+	cmd.MarkFlagRequired("url")
+	f.StringVar(&method, "method", "GET", "HTTP method: GET, POST, or PUT")
+	f.StringToStringVar(&headers, "header", nil, "request header as key=value (repeatable)")
+	f.StringVar(&body, "body", "", "request body (POST/PUT only)")
+	f.StringVar(&timeout, "timeout", "", "request timeout, e.g. 15s (empty = no timeout)")
+	f.StringVar(&proxy, "proxy", "", "proxy URL to use for the request")
+	f.BoolVar(&insecure, "insecure-skip-verify", false, "do not verify the server's TLS certificate")
```
divolgin, line 75: this flag is called `tls-skip-verify` in other commands

divolgin (dismissed):

cmd/collect/cli/http.go
sgalsaleh: yeah that's intentional. it matches what the field is named in the spec / schema: https://troubleshoot.sh/docs/collect/http#insecureskipverify-boolean-optional

divolgin (approved):

PR #2101, opened 2026-08-06
fix(preflight): redact local host preflight collector output

### LINKED ISSUE #2100: Host preflight run collectors leak full process environment (incl. credentials) into the bundle, unredacted
## Summary

Every host `run` collector's captured environment is written to the bundle completely unredacted. `HostRunInfo.Env` (`pkg/collect/host_run.go`) is set to `cmd.Env` — which, absent `ignoreParentEnvs`/`inheritEnvs`, is the full `os.Environ()` of the process running `preflight` — and marshaled verbatim to `<collectorName>-info.json`.

This is exploitable in practice: Embedded Cluster installers commonly export `HTTP_PROXY`/`HTTPS_PROXY` (with embedded Basic Auth credentials, e.g. `http://user:pass@proxy:3128`) into the environment before running host preflights. That credential — plus every other env var the installer process holds — ends up verbatim in a file customers routinely attach to support tickets.

## Root cause

`CollectHostWithContext` (`pkg/preflight/collect.go`) is the *only* collection path in the codebase that never runs collected data through the redaction engine:

- In-cluster support bundles call `collect.RedactResult` (`pkg/supportbundle/collect.go`).
- Remote (SSH) host preflight collectors set `Redact: true` on `RemoteCollector`, which triggers `RedactResult` via `RunCollectorSync` (`pkg/preflight/collect.go` `CollectRemoteWithContext`, `pkg/collect/remote_collector.go`).
- Local host preflight collectors (`CollectHostWithContext`) call neither. The built-in default redactors — including the existing "connection strings with username and password" URL-userinfo pattern that would otherwise catch exactly this case — never run.

Neither spec-level knob fixes it: `ignoreParentEnvs: true` empties the captured env (which also defeats any legitimate proxy-credential check that *needs* those vars to run), and `inheritEnvs` still lands the selected values in `runInfo.Env` unredacted.

## Repro

```go
hostPreflight := &troubleshootv1beta2.HostPreflight{
    Spec: troubleshootv1beta2.HostPreflightSpec{
        Collectors: []*troubleshootv1beta2.HostCollect{
            {
                HostRun: &troubleshootv1beta2.HostRun{
                    HostCollectorMeta: troubleshootv1beta2.HostCollectorMeta{CollectorName: "proxy-credential-check"},
                    Command:           "sh",
                    Args:              []string{"-c", "echo ok"},
                    IgnoreParentEnvs:  true,
                    Env:               []string{"HTTPS_PROXY=http://alice:sw0rdfish@proxy.example.com:3128"},
                },
            },
        },
    },
}
_, _ = CollectHostWithContext(context.Background(), CollectOpts{BundlePath: bundlePath, ProgressChan: ch}, hostPreflight)
```

`host-collectors/run-host/proxy-credential-check-info.json` contains `"env":["...","HTTPS_PROXY=http://alice:sw0rdfish@proxy.example.com:3128"]` verbatim — no `***HIDDEN***` mask.

## Suggested fix

Wire `CollectHostWithContext` through `collect.RedactResult` the same way `CollectRemoteWithContext` and `pkg/supportbundle/collect.go` already do, so the built-in default redactors (and any custom `Redactor` spec, once/if one is threaded through the `preflight` CLI) apply to local host preflight output the same as every other collection path.

I have a fix + regression test ready and will open a PR against this issue.

RangerRick:
Fixes #2100

## Summary

`CollectHostWithContext` (`pkg/preflight/collect.go`) was the only collection path in the codebase that never ran collected data through the redaction engine before writing it to the bundle:

- In-cluster support bundles call `collect.RedactResult` (`pkg/supportbundle/collect.go`).
- Remote (SSH) host preflight collectors set `Redact: true` on `RemoteCollector`, which triggers `RedactResult` via `RunCollectorSync` (`CollectRemoteWithContext`).
- Local host preflight collectors called neither.

In practice this means a `run` collector's captured environment (`HostRunInfo.Env` in `pkg/collect/host_run.go`, set from `cmd.Env`) lands in `<collectorName>-info.json` completely unredacted — including credentials such as a proxy URL with embedded Basic Auth (`HTTPS_PROXY=http://user:pass@proxy:3128`), which installers commonly export into the environment before running host preflights.

## Fix

Wire `CollectHostWithContext` through `collect.RedactResult` the same way `CollectRemoteWithContext` and `pkg/supportbundle/collect.go` already do. This applies the existing built-in default redactors (including the "connection strings with username and password" URL-userinfo pattern) to all local host preflight collector output, not just `run` collectors.

## Test plan

- Added `TestCollectHostWithContext_RedactsSensitiveEnvValues` (`pkg/preflight/collect_test.go`), which runs a `run` host collector with a credential-bearing env var and asserts the persisted `-info.json` no longer contains the credential in cleartext.
- Confirmed the test fails against the pre-fix code (reproduces the leak) and passes after the fix.
- `make test` passes locally (fmt, vet, generate, full unit suite).

### DIFF
--- a/pkg/preflight/collect.go
+++ b/pkg/preflight/collect.go
@@ -145,6 +145,15 @@ func CollectHostWithContext(
 		span.End()
 	}
 
+	// Local host preflight collectors are the only collection path (cluster,
+	// remote host, in-cluster support bundle) that skipped redaction entirely.
+	// A `run` collector's captured environment in particular can carry
+	// credentials verbatim (e.g. HTTPS_PROXY with embedded Basic Auth) into
+	// the bundle. See https://github.com/replicatedhq/troubleshoot/issues/2100.
+	if err := collect.RedactResult(opts.BundlePath, collect.CollectorResult(allCollectedData), nil); err != nil {
+		return nil, errors.Wrap(err, "failed to redact host collector results")
+	}
+
 	// The values of map entries will contain the collected data in bytes if the data was not stored to disk
 	collectResult.AllCollectedData = allCollectedData
 
--- a/pkg/preflight/collect_test.go
+++ b/pkg/preflight/collect_test.go
@@ -1,6 +1,9 @@
 package preflight
 
 import (
+	"context"
+	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


xavpaice (approved):
Many thanks for the contribution, approved.

PR #2102, opened 2026-08-10
fix(preflight): add tracing span and align error handling for host collector redaction

xavpaice:
Follow-up to #2101.

#2101 added redaction for local host preflight collectors, which was the only collection path that skipped redaction entirely. This PR addresses the review feedback that was not included before #2101 was merged:

1. **Adds an OpenTelemetry span around the redaction step**, matching the tracing already present for host and in-cluster collectors in the support-bundle path.
2. **Aligns the error handling**: when redaction fails, "CollectHostWithContext" now returns the unredacted "collectResult" along with the error, matching the behavior of "remote_collector.RunCollectorSync" and "supportbundle.runHostCollectors". This prevents SDK consumers from seeing inconsistent return behavior across collection paths.

The regression test added in #2101 ("TestCollectHostWithContext_RedactsSensitiveEnvValues") continues to verify that "run" host collector environment variables are redacted before being written to the bundle.

### DIFF
--- a/pkg/preflight/collect.go
+++ b/pkg/preflight/collect.go
@@ -150,9 +150,15 @@ func CollectHostWithContext(
 	// A `run` collector's captured environment in particular can carry
 	// credentials verbatim (e.g. HTTPS_PROXY with embedded Basic Auth) into
 	// the bundle. See https://github.com/replicatedhq/troubleshoot/issues/2100.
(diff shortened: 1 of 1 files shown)


xavpaice:
Good catch — fixed. `collectResult.AllCollectedData` is now assigned before the redaction call, so if redaction fails the returned result still contains the unredacted collected data, consistent with the other collection paths.

banjoh (approved):

PR #2103, opened 2026-08-10
Contain support bundle tar extraction within the destination directory

bunlongheng:
While looking at how "analyze" pulls apart a downloaded support bundle, I noticed ExtractTroubleshootBundle in pkg/analyze/download.go joins each tar entry name straight onto the destination dir and writes it out, with no check that the result stays inside that dir. A bundle with entries like ../../etc/something (or a symlink pointing outside) would let the archive write files anywhere the process can reach. Since bundles are usually generated by an end user and then handed to someone else to analyze, that felt worth closing off.

This adds a small containment check before anything is written: the resolved path (and the symlink target) has to live under destDir, otherwise it errors out. Normal bundles are unaffected. Build and gofmt are clean. Happy to adjust the wording or add a test if you prefer.

### DIFF
--- a/pkg/analyze/download.go
+++ b/pkg/analyze/download.go
@@ -8,6 +8,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	getter "github.com/hashicorp/go-getter"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


divolgin:
The PR’s security cases should be added to the existing `pkg/analyze/download_test.go`, ideally as a new table-driven `TestExtractTroubleshootBundle`. It can construct small gzip/tar archives in memory and call the function directly.

Useful cases:

- Regular file within the destination succeeds.
- Directory within the destination succeeds.
- `../outside` regular-file entry is rejected.
- `../outside` directory entry is rejected.
- In-root symlink target succeeds.
- `../../outside` symlink target is rejected.
- A sibling path sharing the destination prefix, such as destination `/tmp/bundle` and entry resolving to `/tmp/bundle-escape`, is rejected.

There are examples of creating gzip/tar test fixtures in `pkg/supportbundle/parse_test.go` and `pkg/supportbundle/extract_license_test.go`. No new test file should be necessary.

bunlongheng:
Thanks for the detailed guidance, @divolgin - the case list made this easy to get right. Added a table-driven `TestExtractTroubleshootBundle` to the existing `pkg/analyze/download_test.go` (no new test file), building small gzip/tar archives in memory and calling the function directly, following the fixture style from `pkg/supportbundle/extract_license_test.go`.

All the cases you listed are covered:

- Regular file within the destination succeeds
- Directory within the destination succeeds
- `../outside` regular-file entry is rejected
- `../outside` directory entry is rejected
- In-root symlink target succeeds
- `../../outside` symlink target is rejected
- Sibling path sharing the destination prefix (destination `<tmp>/bundle`, entry resolving to `<tmp>/bundle-escape`) is rejected - this one verifies the containment check is separator-aware, not a naive string prefix match

The rejection cases also assert that nothing from the offending entry was created outside the destination. All 7 pass locally with `go test ./pkg/analyze/ -run TestExtractTroubleshootBundle -v`. Let me know if you'd like any additional cases.

PR #2105, opened 2026-08-11
fix(runpod): strip pod spec before saving collected output

xavpaice:
The runPod collector was saving the full Kubernetes Pod object (including Spec) to the bundle. Because the PodSpec can contain env vars, commands, args, volumes, and imagePullSecrets, this leaked sensitive data into the collected output.

Changes:
- In savePodDetails, copy the fetched pod and zero its Spec before JSON-marshaling.
- Refactor savePodDetails to accept kubernetes.Interface and drop the unused clientConfig parameter so the function can be unit-tested with the fake client.

Verification:
- Added TestSavePodDetails_StripsPodSpec.
- make test and make build pass.

This also affects the deprecated run collector and goldpinger, both of which delegate to CollectRunPod. Goldpinger only reads pod.Name and pod.Status.ContainerStatuses, so it remains compatible.

### DIFF
--- a/pkg/collect/run_pod.go
+++ b/pkg/collect/run_pod.go
@@ -69,7 +69,7 @@ func (c *CollectRunPod) Collect(progressChan chan<- interface{}) (result Collect
 		if err != nil {
 			return
 		}
-		result, err = savePodDetails(ctx, client, result, c.BundlePath, c.ClientConfig, pod, c.Collector)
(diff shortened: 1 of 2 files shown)


banjoh (approved):

gilday:
Should this zero out `ObjectMeta.annotations` and `.labels` in addition to the Spec?

PR #2111, opened 2026-08-19
feat(supportbundle): support secret/... redactor URIs

xavpaice:
## Summary

Adds support for `secret/...` URIs to `LoadRedactorSpec`, matching the existing support in `LoadSupportBundleSpec`.

## Changes

- `pkg/supportbundle/load.go`: `LoadRedactorSpec` now accepts `secret/<namespace>/<secret-name>[/<data-key>]` URIs, with a default data key of `redactor-spec`.
- `pkg/supportbundle/load.go`: added a package-level `loadFromSecret` hook so the new code path can be unit-tested without requiring a cluster.
- `pkg/supportbundle/load_test.go`: added `TestLoadRedactorSpec` covering default and custom data keys, as well as validation for too few/too many path components.

## Motivation

KOTS stores rendered redactor specs in Kubernetes ConfigMaps today and passes `configmap/...` URIs to `kubectl support-bundle --redactors=...`. To move those redactor specs to Secrets, the Troubleshoot library and CLI need to be able to load redactor specs from `secret/...` URIs.

## Related

- KOTS story: https://app.shortcut.com/replicated/story/139291

### DIFF
--- a/pkg/supportbundle/load.go
+++ b/pkg/supportbundle/load.go
@@ -20,6 +20,9 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// loadFromSecret is a package-level hook so tests can stub out cluster access.
+var loadFromSecret = specs.LoadFromSecret
+
 // GetSupportBundleFromURI downloads and parses a support bundle from a URI and returns a SupportBundle object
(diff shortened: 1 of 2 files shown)


kriscoleman (approved):
nice

PR #2116, opened 2026-08-20
fix(cluster-resources): stop emitting duplicate unredacted YAML copy

xavpaice:
## Problem

`clusterResources` writes every custom resource twice: once as JSON and once as YAML under `cluster-resources/custom-resources/`. The built-in redactors are authored for JSON shapes, so the YAML copy is left unredacted. Adding a vendor-side YAML redactor is also a footgun because YAML literal blocks (e.g. PEM keys) cannot be safely masked with line-oriented redactors.

## Fix

- Stop writing the duplicate YAML file in `storeCustomResource`; emit only JSON. JSON is valid YAML, so YAML-aware analyzers can still parse the file.
- Convert the built-in kurl installer redactors to target the JSON file with scoped regex redactors so they continue to apply after the YAML copy is removed.

## Compatibility

No built-in analyzers require the YAML copy:
- Velero analyzers already use `*.json` globs.
- The AI agents\' CRD/yaml-compare paths accept either `.json` or `.yaml`.
- `yaml-compare` uses `yaml.Unmarshal`, which parses JSON.

Vendor specs that hardcode a `.yaml` filename under `cluster-resources/custom-resources/` will need to switch to `.json`.

## Testing

- `go test ./pkg/collect/... ./pkg/redact/...`
- `make test`
- `make build`
- Added `Test_RedactKurlInstallerJSON` to verify the new scoped redactors.

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -2294,23 +2294,18 @@ func mutatingWebhookConfigurations(ctx context.Context, client kubernetes.Interf
 	return b, nil
 }
 
-// storeCustomResource stores a custom resource as JSON and YAML
-// We use both formats for backwards compatibility. This way we
-// avoid breaking existing tools and analysers that already rely on
-// the YAML format.
+// storeCustomResource stores a custom resource as JSON only.
+// JSON is valid YAML, so any analyzer expecting YAML can consume the JSON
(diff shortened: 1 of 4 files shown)


xavpaice:
Closing to re-open from the same-repo branch (replicatedhq/fix/remove-custom-resource-yaml).

PR #2117, opened 2026-08-20
fix(cluster-resources): stop emitting duplicate unredacted YAML copy

xavpaice:
## Problem

clusterResources writes every custom resource twice: once as JSON and once as YAML under cluster-resources/custom-resources/. The built-in redactors are authored for JSON shapes, so the YAML copy is left unredacted. Adding a vendor-side YAML redactor is also a footgun because YAML literal blocks (e.g. PEM keys) cannot be safely masked with line-oriented redactors.

## Fix

- Stop writing a duplicate YAML file in storeCustomResource; emit only JSON.
- Add a .yaml symlink pointing at each JSON file for backward compatibility. Analyzers that expect YAML can still read the file, and the symlink is redacted through the JSON copy.
- Convert the built-in kurl installer redactors to target the JSON file with scoped regex redactors so they continue to apply. The selector now matches both the cluster-scoped file (installers.cluster.kurl.sh.json) and the namespaced pattern (installers.cluster.kurl.sh/*.json).

## Compatibility

No built-in analyzers require the old YAML file:
- Velero analyzers already use *.json globs.
- The AI agents CRD/yaml-compare paths accept either .json or .yaml.
- yaml-compare uses yaml.Unmarshal, which parses JSON.

Vendor specs that hardcode a .yaml filename under cluster-resources/custom-resources/ will continue to work because the symlink is present. The content will be JSON, which is valid YAML.

## Testing

- go test ./pkg/collect/... ./pkg/redact/...
- make test
- make build
- Added cluster-scoped and namespaced path cases to Test_RedactKurlInstallerJSON.

[sc-139314](https://app.shortcut.com/replicated/story/139314)

### DIFF
--- a/pkg/collect/cluster_resources.go
+++ b/pkg/collect/cluster_resources.go
@@ -294,7 +294,18 @@ func (c *CollectClusterResources) Collect(progressChan chan<- interface{}) (Coll
 	// crs
 	customResources, crErrors := crs(ctx, dynamicClient, client, c.ClientConfig, namespaceNames)
 	for k, v := range customResources {
-		output.SaveResult(c.BundlePath, path.Join(constants.CLUSTER_RESOURCES_DIR, constants.CLUSTER_RESOURCES_CUSTOM_RESOURCES, k), bytes.NewBuffer(v))
+		jsonPath := path.Join(constants.CLUSTER_RESOURCES_DIR, constants.CLUSTER_RESOURCES_CUSTOM_RESOURCES, k)
+		output.SaveResult(c.BundlePath, jsonPath, bytes.NewBuffer(v))
+
+		// Keep a YAML symlink for backward compatibility. It points at the JSON
+		// file, so any analyzer that expects YAML can still read it, and the
+		// content is redacted through the JSON copy.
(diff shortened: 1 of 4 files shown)


xavpaice:
Updated the PR to address the review/test issues:

1. **Backward compatibility:** Every custom-resource JSON file now has a corresponding .yaml symlink pointing at it. This keeps the bundle file list identical for any analyzer or regression baseline that expects the old YAML copy. The symlink is archived as a relative symlink, and the existing redaction code already resolves symlinks to redact the target JSON file.

2. **Greptile review:** The kurl installer redactor now matches both forms:
   - cluster-scoped: cluster-resources/custom-resources/installers.cluster.kurl.sh.json
   - namespaced: cluster-resources/custom-resources/installers.cluster.kurl.sh/*.json

3. **Tests:** Added a cluster-scoped path case to Test_RedactKurlInstallerJSON and confirmed go test ./pkg/collect/... ./pkg/redact/... passes.

xavpaice:
Shortcut story for this work: https://app.shortcut.com/replicated/story/139314

divolgin (approved):

PR #2125, opened 2026-09-04
fix(redactors): mask connection strings without a trailing database name

sgalsaleh:

The built-in redactors that mask credentials in a database URI only match when a trailing `/db-name` is present, so these reached support bundles unredacted:

- `postgres://user:pass@host:5432` — and any URI ending at host/port or a bare `/` (`mysql://`, `redis://`, `amqp://…:5672/`)
- `user:pass@tcp(host:3306)` — the Go MySQL DSN form

Two changes in `pkg/redact/redact.go`:

1. The URI credential redactor was scheme-locked to `https?|ftp`; generalized it to any scheme. This rule never required a path, so it covers every no-dbname form. Its mask/host/port groups and name are unchanged, so `http`/`ftp` output is identical.
2. Added a companion redactor for `user:pass@tcp(host:port)` with no trailing `/db-name`.

The two fuller redactors (which also mask host and database name) are untouched, so URIs *with* a database name redact byte-for-byte as before. As with the existing `http`/`ftp` rule, the no-dbname form masks user and password and leaves the hostname.

Making the `/dbname` tail optional instead doesn't work: the substitution string is built once per line from `re.SubexpNames()`, so a `mask` group emits `***HIDDEN***` even when it didn't participate, leaving a phantom mask where the path would be.

New `Test_DefaultRedactors_ConnectionStrings` runs the real default chain over each no-dbname variant, the with-dbname regressions, and strings that must not be redacted. `Test_Redactors` `wantRedactionsLen` 43 → 44: output is identical (`JSONEq` passes), but the postgres line is now attributed to two redactors at −6 chars each instead of one at −12.

Fixes: [sc-139337](https://app.shortcut.com/replicated/story/139337)

## Checklist

- [x] Any documentation required has been added/updated. (troubleshoot.sh lists redactor names only; names are unchanged)

- [ ] Yes
- [x] No

More content is masked than before — credentials in non-http URI schemes without a database name. Hostnames and non-credential URLs are unaffected.

### DIFF
--- a/pkg/redact/redact.go
+++ b/pkg/redact/redact.go
@@ -314,10 +314,11 @@ func getRedactors(path string) ([]Redactor, error) {
 			},
 			name: "Redact database connection strings that contain username and password",
 		},
-		// user:password@tcp(host:3309), with no trailing /db-name
+		// user:password@tcp(host:3309), with no trailing /db-name.
+		// the password runs to the @, so that slashes and colons in it are masked too.
 		{
 			regex: LineRedactor{
-				regex: `(?P<mask>[^:\"\/\s]*){1}(:)(?P<mask>[^:\"\/\s]*){1}(@tcp\()`,
+				regex: `(?P<mask>[^:\"\/\s]*){1}(:)(?P<mask>[^@\"\s]*){1}(@tcp\()`,
 				scan:  `@tcp`,
 			},
 			name: "Redact database connection strings that contain username and password",
--- a/pkg/redact/redact.go
+++ b/pkg/redact/redact.go
@@ -294,12 +294,15 @@ func getRedactors(path string) ([]Redactor, error) {
 			},
 			name: "Redact values for environment variables with names beginning with 'user'",
 		},
-		// connection strings with username and password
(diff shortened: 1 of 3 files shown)


pkg/redact/redact.go
```diff
@@ -311,6 +314,14 @@ func getRedactors(path string) ([]Redactor, error) {
 			},
 			name: "Redact database connection strings that contain username and password",
 		},
+		// user:password@tcp(host:3309), with no trailing /db-name
+		{
+			regex: LineRedactor{
+				regex: `(?P<mask>[^:\"\/\s]*){1}(:)(?P<mask>[^:\"\/\s]*){1}(@tcp\()`,
+				scan:  `@tcp`,
```
banjoh: How about `tcp6`? For completeness you can also add `unix` though I do not know whether that's ever used in k8s environments

sgalsaleh: addressed

banjoh (dismissed):

divolgin (approved):

PR #2131, opened 2026-09-11
Add SLSA provenance attestations for CLI releases

osc135:
## Summary

- grant the release job the permissions required to mint and publish keyless GitHub/Sigstore attestations
- generate SLSA build provenance for all preflight and support-bundle release archives
- publish the generated Sigstore bundle as a versioned GitHub release asset with `gh release upload`
- document keyless verification with `cosign verify-blob-attestation`, including the expected workflow, tag, issuer, and predicate type

## Testing

- `git diff --check`
- parsed `.github/workflows/release.yaml` as YAML
- syntax-checked the `gh release upload` shell commands
- confirmed the installed GitHub CLI supports `gh release upload --clobber`

The attestation and release upload steps can only run end-to-end from a tag-triggered release after this change is merged.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,17 +19,17 @@ jobs:
       artifact-metadata: write
     steps:
       - name: Checkout
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+      - uses: actions/setup-go@v7
         with:
           go-version: '1.26'
           check-latest: true
 
       - name: Install Cosign
-        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+        uses: sigstore/cosign-installer@v4.1.2
         with:
           cosign-release: 'v3.1.3'
 
@@ -40,7 +40,7 @@ jobs:
           COSIGN_PASSWORD: ${{ secrets.COSIGN_PASSWORD }}
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
+        uses: goreleaser/goreleaser-action@v7
         with:
           version: "v2.12.3"
           args: release --clean --config deploy/.goreleaser.yaml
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -12,19 +12,24 @@ concurrency:
...
     runs-on: troubleshoot_release
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
+      artifact-metadata: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
divolgin: `checkout@v7` is the right way to pin these

osc135: im curious (happy to make the change of course) but why should we pin to `checkout@v7` instead of the sha?

.github/workflows/release.yaml
```diff
@@ -35,22 +40,41 @@ jobs:
...
+          BUNDLE_PATH: ${{ steps.attest_cli.outputs.bundle-path }}
+        run: |
+          bundle="dist/troubleshoot_${GITHUB_REF_NAME}_provenance.sigstore.json"
+          cp "${BUNDLE_PATH}" "${bundle}"
+          gh release upload "${GITHUB_REF_NAME}" "${bundle}" --clobber
+
       - name: Update new preflight version in krew-index
         if: ${{ !contains(github.ref_name, '-') }}
-        uses: rajatjindal/krew-release-bot@v0.0.51
+        uses: rajatjindal/krew-release-bot@c970b8a8f6dbc2f2285a26e3ae160903b87002c3 # v0.0.51
```
divolgin: no need to change this

.github/workflows/release.yaml
```diff
@@ -35,22 +40,41 @@ jobs:
...
+        uses: goreleaser/goreleaser-action@f06c13b6b1a9625abc9e6e439d9c05a8f2190e94 # v7.2.3
         with:
           version: "v2.12.3"
           args: release --clean --config deploy/.goreleaser.yaml
         env:
           GITHUB_TOKEN: ${{ secrets.HOMEBREW_GITHUB_TOKEN }}
 
+      - name: Attest CLI release artifacts
+        id: attest_cli
+        uses: actions/attest@508db95dd578ae2727ebd6217d5ba78e4fbda05d # v4.2.1
```
divolgin: pinning to major or major.minor is sufficient

.github/workflows/release.yaml
divolgin: we want periodic updates. with SHA, we'll never update these.

divolgin (approved):

